PR #40, opened 2026-04-22
fix(sync): validate parquet magic bytes + skip broken files on rebuild

padak:
## Problem

\`da sync\` crashed during view rebuild:

\`\`\`
InvalidInputException: No magic bytes found at end of file
'/…/server/parquet/activity_contact_20260227134402.parquet'
\`\`\`

Root cause: the download succeeded at the HTTP layer (status 200) but the body was either truncated, empty, or an HTML / JSON error page that slipped through with a clean status code. \`_rebuild_duckdb_views\` then ran \`CREATE VIEW\` over it without any guard, and **a single bad file aborted the entire rebuild**, so even the valid parquets were not registered.

## Fix — two defensive hooks

1. **After download, validate PAR1 magic bytes** (head + tail). On failure, delete the file, skip writing to \`sync_state\`, record an error. The \`Errors: N\` summary now surfaces the real reason.
2. **\`_rebuild_duckdb_views\` is tolerant.** Pre-check PAR1 and wrap \`CREATE VIEW\` in try/except — one bad parquet is skipped with a stderr warning instead of taking down the whole rebuild. Leftover broken files from a previous failed sync no longer block subsequent syncs.

## Tests

- \`TestSyncParquetValidation\` — 3 new cases:
  - HTML body with HTTP 200 → recorded as error, no state commit, rebuild not called
  - Broken parquets on disk do not abort rebuild
  - \`_is_valid_parquet\` helper covers empty, too-short, HTML, valid
- Existing \`TestSyncHappyPath\` tests updated — their stream-download mock now writes a PAR1 shell so they stay green

## Base

Targeting \`ps/wheel-name-fix\` to keep the testing-instance deployment flow intact.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/40" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/cli/commands/sync.py
+++ b/cli/commands/sync.py
@@ -101,6 +101,15 @@ def sync(
             target = parquet_dir / f"{tid}.parquet"
             try:
                 stream_download(f"/api/data/{tid}/download", str(target))
+                # HTTP 200 is not enough — the server can (and has) returned
+                # a truncated / empty body with a clean status code. Validate
+                # PAR1 magic bytes so we don't commit state for a broken file.
+                if not _is_valid_parquet(target):
+                    target.unlink(missing_ok=True)
+                    raise ValueError(
+                        "downloaded file is not a valid parquet (missing PAR1 magic bytes) — "
+                        "the server-side snapshot may be corrupt or the stream was truncated"
+                    )
                 local_tables[tid] = {
                     "hash": server_tables[tid].get("hash", ""),
                     "rows": server_tables[tid].get("rows", 0),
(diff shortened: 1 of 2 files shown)


padak:
Superseded by #43 — commits e7aa859 / d14f8f0 on ps/wheel-name-fix (merged via #43) are a strict super-set of this PR's PAR1 check (they add MD5 manifest verification + atomic writes + retry on top). Closing without merge.

PR #48, opened 2026-04-23
chore(deploy): trust proxy headers + document HTTPS env vars

ZdenekSrotyr:
## Why

The app listens on plain HTTP (`uvicorn --host 0.0.0.0 --port 8000`) and expects a TLS terminator in front (Caddy, Cloudflare Tunnel, nginx, GCP LB, …). Today the wiring to that terminator is incomplete:

- Uvicorn is **not** started with `--proxy-headers`, so it ignores `X-Forwarded-Proto` / `X-Forwarded-Host`. `request.url` / `request.url_for(...)` returns `http://localhost:8000/...` instead of `https://<public-domain>/...`. Verified on a live deployment behind a Cloudflare Tunnel: sending `X-Forwarded-Proto: https` + `X-Forwarded-Host: <domain>` to the origin produced `redirect_uri=https%3A%2F%2Flocalhost%3A8000%2Fauth%2Fgoogle%2Fcallback` — headers silently dropped.
- `DOMAIN` and a new `SERVER_URL` env var are referenced in code (`_cookie_secure()` → `Secure` cookie flag; `password._base_url()` → reset-link base) but are **not documented** in `config/.env.template`. Every deployment has to rediscover them.

Net effect on a real deployment: session cookies are issued without the `Secure` flag, OAuth callback URLs only work by accident (when the proxy happens to forward an acceptable Host header), and the app has no way to know it's serving HTTPS.

## What changes

- `docker-compose.yml`, `docker-compose.test.yml`, `docker-compose.override.yml` — uvicorn command gains `--proxy-headers --forwarded-allow-ips='*'`. Uvicorn will then read `X-Forwarded-Proto` / `X-Forwarded-For` and set `request.url.scheme` / `request.client.host` correctly.
- `config/.env.template` — new **HTTPS / REVERSE PROXY** section documenting `DOMAIN` (enables Secure cookie) and `SERVER_URL` (deterministic base URL for OAuth callbacks + external links), with a note that uvicorn now trusts forwarded headers.

No code changes in `app/`, `src/`, or any connector — purely deployment/config.

## Trade-offs — why `--forwarded-allow-ips='*'`

`*` trusts `X-Forwarded-*` from any source IP. This is safe here because the container listens on the Docker bridge, only reachable from the host / the configured reverse proxy — no untrusted client can hit uvicorn directly. A stricter value (e.g. `127.0.0.1,172.16.0.0/12`) would be ideal, but varies per deployment and the permissive default keeps the template working out-of-the-box for Caddy, Cloudflare Tunnel, and ad-hoc nginx alike.

Downstream deployments that expose `:8000` publicly should override `command:` with a narrower allow-list.

## Rollout for existing deployments

Once merged, each instance needs to:

1. Pull the new image (compose change takes effect on container recreate).
2. Add to `.env`:
   ```
   DOMAIN=<public hostname>
   SERVER_URL=https://<public hostname>
   ```
3. `docker compose up -d` to recreate the app container.

## Verification

- CI: green.
- Manual sanity check post-merge: `curl -H 'X-Forwarded-Proto: https' -H 'X-Forwarded-Host: example.com' http://localhost:8000/auth/google/login` should now produce `redirect_uri=https%3A%2F%2Fexample.com%2Fauth%2Fgoogle%2Fcallback`. Before this PR it produced `http://localhost:8000/...`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,4 +24,4 @@ RUN uv build --wheel --out-dir /app/dist
 RUN uv pip install --system --no-cache .
 
 EXPOSE 8000
-CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "8000"]
+CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "8000", "--proxy-headers", "--forwarded-allow-ips", "*"]
--- a/app/auth/providers/google.py
+++ b/app/auth/providers/google.py
@@ -111,13 +111,15 @@ async def google_callback(request: Request):
             request.session.pop("login_next", None), default="/dashboard"
         )
 
-        # Redirect to target with token in cookie
-        is_production = os.environ.get("TESTING", "").lower() not in ("1", "true")
+        # Redirect to target with token in cookie. Match password/email providers:
+        # Secure only when DOMAIN is set (production with TLS), so the cookie is
+        # actually sent over plain HTTP in dev.
+        use_secure = os.environ.get("DOMAIN", "") != ""
         response = RedirectResponse(url=target, status_code=302)
         response.set_cookie(
             key="access_token", value=jwt_token,
             httponly=True, max_age=86400, samesite="lax",
-            secure=is_production,
+            secure=use_secure,
         )
         return response
 
--- a/config/.env.template
+++ b/config/.env.template
@@ -44,4 +44,22 @@ SESSION_SECRET=              # python -c "import secrets; print(secrets.token_he
 # DATA_DIR=/data            # Default: /data in Docker, ./data locally
 # LOG_LEVEL=info            # debug, info, warning, error
 # CORS_ORIGINS=http://localhost:3000,http://localhost:8000
-# DOMAIN=data.yourcompany.com  # For Caddy TLS (production profile)
+
+# ── HTTPS / REVERSE PROXY ───────────────────────────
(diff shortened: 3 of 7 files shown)


cvrysanek (approved):
Reviewed against infra repo's PR 49 / vr/https rollout. Changes are complementary:

- Dockerfile + compose files add `--proxy-headers --forwarded-allow-ips=*` — this is the app-side piece that makes Caddy's `X-Forwarded-Proto: https` + `X-Forwarded-Host` actually honored by uvicorn. Without it, OAuth callbacks come back as http://localhost:8000 even when Caddy terminates TLS correctly.
- DOMAIN-gated secure cookie in google.py matches the password/email provider pattern and aligns with infra's startup.sh which already writes DOMAIN=${hostname} into /opt/agnes/.env.
- .env.template documentation is non-runtime but keeps deployers informed.

No conflicts with PR 49 (zs/https-hardening edits app `command` lines; PR 49 adds caddy service + mounts — different hunks). CI green (test, docker-build, build-and-push all passing). Devin Review passed.

Rollout impact on merge:
- `stable` image rebuilds → foundryai-poc + foundryai-production pull new image within 5 min via auto-upgrade.timer. Change is additive (uvicorn trusts forwarded headers — no-op when no proxy present; DOMAIN-gated cookie secure flag matches foundryai-poc's already-DOMAIN-set state).
- No behavior change for deployments without DOMAIN.

LGTM to merge.

PR #57, opened 2026-04-25
feat(memory): V1.5 — audience distribution, contradiction privacy, confidence externalization

PavelDo:
## Summary

Three V1.5 features implemented together, all green (112 tests). Base branch is `pabu/local-dev` (includes all V1 hardening from the reviewed PR).

- **#1 — admin_contradictions privacy opt-in**: `GET /api/memory/admin/contradictions` gains `exclude_personal=true` (default). Personal items in contradiction pairs are replaced with `{id, hidden: true}` — contradiction record stays visible for governance, private content is not exposed. `exclude_personal=false` opt-in returns full content (KM_ADMIN only). Closes the open question from TODO.md.

- **#2 — Audience distribution wired end-to-end**: `POST /admin/mandate` and `/admin/batch` now persist `audience` to `knowledge_items.audience` (previously only went to audit log). `list_items()` and `search()` accept `user_groups` filter: `null` = no filter (admins see all); `[]` = only null/all; `["group:finance"]` = also includes that group. `GET /api/memory` derives caller groups from `users.groups` (new JSON column, schema v10 migration), prefixes with `"group:"`. Admins bypass the filter. Items with `audience IS NULL` default to visible-to-all. Wires to Padak's Google Workspace group sync (`users.groups` populated on login).

- **#3 — Confidence externalization**: `_BASE_CONFIDENCE`, `_MODIFIER_EFFECTS`, and decay config moved from hardcoded dicts to `instance.yaml` under `corporate_memory.confidence.*`. `configure(config)` in `confidence.py` lets any caller inject the YAML config; module-level defaults remain for backward compat. `apply_decay()` switched from linear to exponential (`0.5^(age/half_life)`, default half_life=12m) with per-source-type floor — `admin_mandate` floor=0.50 so admin policies never silently decay to zero.

## Schema changes

- `SCHEMA_VERSION` bumped 9 → 10
- V9→V10 migration: `ALTER TABLE users ADD COLUMN IF NOT EXISTS groups JSON`
- `_SYSTEM_SCHEMA` updated with `groups JSON` column

## Test plan

- [x] 112 tests pass (`pytest tests/test_memory_api.py tests/test_corporate_memory_v1.py`)
- [x] `TestAdminContradictionsExcludePersonal` (3 tests) — default hides, opt-in shows, non-personal always enriched
- [x] `TestAudienceDistribution` (6 tests) — mandate persists, batch mandate persists, group filter, no-group filter, admin bypass, null=all
- [x] `TestExponentialDecayWithLinearFallback` — linear fallback mode works via configure()
- [x] `test_admin_mandate_decay_floor` — admin_mandate never decays below 0.50
- [x] `test_configure_overrides_defaults` — configure() replaces module-level values


padak (approved):
# Code review

Approving — the base branch is `pabu/local-dev`, not `main`, so this doesn't ship anywhere user-visible yet. The findings below should all be addressed before this stack rolls forward to `main`.

Overall this is a strong, well-tested change. ADR + tests + prompt-injection hardening are above the typical bar; the issues are mostly about wiring and release hygiene rather than the architecture itself.

---

## BLOCKING — please address before this lands on `main`

### B1. `CHANGELOG.md` is not updated
Per `CLAUDE.md` ("Changelog discipline — non-negotiable") any user-visible change needs a `## [Unreleased]` entry. Concrete ask:

- `git diff v0.11.0..HEAD` and write up everything this stack adds.
- Land it as `## [0.12.0] — <date>` — this PR cuts the next release.
- Flag the breaking bits explicitly (default decay model now exponential, schema v8→v10 with the new `users.groups` JSON column, new `audience` distribution semantics, `apply_decay` signature).

### B2. Bump `version` in `pyproject.toml`
Same release: bump `pyproject.toml` `version` to `0.12.0` as part of this PR. The CHANGELOG and the package version ship together.

### B3. `confidence.configure()` is never called at app startup
`services/corporate_memory/confidence.py:configure()` exists, but `grep -rn "confidence.configure"` outside tests returns nothing. The PR description claims "Confidence externalization — wired end-to-end", but `instance.yaml` values are not picked up in production — only the new defaults (exponential decay + `admin_mandate` floor=0.50) take effect.

Please figure out how you want to close this loop:
- Wire `configure(instance_config["corporate_memory"]["confidence"])` from `app/main.py` `lifespan` / `create_app()` before the startup banner, **or**
- Drop the "externalized" claim from the PR description and ship only the new defaults, tracking the wiring as a follow-up.

Either is fine — pick one and make the PR description / TODO.md match what actually ships.

---

## HIGH

### H1. Re-scope: define "Agnes groups" + a Groups Mapping admin UI instead of raw Workspace strings
The PR description references "Padak's Google Workspace group sync" on a separate branch, and `users.groups` is treated as a list of free-form Workspace strings prefixed with `group:`. That couples Agnes too tightly to one identity provider and leaks Workspace group names into the data model.

Bigger picture from Padak: rather than wire one ad-hoc Google-groups → memory-admin path, model **Agnes groups (= Agnes roles)** as a first-class building block, plus a **Groups Mapping** screen in the Agnes admin UI:

- Admin picks 1..N Google Workspace groups from the company directory.
- Maps each to an Agnes role/group (e.g. `MEMORY_ADMIN`, `FINANCE_VIEWER`, ...).
- Users in the mapped Workspace groups auto-acquire the corresponding Agnes role on login.

Future features will introduce more Agnes roles (each feature ships its own role contract). So Agnes needs:

- A **registry of available Agnes roles** that each feature registers itself into.
- The mapping UI lists registered roles and lets the admin attach Workspace groups to each.

This redesign supersedes the "users.groups derived from raw Google Workspace strings, prefixed with `group:`" path that's in this PR. Please align with Padak before re-scoping H1; the rest of V1.5 (audience persistence, repo filter, admin bypass) can land independently of the mapping UI as long as the data model anticipates it.

### H2. `apply_decay()` signature change — accepted as breaking
Padak: not a concern, nobody is calling this externally yet. No deprecation needed. Just make sure it ends up in the CHANGELOG (B1).

### H3. Default decay semantics change for existing data
Flipping from linear (`-0.02/month`) to exponential (`half_life=12`) materially changes the score for items already in the corpus. After 12 months a `0.90` item goes from `0.66` (old) to `0.45` (new). `admin_mandate` items stay ≥0.50 because of the floor; everything else (`claude_local_md`, `session_transcript`, `user_verification`) drops faster.

Please figure out how you want to land this — options:
- A per-source-type floor for `user_verification.correction` (e.g. ≥0.40)?
- A one-release ramp keeping linear as the default, then flipping?
- Accept the discontinuity and call it out loudly in CHANGELOG?

Up to you — just don't ship it silently.

### H4. `POST /api/memory/admin/contradictions` docstring says "for testing/seeding"
The endpoint is production-facing with `KM_ADMIN` auth. The "for testing/seeding" framing reads like a backdoor that snuck into the public API. Please decide whether this is a real admin tool (then update the docstring + back it with UI) or a backdoor (then gate it behind `LOCAL_DEV_MODE` or move it under a dedicated `/admin/dev/` route group). Don't ship a third option where it's both.

---

## MEDIUM — please address

### M1. Silent fallback when deserializing `users.groups` (`app/api/memory.py:65–70`)
```python
if isinstance(raw_groups, str):
    import json as _json   # local import — top-level `import json` is already in this file
    try:
        raw_groups = _json.loads(raw_groups)
    except Exception:        # too broad
        raw_groups = []      # silently degrades a user to "only NULL/all visibility"
```
Three fixes:
- Drop the local import — top-level `json` is already imported in `memory.py`.
- Catch `json.JSONDecodeError`, not bare `Exception`.
- `logger.warning(...)` on parse failure so a corrupted `users.groups` cell doesn't silently strip a user's visibility.

Even better: pull this into a `UserRepository.get_groups(user_id) -> list[str]` helper so the deserialization lives in one place. (This also dovetails with H1 — if Agnes groups become first-class, this helper is where the resolution will live.)

### M2. `BATCH_CONTRADICTION_PROMPT` has unbounded size
`DEFAULT_CANDIDATE_LIMIT=100` × ~1 KB content per candidate ≈ 100 KB prompt. With `max_tokens=4096` on the response, a busy domain truncates the JSON → `LLMError` → `return []` (`services/corporate_memory/contradiction.py:2287`) → silent loss of contradictions. ADR mentions sharding; minimum: log a warning above ~50 candidates. Optimum: implement the sharding path that TODO.md already tracks for V2.

### M3. ADR / PR / code mismatch on schema version
- ADR text: "Schema bumped from v8 to v9".
- PR description: "schema v10".
- Code: `SCHEMA_VERSION = 10` with both `_V8_TO_V9` and `_V9_TO_V10` migrations applied.

Either extend the existing ADR with a v9→v10 (`users.groups` column) section, or write `docs/ADR-corporate-memory-v1.5.md` for the audience work. Knowledge drift between ADR and code on something this consequential will bite the next maintainer.

### M4. `merge` (LLM action) vs `merged` (API resolution) terminology
`_VALID_ACTIONS = {"kept_a", "kept_b", "merge", "both_valid"}` in `contradiction.py` vs `ResolveContradictionRequest.resolution`'s comment listing `merged`. Probably intentional (LLM proposes `merge`, admin commits as `merged`), but please add a one-line comment so a future maintainer doesn't "fix" one to match the other.

### M5. Double `repo.get_by_id` in `admin_contradictions`
For each contradiction the loop does two `get_by_id` calls (`app/api/memory.py:200–207`). Pre-existing N+1, but the privacy enrichment makes it more visible. A single `SELECT ... WHERE id IN (...)` over the distinct item IDs would cut DB round-trips from `2N` to 1. Worth doing while you're here.

### M6. `verification_detector/__main__` is no longer manual-only
The module docstring still claims "Trigger is manual-only today", but `detect_and_record` is now invoked inline per new item. A bulk replay of sessions triggers N×100 LLM calls automatically. Please add a cost guard or `--no-contradictions` flag and update the docstring.

### M7. Triple `get_corporate_memory_config()` call in `app/web/router.py:271`
```python
groups=list(get_corporate_memory_config().get("groups", {}).values()) if isinstance(get_corporate_memory_config().get("groups"), dict) else get_corporate_memory_config().get("groups", []),
```
Three calls in one expression. Even if cached, it's a readability speed bump. Quick refactor:
```python
cfg = get_corporate_memory_config()
groups_raw = cfg.get("groups", {})
groups = list(groups_raw.values()) if isinstance(groups_raw, dict) else (groups_raw or [])
```

---

## LOW

### L1. `docs/pd-ps-comments.md` references — please delete
Padak: that doc is internal review notes — repo bloatware. Drop both references (in `ADR-corporate-memory-v1.md` and `TODO.md`). The originating notes can live in `docs/research.md` (gitignored per the global instructions); the ADR can just say "internal review".

### L2. `docs/TODO.md` — please remove from the repo
Padak: anything important from there should move to GitHub Issues with the right milestone (`V1.5`, `V2`). Project TODO files in repos rot fast; the issue tracker is the right home.

### L3. `docs/VIBECODING_GUIDE.md` — please remove from this OSS repo
Padak: this is personal documentation, not a fit for the public OSS distribution. Pull it into your own notes / private repo if it's useful to you — the OSS repo isn't the right venue.

### L4. `_PRIVILEGED_VIEWER_ROLES` includes `Role.ADMIN`
`KM_ADMIN` is the explicit corporate-memory governance role; `ADMIN` is the system super-admin. If platform-level admins are not supposed to read personal memory items (compliance reason), `ADMIN` should not be in this set. ADR Decision 1 doesn't make the intent explicit. Please confirm intent and either keep `ADMIN` with a one-line comment why, or drop it.

### L5. UI categories fix → CHANGELOG `### Fixed` bullet
Removing the hardcoded `Data Analysis` / `API` / `Performance` buttons in favour of a Jinja `{% for cat in categories %}` loop is exactly the "Hardcoded UI vs Dynamic Data" pattern fix. Add a `### Fixed` bullet to the CHANGELOG entry from B1.

### L6. `dashboard.html` URL fix → CHANGELOG `### Fixed` bullet
`/api/corporate-memory/stats` → `/api/memory/stats` was a dead link. Add a `### Fixed` bullet to the CHANGELOG entry from B1.

---

## MISC

### CI is red
`tests/test_corporate_memory_v1.py::TestSchemaValidation::test_*_golden_valid` (×3) fail with `ModuleNotFoundError: No module named 'jsonschema'`. The PR doesn't add the dep to `pyproject.toml`. Either add it to `[project.optional-dependencies].dev`, or wrap the import with `pytest.importorskip("jsonschema")`. Either way, CI must be green before merge.

### Praise (please keep doing this)
- **Privacy enforcement is rigorous**: 5 vectors fixed (list, search, stats, provenance, vote), 11 regression tests, 404-not-403 to avoid existence leaks. ADR Decision 1 lays out the rationale cleanly.
- **Prompt injection hardening**: `<turn role="…">` / `<item id="…">` wrappers with end-tag neutralization plus an explicit Trust Boundary instruction in the prompt. Combined with strict structured outputs that's real defense-in-depth.
- **LLM trust inversion on confidence**: removing `base_confidence` from the verification schema means the LLM can't set its own credibility — derived deterministically in code. Right governance pattern.
- **Batch contradiction redesign**: N pairwise calls → 1 batched call + drop hallucinated `candidate_id` with a warning. Cost optimization × scale + safety in one move.
- **Per-source-type confidence floor** (`admin_mandate=0.50`) so admin policies never silently decay to zero.
- **Docstring quality**: `find_contradiction_candidates` in `knowledge.py` has ~14 lines of WHY (why domain-only narrowing, why drop personal items unconditionally). Reads beautifully.

### Other notes worth tracking (not blocking this PR)
- The new `verification_evidence` table accumulates rows but no job re-computes `knowledge_items.confidence` from "additional verifiers" yet — TODO.md (or a GitHub Issue per L2) already tracks this as Q3 V2 work.
- `keboola-deploy.yml` consumers pin to `keboola-deploy-latest`; once 0.12.0 cuts, an explicit redeploy tag is the right way to roll the schema migration to those instances.

---

## Approval

Approving — base branch is `pabu/local-dev`, not `main`. This doesn't ship to anyone yet. Please address the BLOCKING items before whatever PR carries this stack opens against `main`.

PR #71, opened 2026-04-26
release(0.11.3): internal roles + external→internal group mapping (foundation)

padak:
## Summary

Two-layer authorization model for Agnes:

- **External groups** — Cloud Identity / Google Workspace groups, already pulled into `session.google_groups` at sign-in. Org-managed.
- **Internal roles** — Agnes-defined capabilities (e.g. `context_admin`, `agent_operator`, `dataset_finance_reader`). App-managed; modules self-register at import time.
- **Group mappings** — admin-curated many-to-many table binding external group IDs to internal role keys.

The resolver joins `session.google_groups` against `group_mappings` once at sign-in and writes the resolved role keys into `session["internal_roles"]`. Per-request permission checks (`Depends(require_internal_role("…"))`) read the cached list off the session — no DB hit per request. Refresh requires re-login, same semantics as `session.google_groups`.

Authored by web-Claude as a single commit (`ef6b2c1`); a polish commit on top (`d590579`) addresses two findings from a code-reviewer pass — see *Review iterations* below.

**No admin UI yet** — this PR is the foundation. Mapping rows must be created via the repository directly until the management UI ships in a follow-up.

### Schema v8 (auto-migration v7→v8)

Two new tables in `src/db.py`:

- **`internal_roles`** — `id`, `key UNIQUE`, `display_name`, `description`, `owner_module`, timestamps. Holds app-defined capabilities; modules call `register_internal_role(...)` at import time and the startup hook syncs the registry into this table (idempotent, never deletes).
- **`group_mappings`** — `id`, `external_group_id`, `internal_role_id` FK, `assigned_at`, `assigned_by`, `UNIQUE(external_group_id, internal_role_id)`. Admin-managed bindings.

Migration uses `CREATE TABLE IF NOT EXISTS` (rerun-safe) and matches the v5/v6/v7 patterns already in `_ensure_schema`.

### `app/auth/role_resolver.py` (190 LOC, new)

- `register_internal_role(key, *, display_name, description, owner_module)` — validates `lower_snake_case` key (`^[a-z][a-z0-9_]{0,63}$`), no-op on re-import with same fields, raises `ValueError` on conflicting fields (two modules collided on a key).
- `sync_registered_roles_to_db(conn)` — startup reconciliation: inserts new keys, updates drifted metadata, **never deletes** (preserves mappings even if a module unloads).
- `resolve_internal_roles(external_groups, conn)` — joins `group_mappings`, returns sorted/deduped role-key list.
- `require_internal_role("key")` — FastAPI dependency factory; reads `session["internal_roles"]`, raises 403 with `Requires internal role 'key'` when missing.

### Integration points

- **OAuth callback** (`app/auth/providers/google.py`) — after `_fetch_google_groups`, before issuing JWT: `request.session["internal_roles"] = resolve_internal_roles(...)`. Best-effort; resolver errors fall back to `[]` and don't break login.
- **Dev-bypass** (`app/auth/dependencies.py`) — recomputes `internal_roles` whenever the cached list is missing or external groups change. Mirrors the OAuth callback's "always-write" semantics so a fresh dev session never reaches `require_internal_role` with the key absent.
- **Startup hook** (`app/main.py`) — `sync_registered_roles_to_db(conn)` runs once after schema migration; logs the registered role keys at startup.

### Docs + tests + CHANGELOG

- **`docs/internal-roles.md`** — architecture overview, module-author workflow (registering a role), admin workflow (creating a mapping via repository, until the UI ships), callsite example, **PAT/headless constraint**, resolution timing, future work.
- **`tests/test_role_resolver.py`** (285 LOC) — 21 tests across registry/sync/resolver/dependency: key validation, re-import idempotency, conflict detection, drift updates, FK joins, `require_internal_role` allow/deny.
- **`CHANGELOG.md`** — `## [Unreleased] → ### Added` bullet.

## Review iterations addressed in commit `d590579`

The first commit (`ef6b2c1`) went through a code-reviewer subagent pass before opening this PR. Two findings adopted; three skipped (legit pre-existing patterns or design choices, not regressions).

### Adopted

- **Dev-bypass populated `internal_roles` on first request, not just on group change.** The previous guard only resolved when `groups_changed=True`, which left a hole for `LOCAL_DEV_GROUPS=""` (explicit empty) on the first browser request: `target=[]`, `current=None`, neither write branch fired, `internal_roles` stayed unset, and `require_internal_role` then 403'd with no roles to check against. The OAuth callback writes `session["internal_roles"]` unconditionally on sign-in (even `[]`); dev-bypass now matches that semantics — populates whenever the key is absent or groups changed.

- **Documented PAT/headless constraint.** `session["internal_roles"]` is never populated for Bearer-token clients (no signed session cookie), so any `require_internal_role(…)` endpoint returns 403 for PAT regardless of group memberships. Same constraint already applies to `session.google_groups` and is intentional (PAT JWTs don't snapshot group state — they could change after issuance), but `docs/internal-roles.md` didn't surface it. New *"PAT and headless requests"* section spells out the constraint, the rationale, and three escape valves (`users.role` for the gate, route through OAuth, or wait for the planned `da admin grant-role` CLI helper).

### Skipped (with rationale)

- *Repository `_row_to_dict` reads `self.conn.description` after `fetchone()`* — flagged as fragile, but it's a pre-existing pattern across all repositories in `src/repositories/` and not a regression introduced by this PR.
- *Module-self-registration race if request arrives before sync completes* — startup ordering already runs `sync_registered_roles_to_db` before the FastAPI app starts accepting requests; a request can't beat it. The race-on-different-fields case is correctly raised as `ValueError`.
- *Session cache staleness when admin changes a mapping* — explicitly documented in both the resolver docstring and `docs/internal-roles.md`. Same trade-off as `session.google_groups`; cache invalidation would be a separate PR.

## Test plan

- [x] `pytest tests/test_role_resolver.py -v` → **21/21 passed locally** (Python 3.13)
- [x] `pytest tests/test_auth_providers.py -v` → 33/33 still pass (no regression from the dev-bypass change)
- [x] Schema migration: rerun on existing v7 DB is no-op (CREATE TABLE IF NOT EXISTS); fresh DB lands at v8 with both tables present.
- [ ] CI: standard `test` + `docker-build` jobs on the PR; `docker-e2e` skips on push events as designed.
- [ ] Manual smoke (post-merge, with admin UI): create one mapping via repo, sign in, hit `Depends(require_internal_role(...))` endpoint — returns 200.

## Follow-ups (out of scope)

Listed under *Future work* in `docs/internal-roles.md`:
- **`/admin/role-mapping` UI** — list registered roles + their mappings, add/remove mapping rows, surface drift between code registry and DB rows.
- **Audit-log entries** for mapping create/delete (`audit_log` with `action="role_mapping.created"` etc).
- **`da admin grant-role <user-email> <role-key>` CLI helper** — explicit per-user grant outside the group-mapping flow, for ad-hoc CLI / PAT flows.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/71" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


PavelDo:
Great foundation — the group mapping table design is exactly what we agreed on and it's done cleanly. A few things to align before this lands:

## Sequencing issue (must fix before merge)

`pabu/local-dev` is currently at schema v10 (V1 + V1.5 work). This PR creates a v7→v8 migration with `internal_roles` + `group_mappings`. Both branches define a **v7→v8 migration with different tables** — if merged without coordination, two schemas claim the same version slot and the migration chain breaks silently on existing DBs.

Suggested order:
1. Merge `pabu/local-dev` → `main` first (main becomes v10)
2. Rebase this PR onto that `main` as a v10→v11 migration

Also: this PR sets `version = "0.11.3"` but `pabu/local-dev` is already at `0.12.0`. One needs to yield before either hits main.

## Architecture gap — audience filter still disconnected

After this PR, `session["internal_roles"]` holds resolved capabilities. But `GET /api/memory` audience filter reads `user.get("groups")` from DuckDB (Agnes group strings like `"finance"`). These two systems are parallel and don't connect — the memory audience filter will still be inert in production after this merges.

We need to decide: does the memory audience filter route through internal roles, or do we bridge Google groups to `users.groups` separately on login?

## Security findings

- **Stale session roles (medium)**: Admin revokes a mapping → user retains the capability until re-login. Documented in `docs/internal-roles.md`, but worth adding the recommended workaround explicitly: deactivate + reactivate the user forces re-login.
- **SESSION_SECRET length**: `require_internal_role` trusts the session; Starlette HMAC is only as strong as the secret. Add a startup assertion that `SESSION_SECRET` is ≥ 32 bytes (same as `JWT_SECRET_KEY`).
- **`_clear_registry_for_tests()`**: Test-only helper exported from a production module. Should be behind `if os.getenv("TESTING"):` or moved to the test file.
- **No audit log on resolution**: `resolve_internal_roles()` logs nothing. When a user's roles are wrong (mapping error, group fetch failure → empty), there's no trace to debug from. An info-level log ("user X resolved internal_roles=[...]") would help.

## Missing tests (vibecoding check)

✅ What's covered: registration, sync idempotency, drift updates, no-delete, many-to-many resolution, deduplication, malformed input, `require_internal_role` allow/deny.

❌ What's missing:
- **Stale session**: seed mapping → user logs in → change mapping → assert user still gets old roles (documents the known limitation with a failing-then-annotated test)
- **PAT caller** hitting `require_internal_role` — what exact 403 detail does the API consumer see?
- **DB error** in `resolve_internal_roles` → graceful `[]` fallback (exists in code, not tested)
- **Full OAuth integration**: `_fetch_google_groups` → `resolve_internal_roles` → `session["internal_roles"]` populated end-to-end

## Minor

- `CLAUDE.md:235` still says v7 (Devin caught this — quick fix)
- PAT 403 response body: the `403 Forbidden` from `require_internal_role` doesn't explain that PATs can't carry session roles. Adding `detail="This endpoint requires an interactive session; PAT tokens do not carry internal roles."` would save API consumers debugging time.

## Overall

The architecture is correct and production-quality — group mapping table is exactly the right call. Three things before merge: schema version sequencing, the audience filter bridge decision, and the four missing tests. Happy to coordinate on the merge order once we align on sequencing.

padak:
## Pavel — review feedback addressed

Round-2 polish in 8063548 (just pushed). Walkthrough:

**Sequencing — confirmed: this PR merges first, you rebase `pabu/local-dev` on top.** Talked it over offline, agreed it's the cleaner path because your memory PR depends on these `internal_roles` rows existing for the audience filter. Once `main` carries v8 + `0.11.3`, your branch becomes a v8→v9 (or v10 if V1.5 stays bundled) migration with `0.12.0` as the next bump — single linear chain instead of two parallel v7→v8 forks.

**Architecture gap — audience filter still disconnected.** Acknowledged, deferred to your rebase PR. The right place to wire `session["internal_roles"]` (or whatever the agreed bridge is) into the memory audience filter is in the same change that introduces the filter, not preemptively here. Happy to pair on the bridge design when you rebase — the open question is whether `users.groups` (DuckDB column from V1.5) gets repurposed as the internal-role mirror, or whether the memory filter starts reading `session["internal_roles"]` directly. Either works mechanically; trade-off is "DB is authoritative + works for PAT" vs "session is authoritative + cheaper per-request".

**Security findings — all four addressed:**
- *Stale session* — added the deactivate+reactivate workaround paragraph to `docs/internal-roles.md` (the supported "force re-resolve now" lever).
- *SESSION_SECRET length* — startup `UserWarning` when shorter than 32 chars in `app/main.py`, mirrors the existing `JWT_SECRET_KEY` gate. Both HMAC surfaces sign trust-laden state, so consistent gating.
- *`_clear_registry_for_tests()`* — now refuses unless `TESTING=1`, raises `RuntimeError` with explanation otherwise.
- *No audit log on resolution* — INFO log on every successful resolve in both writers (OAuth callback + dev-bypass): `"Resolved N internal role(s) for <email>: <roles>"`.

**Missing tests — all four added** (`tests/test_role_resolver.py`, +4 tests, full suite stays green at 1395 passed):
- `test_stale_session_keeps_old_roles_after_mapping_change` — pins the documented limitation as a contract test, comment explains how to update if/when invalidation lands.
- `test_pat_caller_gets_pat_specific_403_detail` — covers the new PAT-aware 403 message wording.
- `test_oauth_pipeline_groups_to_internal_roles` — end-to-end groups→`session["internal_roles"]` data flow.
- `test_dev_bypass_falls_back_to_empty_on_resolver_error` — monkeypatch resolver to raise, asserts `internal_roles=[]` and the gate fails closed.

**Minor:**
- `CLAUDE.md:235` — bumped to v8 (Devin caught this too — same fix covers both reviews).
- *PAT 403 detail* — `require_internal_role` now distinguishes "no session" (Bearer/PAT) from "signed-in, missing role" and surfaces an actionable message in the first case: `"Requires internal role 'X'. This endpoint needs an interactive (OAuth) session — Bearer/PAT tokens do not carry session-resolved roles by design."`

Bumped CHANGELOG `[0.11.3]` with the user-visible additions (PAT detail, deactivate+reactivate doc) and an `### Internal` block for the audit log + SESSION_SECRET warning + TESTING gate. Diff is +231/-10 across 8 files; all in the spirit of the original PR scope. Ready for re-review.

Issue #83, opened 2026-04-27
[SECURITY][CRITICAL] Jira webhook fail-open + path traversal on issue key


davidrybar-grpn:
## Why this is a parent issue
Two issues on the same endpoint (`/webhooks/jira`) that compound into a single attack chain: signature verifier fails open AND the issue key flows unvalidated into filesystem paths. Land the validation on the same boundary in one PR — both concerns are resolved with two small additions in `app/api/jira_webhooks.py`.

> **Note:** Jira connector is not enabled in the current GRPN PoC deploy, so this isn't `grpn-priority`. It is still a real critical bug for any deploy that enables Jira webhooks.

## Sub-tasks
- [ ] **C4 (CRITICAL)** — `_verify_signature` returns `True` when `JIRA_WEBHOOK_SECRET` is unset → fail-open.
  - File: `app/api/jira_webhooks.py:27-48`
  - Missing `JIRA_WEBHOOK_SECRET` is the most likely fresh-deploy misconfiguration. Current code accepts an unsigned POST → writes to `webhook_events/` → triggers `process_webhook_event` (disk writes, `rebuild_source`, parquet writes).
  - Fix: flip to fail-closed; explicit `JIRA_WEBHOOK_SKIP_VERIFY=1` opt-out env var preserves a documented escape hatch:
    ```python
    if not secret:
        if os.environ.get("JIRA_WEBHOOK_SKIP_VERIFY") == "1":
            logger.warning("JIRA webhook signature verification disabled via JIRA_WEBHOOK_SKIP_VERIFY=1")
            return True
        logger.error("JIRA_WEBHOOK_SECRET not configured; rejecting webhook")
        return False
    ```
  - Tests:
    - `test_jira_webhook_rejects_when_secret_unset` — secret empty, no opt-out → 401, `process_webhook_event` not called.
    - `test_jira_webhook_accepts_when_skip_verify_explicit` — opt-out explicit → accepted (regression for the escape hatch).
    - `test_jira_webhook_rejects_bad_signature_with_secret` — secret set, signature wrong → 401 (existing behavior).

- [ ] **C10 (HIGH)** — `issue_key` from webhook payload flows into filesystem paths without validation.
[...]

Issue #81, opened 2026-04-27
[SECURITY][CRITICAL] Harden the connector → orchestrator trust boundary


davidrybar-grpn:
## Why this is a parent issue
The orchestrator and the read-only analytics path treat data produced by connectors (extract.duckdb files, registry rows) as inherently safe. They aren't — connectors share the same code-review surface as the rest of the server, but live in a separate trust zone in the threat model. Three concrete weaknesses share this boundary; fixing them in one PR (or one short series) is more cohesive than scattering them.

## Sub-tasks
- [ ] **[PoC-blocker]** **C1 (CRITICAL)** — `_remote_attach` allows arbitrary DuckDB community extensions + secret exfiltration via `token_env`.
  - Files: `src/orchestrator.py:200-256`, `src/db.py:340-424`
  - Both paths read `_remote_attach` rows from each extract.duckdb and execute `INSTALL <ext> FROM community; LOAD <ext>; ATTACH '<url>' AS <alias> (TYPE <ext>, TOKEN '<token>')`. `extension`/`alias` are validated only against a format regex — no allowlist of *which* extensions can be installed. `token_env` is read with `os.environ.get(token_env, "")` for any env var name (e.g. `SESSION_SECRET` exfiltrated as TOKEN). The orchestrator path also lacks the `safe_url = url.replace("'", "''")` escape that `src/db.py:411` already applies → SQL injection in the ATTACH options.
  - Fix: centralize remote-attach into one helper used by both paths, enforcing in order:
    1. Hard allowlist of extensions: `{"bigquery", "keboola"}`
    2. Hard allowlist of `token_env` names: `{"KBC_TOKEN", "KBC_STORAGE_TOKEN", "GOOGLE_APPLICATION_CREDENTIALS"}`
    3. Mirror the URL escape from `src/db.py:411` so both paths are byte-identical.
  - Tests:
    - `test_remote_attach_rejects_unknown_extension` — `extension="httpfs"` → no INSTALL/LOAD, WARNING logged.
    - `test_remote_attach_rejects_unallowed_token_env` — `token_env="SESSION_SECRET"` → rejected.
    - `test_remote_attach_url_with_quote_is_escaped` — URL containing `'` doesn't break out of the literal.
    - `test_remote_attach_accepts_known_pair` — `bigquery` + `GOOGLE_APPLICATION_CREDENTIALS` → succeeds (smoke).

[...]

Issue #44, opened 2026-04-22
[SECURITY][CRITICAL] Script-API sandbox escape → RCE for analyst role


padak:
Source: [\`docs/padak-security.md\` §1](https://github.com/keboola/agnes-the-ai-analyst/blob/main/docs/padak-security.md#1-critical-script-api-sandbox-escape--rce-for-the-analyst-role)

## Trigger (verified analytically)

Role required: \`Role.ANALYST\` (not admin).

\`\`\`
POST /api/scripts/run
body: {
  code: \"().__class__.__base__.__subclasses__()[N].__init__.__globals__['system']('id')\"
}
\`\`\`

The AST walker and string allowlist in \`app/api/scripts.py:116–180\` block direct \`import os\` / \`exec\`, but neither stops attribute traversal through Python's class hierarchy (\`__class__ → __base__ → __subclasses__() → __globals__\`).

## Impact

Arbitrary OS commands under the FastAPI process uid. Access to \`DATA_DIR\` (DuckDB state, cached parquet), \`.jwt_secret\`, env vars, anything mounted into the container.

## Suggested fix

**Minimum (same-day):**
- Add to the string-pattern blocklist: \`__subclasses__\`, \`__globals__\`, \`__mro__\`, \`__bases__\`, \`__class__\`, \`__dict__\`, \`__code__\`.
- In the AST walker, reject any \`ast.Attribute\` whose \`attr\` starts and ends with \`__\`.

**Correct (longer-term):** don't run untrusted Python in-process at all. Options, in decreasing scope:
1. Drop server-side script execution entirely — move it to client-side or remove the feature.
2. Gate the endpoint behind \`Role.ADMIN\` if it must stay.
3. Run the sandbox in nsjail / gVisor / Pyodide with no shared filesystem or network.

## Ack criteria

- PoC payload above returns \`403\` or \`400\` (not the \`id\` output).
- Unit test covering each of the blocklisted dunder paths.
- Docs note: this endpoint is not a sandbox, it's a convenience — do not expose to untrusted users.

davidrybar-grpn:
External codebase review (`docs/CODEBASE_REVIEW.md` §C3) re-verified this against the current branch and reached the same conclusion. Two small additions:

[...]

PR #106, opened 2026-04-28
ma/staging → main: RBAC v13 + marketplace + #81 security hardening + OSS neutralization

minasarustamyan:
## Summary

Bundles 14 commits accumulated on `ma/staging` against `main`. Three large workstreams plus a handful of follow-up UI fixes:

### RBAC redesign + Claude Code marketplace (`ed524af` and friends)
- **BREAKING** schema v13 — drops the `core.viewer/analyst/km_admin/admin` hierarchy and the `internal_roles` / `group_mappings` / `user_role_grants` / `plugin_access` tables. Replaced by `user_group_members` + `resource_grants`. Atomic v12→v13 backfill.
- New authorization vocabulary in `app.auth.access`: `require_admin` (Admin-group god mode) and `require_resource_access(ResourceType.X, "{path}")` for entity-scoped grants. Single DB lookup per request, no session cache.
- New aggregated Claude Code marketplace at `/marketplace.zip` + `/marketplace.git/*` (PAT-gated, RBAC-filtered, content-addressed cache). Marketplace **drops admin god-mode** (`7572969`) — admins now curate their own view via grants like everyone else.
- Admin UI consolidated into `/admin/access`; CLI subcommands replaced (`da admin group/grant *`).
- `ResourceType.TABLE` (`852dc6c`) — admins can grant table-level access; runtime enforcement still flows through legacy `dataset_permissions` (see `docs/TODO-rbac-data-enforcement.md`).

### Issue #81 security/ops hardening (Groups A–D)
- **A** (`23be8ad`): orchestrator ATTACH hardening (extension/url/alias allow-listing).
- **B** (`ef74ec0`): Keboola extractor exits with **3 codes** — `0` success / `1` total fail / `2` **partial** fail. Sync API logs `PARTIAL FAILURE (exit 2)` as a data-quality alert. **Operators whose alerting treats any non-zero exit as a hard error must teach it that exit 2 is partial.**
- **C** (`72230c3`): schema v10 `view_ownership` — refuses to silently overwrite a prior connector's view name on collision; logs ERROR with both sources.
- **D** (`569cd90`): extractor-side identifier validation.
- Plus #83 (`2f783c5`) — Jira webhook fail-closed when `JIRA_WEBHOOK_SECRET` unset, and path-traversal fix.
- Plus #44 (`24e81fb`) — entire Script API (`/api/scripts/*`) is now admin-only; planted-script + sandbox-bypass risk.

### OSS vendor-neutralization waves 1 & 2 (`4e4d2a3`)
Generic ops scripts moved `scripts/grpn/` → `scripts/ops/`. Customer-specific identifiers (`prj-grp-…`, `FoundryAI`, internal hostnames, dev/prod VM IPs) replaced with placeholders across code docstrings, test fixtures, Terraform variable descriptions, the Caddyfile, OAuth probe, and the hackathon-era planning docs. **Downstream infra repos that copy `scripts/grpn/agnes-tls-rotate.sh` or `agnes-auto-upgrade.sh` from this repo must update the path.**

### UI / deploy follow-ups
- `b931469`: `docker-compose.host-mount.yml` uses `o: bind,rbind` — plain bind shadowed sub-mounts (dual-disk VMs were silently writing state to the wrong disk).
- `fc3f43d`: `/catalog`, `/admin/tables`, `/admin/permissions` link `style-custom.css` so the shared header renders styled.
- `f06c94a`: `/admin/access` group sidebar grant-count badge no longer reverts to a stale snapshot when switching groups.
- `968a2af`: per-resource-type colored stripes on `/admin/access`.

## BREAKING flags operators should grep for

- RBAC schema v13 (rebuild role-grant audits; admin promotion is now Admin-group membership).
- Marketplace endpoint admin god-mode dropped — admins must grant plugins to themselves to see them.
- Admin REST surface unified under `/api/admin/{groups,grants,resource-types}`; old `role_management` / `plugin_access` endpoints gone.
- CLI: `da admin role/mapping/grant-role/revoke-role/effective-roles` removed.
- Keboola exit code 2 now means partial success.
- Script API admin-only.
- Generic ops scripts moved to `scripts/ops/`.

## Test plan

- [ ] Local: pytest tests/ -v (149 changed test files; verify suite green).
- [ ] Run v12→v13 migration against a snapshot of a populated system DB; check `user_group_members` / `resource_grants` backfilled and legacy tables dropped.
- [ ] Sign in via Google OAuth on a fresh account → verify Everyone membership auto-assigned and Workspace groups synced (source='google_sync').
- [ ] Promote `SEED_ADMIN_EMAIL` → confirm Admin-group membership shows source='system_seed'.
- [ ] `/admin/access`: switch between groups, toggle grants, verify badge counts stay in sync; confirm `ResourceType.TABLE` items render grouped by bucket.
- [ ] Marketplace: as admin without explicit grants, `/marketplace.zip` is empty; after granting a plugin to Admin group, plugin appears.
- [ ] Keboola partial-failure smoke: induce one table failure, confirm extractor exits 2 and sync API logs PARTIAL FAILURE alert.
- [ ] View collision: register two connectors emitting the same `_meta.table_name`, confirm second view is rejected with the new ERROR.
- [ ] Jira webhook: with `JIRA_WEBHOOK_SECRET` unset, confirm 401/403 (fail-closed).
- [ ] Script API: as non-admin, confirm `/api/scripts/*` returns 403.
- [ ] Dual-disk host-mount overlay: redeploy with the rbind fix and confirm `system.duckdb` writes land on the dedicated state disk.


cvrysanek (commented):
Read-through review on the staging→main bundle. CI is green and the workstreams are well-scoped — most of these are migration-safety questions and a couple of behavior-change call-outs operators will trip over. No blockers from me; just want a few things either tightened or explicitly waved through before merge.

High-level themes:
- v12→v13 finalize runs DDL + multi-table backfill on a single connection without an explicit transaction; one mid-flight failure leaves a half-migrated DB.
- The admin backfill only carries `core.admin` forward — module-author roles (`context_engineering.admin`, `corporate_memory.curator`, etc.) are dropped on the floor.
- `ResourceType.TABLE` is admin-grantable in the UI but runtime enforcement is still legacy `dataset_permissions`. Shipping the UI before the enforcement half lands creates a user-visible drift (admin grants → user still 403s).
- `/api/scripts/*` admin-only is a contract change for any cron / scheduler PAT consumer — worth a CHANGELOG line.
- (Out-of-PR but in scope of the bundle's PR description) the Keboola exit-code-2 = partial contract is in base, not this diff. Worth noting in CHANGELOG that anything outside the sync API path treating non-zero as fatal — systemd `Type=oneshot`, k8s Jobs, cron-to-alerting wrappers — will misclassify partial success as a hard outage. The structured `_meta` table is the safer signal for downstream consumers.

src/db.py
```diff
@@ -637,6 +732,192 @@ def get_analytics_db_readonly() -> duckdb.DuckDBPyConnection:
...
+            )
+        elif not existing[1]:
+            # Promote pre-existing manual group to system without touching desc.
+            conn.execute(
+                "UPDATE user_groups SET is_system = TRUE WHERE id = ?",
+                [existing[0]],
+            )
+
+
+def _v12_to_v13_finalize(conn: duckdb.DuckDBPyConnection) -> None:
```
cvrysanek, line 787: `_v12_to_v13_finalize` runs ~5 backfill passes and 4 `DROP TABLE`s on the same connection with no `BEGIN`/`COMMIT` wrapping the work. If step 6 (the legacy drops) fails halfway in a forked install — e.g. a residual FK from a downstream extension — you've already mutated `user_group_members` / `resource_grants` and the outer caller still bumps `schema_version` to 13 at line 1156. Half-migrated DB, no rollback path beyond the snapshot file.

Suggest either:
1. Wrap the body in `conn.begin()` … `conn.commit()` and let an exception propagate (the snapshot restore path already exists in `_ensure_schema`), or
2. Move the schema_version bump *into* this function so a failure leaves the DB on v12 with the new tables present-but-empty (still re-runnable thanks to the `IF NOT EXISTS` and `ConstraintException` swallowing).

src/db.py
```diff
@@ -637,6 +732,192 @@ def get_analytics_db_readonly() -> duckdb.DuckDBPyConnection:
...
+    has_internal_roles = conn.execute(
+        "SELECT 1 FROM information_schema.tables WHERE table_name = 'internal_roles'"
+    ).fetchone()
+    has_user_role_grants = conn.execute(
+        "SELECT 1 FROM information_schema.tables WHERE table_name = 'user_role_grants'"
+    ).fetchone()
+    if has_internal_roles and has_user_role_grants:
+        admin_users = conn.execute(
+            """SELECT DISTINCT g.user_id
+               FROM user_role_grants g
```
cvrysanek: Admin backfill matches `r.key = 'core.admin'` only. Any user holding a non-core capability grant (`context_engineering.admin`, `corporate_memory.curator`, anything a module author registered with `register_internal_role`) silently loses it on v12→v13: the row is in `user_role_grants` until step 6 drops the table, but no `resource_grants` equivalent is written.

The PR description frames v13 as collapsing the role hierarchy onto the Admin group, which is fine for `core.*`. It doesn't explain what happens to module-registered capability roles — are those expected to be re-issued by the module owner post-upgrade, or is there a backfill I'm missing? If the former, the migration should at least log the dropped grants (count + role keys) so operators know what they need to recreate.

src/db.py
```diff
@@ -637,6 +732,192 @@ def get_analytics_db_readonly() -> duckdb.DuckDBPyConnection:
...
+                if not group_row:
+                    continue
+                try:
+                    conn.execute(
+                        """INSERT INTO user_group_members
+                           (user_id, group_id, source, added_by)
+                           VALUES (?, ?, 'google_sync', 'system:v13-backfill')""",
+                        [user_id, group_row[0]],
+                    )
+                except duckdb.ConstraintException:
```
cvrysanek, line 885: `except duckdb.ConstraintException: pass` is repeated four times in finalize (837, 863, 876, 897) and is the right call for the re-run-safety scenario. But the silent pass also masks the case where two source paths converge on the same `(user_id, group_id)` with different `source` values — e.g. a user is in the Google-synced "engineering" group AND admin-added to it. The PK collides, the second insert is dropped, and the earlier `source='google_sync'` row wins forever. That row is then eligible for the nightly Google sync's DELETE+INSERT, undoing the admin-added intent.

Worth at least a `logger.debug` at each site so the collisions are observable. Better: do a `SELECT 1 FROM user_group_members WHERE user_id=? AND group_id=?` first and log when sources disagree.

src/db.py
```diff
@@ -637,6 +732,192 @@ def get_analytics_db_readonly() -> duckdb.DuckDBPyConnection:
...
+        try:
+            conn.execute(stmt)
+        except Exception as e:
+            logger.warning("v13 drop failed (%s): %s", stmt, e)
+
+    # 7. Drop users.groups column. DuckDB supports DROP COLUMN; silently no-op
+    # if it's already gone (fresh-install path or partial re-run).
+    if has_groups_col:
+        try:
+            conn.execute("ALTER TABLE users DROP COLUMN groups")
```
cvrysanek: `ALTER TABLE users DROP COLUMN groups` is wrapped in a bare `except Exception` that only `logger.warning`s on failure. The DEPRECATED comment at the top of `_SYSTEM_SCHEMA` says "DuckDB ALTER DROP COLUMN may be blocked by historic FKs … legacy columns are NULL-ed in the migration." — but the code here doesn't NULL the column on failure, it leaves whatever was there. Combined with `has_groups_col` only being read once at the top, a failed drop on this run will not retry on the next start (column still present, but the v13 backfill has already shipped membership into `user_group_members`).

Either:
1. `UPDATE users SET groups = NULL` before the drop attempt so a failed-drop install at least matches the documented contract, or
2. Re-attempt the drop on every `_ensure_schema` call when the column is detected on a v13 DB (cheap; succeeds on the first DuckDB version that can do it).

app/api/scripts.py
```diff
@@ -41,7 +41,7 @@ class ScriptResponse(BaseModel):
 
 @router.get("")
 async def list_scripts(
-    user: dict = Depends(require_role(Role.ADMIN)),
+    user: dict = Depends(require_admin),
```
cvrysanek, line 44: Tightening the Script API to admin-only is the right closure for #44. Two follow-ups for the CHANGELOG / migration notes:

1. The previous gate was `require_role(Role.ADMIN)`, which under the v9 model accepted `core.admin` *or* anything with the `admin` implies. Under v13 `require_admin` is strict Admin-group membership. Anyone holding a v9 admin grant via group_mapping that didn't get carried into Admin-group membership by the v13 finalize will get 403 here. The finalize backfill (line 851) only carries `core.admin` direct grants, not group-mapped ones.
2. `/api/scripts/{id}/run` and `/api/scripts/run` are common cron / scheduler-side targets. If any operator wired a non-admin PAT into a scheduler that hits these, this PR breaks them silently. Worth a CHANGELOG line with the migration recipe ("add the PAT user to the Admin group").

docs/TODO-rbac-data-enforcement.md
```diff
@@ -0,0 +1,130 @@
+# TODO — RBAC data enforcement (drop `dataset_permissions`)
```
cvrysanek, line 1: Calling this out as a release-quality concern, not a doc nit: the `/admin/access` UI now lets admins grant `ResourceType.TABLE` to groups, but `src/rbac.py:can_access_table` is unchanged and still consults `dataset_permissions`. End-state for an admin who reads the UI:

1. Admin grants "engineering" group access to `bucket.foo` via `/admin/access`.
2. User in engineering hits `/api/catalog`, gets filtered out / 403.
3. Admin opens a ticket: "the grant doesn't work."

Two options before merge:
- Land step 1 of this TODO (refactor `can_access_table` to delegate to `app.auth.access.can_access`) inside this PR. Doc says it preserves the external signature, so the blast radius is contained.
- Or hide the `ResourceType.TABLE` chip on `/admin/access` behind a feature flag (env var; default off) so the UI doesn't claim a capability the runtime doesn't honor.

Leaving it as-is is also defensible if the gap is documented in the CHANGELOG and surfaced in the UI itself ("Table grants are not yet enforced; tracked in #N"), but quietly shipping the half-built feature is the worst of the three.

docker-compose.host-mount.yml
```diff
@@ -17,5 +23,5 @@ volumes:
     driver: local
     driver_opts:
       type: none
-      o: bind
+      o: bind,rbind
```
cvrysanek, line 26: Good fix. The dual-disk silent-shadowing failure mode is exactly the kind that survives staging because the test VM is single-disk. Worth a follow-up: a one-shot startup probe that asserts a known sub-mount path is non-empty inside the container (or that `findmnt /data/state` reports a separate device), so the next regression — someone reverts to plain `bind`, or adds `/data/state` as a third disk and forgets — surfaces at boot instead of after the next sync.

app/marketplace_server/packager.py
```diff
@@ -0,0 +1,171 @@
...
+    """Return a JSON-serializable summary for diagnostic / admin endpoints.
+
+    Mirrors the PoC's /marketplace/info contract.
+    """
+    plugins = marketplace_filter.resolve_allowed_plugins(conn, user)
+    etag = marketplace_filter.compute_etag(plugins)
+    return {
+        "user_id": user.get("id"),
+        "email": user.get("email"),
+        "groups": marketplace_filter.resolve_user_groups(user),
```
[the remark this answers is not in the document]
minasarustamyan: Fixed in 283a650. Defined marketplace_filter.resolve_user_groups(conn, user) returning the names of groups the user belongs to (sorted alphabetically), and updated both call sites in app/marketplace_server/packager.py to pass conn through. The diagnostic `groups` field in /marketplace/info JSON and the ZIP's .agnes/version.json now reflects actual membership, populated via _user_group_ids so Everyone is implicit (matching the RBAC filter semantics in resolve_allowed_plugins).

app/api/users.py
```diff
@@ -90,31 +124,65 @@ def _to_response(
...
+    admin_group = conn.execute(
+        "SELECT id FROM user_groups WHERE name = ?", [SYSTEM_ADMIN_GROUP],
+    ).fetchone()
+    if not admin_group:
+        return
+    members = UserGroupMembersRepository(conn)
+    if is_admin:
+        members.add_member(user_id, admin_group[0], "admin", actor_email)
+    else:
+        members.remove_member(user_id, admin_group[0])
```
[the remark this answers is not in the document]
minasarustamyan: Asymmetry is intentional — _set_admin_membership backs the CLI override gesture (`da admin set-role`), while the source-scoped delete in app/api/access.py:480 backs the per-row Remove buttons in /admin/access where the frontend already filters by source.

Rather than locking this in with a docstring, we're tracking the broader role-based legacy cleanup as a TODO in docs/TODO-rbac-data-enforcement.md (added in 283a650), gated on Google Workspace group sync being verified reliable in production. This asymmetry gets resolved together with the rest.

src/db.py
minasarustamyan: Done in 283a650 — wrapped the body in BEGIN TRANSACTION / COMMIT / ROLLBACK. Per-step soft-fails on DROP TABLE / ALTER (already caught inline) do not abort the transaction — only an unhandled exception from a backfill SELECT or INSERT does. The outer caller in _ensure_schema then skips the schema_version bump on line 1156 and the next start retries the whole v12→v13 step.

One nit on the original framing: the schema_version bump doesn't actually fire on a thrown exception in the current code either — Python exception propagates and skips the UPDATE on line 1154-1157. So "half-migrated DB with version=13" wasn't the failure mode in practice; "half-migrated DB with version=12, retried on next start" was, and that's now strictly better with the explicit transaction.

src/db.py
minasarustamyan: Verified by git history grep that no production code in this repo ever called register_internal_role to register non-core roles — only test fixtures used the API. So in any direct-deploy install, internal_roles contains only the 4 seeded core.* rows and the v13 backfill handles core.admin correctly; the lower-tier core.* roles (viewer/analyst/km_admin) collapse to "non-admin" in the v13 binary model, which matches the intended semantics.

Added defensive warning log in b4ec4c4 — before the legacy DROP step, the migration emits a warning naming each non-core role key + grant count it's about to drop, so forked installs that did register custom capability roles via register_internal_role get visible breadcrumbs instead of silent loss. In our deploy this is a no-op (zero matching rows).

src/db.py
minasarustamyan: Added logger.debug at all four sites in b4ec4c4 (steps 2/3/4/5 in _v12_to_v13_finalize). Steps 3 and 4 explicitly note in the log that a system_seed insert was skipped because step 2 already wrote a google_sync row for the same (user, group) pair — so the source conflict you flagged is observable rather than hidden. Step 5 (resource_grants) logs at the same level for symmetry, even though its UNIQUE constraint is on (group_id, resource_type, resource_id) without the source semantics.

Did not reorder the steps to make system_seed win on collision: in this deploy users.groups was never populated (Workspace sync was never functional → fetch_user_groups always returned [] → set_groups was never called), so the collision case is empty. For OSS forks where the JSON column had data, the broader role-based cleanup pass (tracked in TODO-rbac-data-enforcement.md, gated on Workspace sync being verified) is the right place to revisit.

The runtime add_member in src/repositories/user_group_members.py shares the same except: pass pattern. That's a separate concern about admin actions being silently no-op'd against fragile google_sync rows — not changing it in this PR.

src/db.py
minasarustamyan: Non-issue in this deploy. The users.groups JSON column was added in v12 (commit 11a8b70) but populated only by UserRepository.set_groups, which is called from the Google OAuth callback when fetch_user_groups returns a non-empty list. In this deploy Google Workspace group sync was never functional (Cloud Identity API / SA role / scope setup deferred), so fetch_user_groups always returned [] and set_groups was never invoked. Result: the column is NULL across all rows → step 2 of v12→v13 finalize (SELECT … WHERE groups IS NOT NULL) finds zero rows → step 7 DROP proceeds against an empty column with no FKs and succeeds cleanly.

The defensive UPDATE … SET groups = NULL before DROP and the per-connect retry hook are good ideas for forked installs that DID wire up Workspace groups under v12, but the risk is too narrow to justify expanding this PR's scope. If an OSS consumer reports the issue we'll fix it as a focused follow-up. Not changing the code here.

app/api/scripts.py
minasarustamyan: Both follow-ups handled in b4ec4c4:

1. v9 admin via group_mapping — tracked as a bullet in docs/TODO-rbac-data-enforcement.md under Open questions. In this deploy the set is empty (Google Workspace group sync was never functional → group_mappings table never populated → no admin promotions came through that path), so the loss is theoretical here. For OSS forks that did wire up Workspace groups under v9, those admins need to be re-promoted via /admin/access post-upgrade — flagged for the broader role-based cleanup pass once Workspace sync is verified end-to-end.

2. Cron / scheduler PAT migration recipe — appended to the Script API admin-only entry in CHANGELOG.md ("Migration for cron / scheduler PATs: add the PAT user to the Admin group via /admin/access or `da admin group add-member Admin <pat-user-email>`. PATs themselves do not need re-issuing — group membership is read at request time.").

docs/TODO-rbac-data-enforcement.md
minasarustamyan: Acknowledged. The half-built nature is intentionally disclaimed in the CHANGELOG entry for ResourceType.TABLE ("Listing and grant storage only — runtime enforcement still flows through legacy dataset_permissions; the migration plan lives in docs/TODO-rbac-data-enforcement.md") and tracked as step 1 in that TODO doc (refactor src/rbac.py:can_access_table to delegate to app.auth.access.can_access).

Not landing the runtime cutover in this PR — it's part of the broader role-based cleanup pass that's gated on Google Workspace group sync being verified end-to-end (also tracked in the same TODO). Once that's confirmed reliable, the table-grant runtime enforcement lands together with the rest of the role legacy strip.

This PR is already large (14 commits, 150 files) and deferring this specific item to a focused follow-up PR keeps the scope bounded; the gap is documented in both the changelog and the TODO so admin operators reading either know the table-grant chip on /admin/access is currently listing-only.

docker-compose.host-mount.yml
minasarustamyan: Good catch. Agreed it's worth a follow-up startup probe — would ideally assert findmnt /data/state reports a separate device on dual-disk hosts (or that a known sub-mount path is non-empty inside the container), with a clear log line at boot instead of silent write-to-wrong-disk discovered after the next sync.

Not landing in this PR (already large; the rbind fix itself is correct and tested on the dual-disk host that surfaced the original regression). Tracking via this thread for a focused operational follow-up PR.

src/repositories/user_groups.py
```diff
@@ -0,0 +1,147 @@
...
+        description: Optional[str] = None,
+    ) -> None:
+        # Block mutation of system groups — name/description are seeded and
+        # callers must not be able to rename "Admin" / "Everyone" out from
+        # under the marketplace filter.
+        existing = self.get(group_id)
+        if existing and existing.get("is_system"):
+            raise SystemGroupProtected(
+                f"group {existing.get('name')!r} is a system group and cannot be modified"
+            )
```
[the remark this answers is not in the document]
minasarustamyan: Fixed in 62b238c. Aligned the API endpoint with the repository's unconditional behavior — system groups are now immutable end-to-end. PATCH /admin/groups/{id} on a system group returns 409 "System groups are immutable" regardless of which field the payload tries to change, matching the admin UI which already hides the Edit button for is_system rows (admin_group_detail.html:148).

The original endpoint comment ("Allow description edits but not rename") encoded a contract that the repository never honored and the UI never exposed — so the practical impact was zero (no supported flow ever hit it), but the misleading error message made direct API callers think rename was the issue when only description was changing. Picked the "immutable" direction for consistency: a future deliberate decision to allow description edits would mean loosening repo + adding UI button + documenting, all in one go.

app/marketplace_server/router.py
```diff
@@ -0,0 +1,56 @@
...
+async def marketplace_zip(
+    request: Request,
+    user: dict = Depends(get_current_user),
+    conn: duckdb.DuckDBPyConnection = Depends(_get_db),
+) -> Response:
+    if_none_match = request.headers.get("if-none-match", "").strip().strip('"')
+    data, etag = packager.build_zip(conn, user)
+
+    if if_none_match and if_none_match == etag:
+        return Response(status_code=304, headers={"ETag": f'"{etag}"'})
```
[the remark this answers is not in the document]
minasarustamyan: Acknowledged. The split (compute_etag_only → compare → build_zip) is the right shape — on a 304 response it saves a second pass of file reads + ZIP compression on top of the SHA256 pass that compute_etag already did. Worth more once the SessionStart hook is wired up on real clients hitting the endpoint per session.

Not landing in this PR — it's a perf optimization, not a correctness issue, and we want to keep this PR's surface bounded for merge. Tracking via this thread for a focused operational follow-up PR.

app/api/access.py
```diff
@@ -0,0 +1,852 @@
...
+        raise HTTPException(status_code=404, detail="Group not found")
+    if g.get("is_system"):
+        raise HTTPException(status_code=409, detail="Cannot delete a system group")
+    try:
+        repo.delete(group_id)
+    except SystemGroupProtected:
+        raise HTTPException(status_code=409, detail="Cannot delete a system group")
+    # Cascade members + grants for this group so dangling references go away.
+    conn.execute("DELETE FROM user_group_members WHERE group_id = ?", [group_id])
+    conn.execute("DELETE FROM resource_grants WHERE group_id = ?", [group_id])
```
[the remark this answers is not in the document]
minasarustamyan: Fixed in 62b238c. Wrapped DELETE /admin/groups/{id} in an explicit BEGIN TRANSACTION / COMMIT / ROLLBACK and reordered the cascade so child rows (user_group_members + resource_grants) delete before the group row inside the transaction. Either all three deletes commit or rollback restores all three together — no orphan-row scenario, no UI cleanup gap.

Same transaction pattern as the v12→v13 finalize wrap from the previous round (cvrysanek review on src/db.py:778), kept consistent across the two atomicity-sensitive paths in this PR.

src/db.py
```diff
@@ -637,6 +732,251 @@ def get_analytics_db_readonly() -> duckdb.DuckDBPyConnection:
...
+                        [user_id, admin_group_id],
+                    )
+                except duckdb.ConstraintException:
+                    logger.debug(
+                        "v13 backfill step 3 (admin system_seed): skipped "
+                        "insert for user=%s — already in Admin group "
+                        "(possibly from step 2 google_sync of 'Admin' "
+                        "Workspace group; system_seed intent is dropped)",
+                        user_id,
+                    )
```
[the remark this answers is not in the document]
minasarustamyan: Tracked together with cvrysanek's identical concern (review on src/db.py:851 from the previous round). Both replies point at the same TODO bullet added in 283a650 to docs/TODO-rbac-data-enforcement.md under Open questions:

> v9 admins via group_mapping not carried to v13. The v12→v13 finalize backfill at src/db.py:851 matches core.admin grants in user_role_grants only — direct grants. […] In this deploy the set is empty (Google Workspace group sync was never functional → group_mappings table was never populated → no admin promotions came through that path), so the loss is theoretical here. For forked installs that wired up Workspace groups under v9, those admins need to be re-promoted via /admin/access post-upgrade.

Thanks for the second pair of eyes — the TODO doc captures the forked-install impact and is the right place to track the fix (gated on Workspace sync verification, which the broader role-based cleanup pass also depends on).

app/api/marketplaces.py
```diff
@@ -0,0 +1,407 @@
...
+    try:
+        conn.execute(
+            "DELETE FROM marketplace_plugins WHERE marketplace_id = ?",
+            [marketplace_id],
+        )
+        conn.execute(
+            "DELETE FROM resource_grants "
+            "WHERE resource_type = ? AND resource_id LIKE ? || '/%'",
+            [ResourceType.MARKETPLACE_PLUGIN.value, marketplace_id],
+        )
```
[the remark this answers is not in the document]
ZdenekSrotyr: Fixed in f67305d. Switched the resource_grants cleanup to `starts_with(resource_id, ? || '/')` per the suggestion — literal-prefix match with no wildcard semantics, so a slug like `my_mkt` no longer collides with `myXmkt/plugin` siblings. The slug-validation regex (`[a-z0-9][a-z0-9_-]{0,63}`) is unchanged; the bug was the DELETE side trusting LIKE on a charset that legally includes `_`.

app/api/users.py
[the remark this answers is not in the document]
ZdenekSrotyr: Same finding as round 1 (review on app/api/users.py:143 from 2026-04-28T07:40:51Z, my reply at 09:43:52Z) — re-raised with sharper framing this round. The asymmetry is intentional: `_set_admin_membership` backs the explicit operator gesture (`PATCH /api/users/{id}` role change, `da admin set-role`), where the operator's stated intent is "this user is no longer an admin" regardless of how they got the badge. The source-scoped DELETE in app/api/access.py:497/748 backs the per-row Remove buttons in /admin/access where the UI already filters by source. The two paths have different semantics by design.

The `system_seed` re-add on next start is the documented behavior of `SEED_ADMIN_EMAIL` — it's the bootstrap escape hatch, intentional. The `google_sync` row deletion is the trade-off we accept for the role-change gesture being absolute.

Tracked as a follow-up bullet in docs/TODO-rbac-data-enforcement.md alongside the legacy-role cleanup work — if we decide later that role-change should be source-scoped (refusing to demote system_seed admins, surfacing google_sync provenance to the admin), that's a separate UX decision, not a bug fix.

app/marketplace_server/git_backend.py
```diff
@@ -0,0 +1,181 @@
...
+            shutil.rmtree(tmp, ignore_errors=True)
+        else:
+            raise
+    except OSError as e:
+        # Windows: rename fails with WinError 183 if target exists. Same outcome.
+        if target.is_dir():
+            shutil.rmtree(tmp, ignore_errors=True)
+        else:
+            raise RuntimeError(f"git-cache rename failed: {e}") from None
+    return target
```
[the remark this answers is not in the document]
ZdenekSrotyr: Acknowledged — and explicitly disclaimed in the file's docstring (lines 13-15: "Stale entries are never pruned in this iteration — a different content ETag just materializes a new directory next to the old one. First iteration of prune logic is deferred; see plan 'Out of scope'.")

For the canonical deployment (single-tenant, tens of plugins, infrequent content changes) this is a non-issue — the cache reaches a steady state of a few `<etag>.git/` directories per distinct RBAC view and stops growing. For installs where plugin content churns or there are many distinct group → grant configurations, prune is a real follow-up. Tracked in the same TODO doc; the natural design is a TTL-based sweep on `os.path.getmtime` plus a "never prune the currently-active etag" guard.

app/marketplace_server/git_router.py
```diff
@@ -0,0 +1,156 @@
...
+
+            return _CloseOnExhaust(inner, repo)
+        finally:
+            # The DB cursor can be closed early — ensure_repo_for_user and
+            # resolve_token_to_user are done with it by now.
+            if conn is not None:
+                try:
+                    conn.close()
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
ZdenekSrotyr: Verified safe. `get_system_db()` (src/db.py:324-346) returns `_system_db_conn.cursor()` — a DuckDB **cursor**, not the underlying singleton connection. The docstring explicitly calls this out (line 329): "Returns a cursor so callers can safely close() it without closing the underlying connection." The WSGI handler's `finally: conn.close()` closes the per-request cursor; the process-wide `_system_db_conn` singleton stays open and the FastAPI dependency `_get_db` follows the exact same pattern. No race, no concurrent-request interference.

services/scheduler/__main__.py
```diff
@@ -96,23 +117,46 @@ def _call_api(endpoint: str, method: str = "POST") -> bool:
...
-        now = time.time()
-        for name, interval, endpoint, method in JOBS:
-            if now - last_run[name] >= interval:
-                logger.info(f"Running job: {name}")
-                _call_api(endpoint, method)
-                last_run[name] = now
-        time.sleep(10)  # check every 10 seconds
+        now_iso = datetime.now(timezone.utc).isoformat()
+        for name, schedule, kind, target in JOBS:
+            if not is_table_due(schedule, last_run[name]):
```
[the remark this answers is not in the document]
ZdenekSrotyr: Verified non-issue. `is_table_due` (src/scheduler.py:47-99) is purely timestamp-based — it takes `(schedule: str, last_sync_iso: str | None, now: datetime | None)` and decides only on (a) interval-elapsed math (`every Xm/Xh`) or (b) daily-time crossover (`daily HH:MM[,HH:MM]`). No `table_registry` lookup, no table-name-aware branching, no DB access. The name is historical (table-sync was the first caller); the contract is generic schedule-evaluation. Reuse for the marketplaces job is safe.

The schedule formats Devin flagged as 'support assumed' (`every 15m`, `daily 03:00`) are both directly handled — interval branch at line 80-89, daily branch at line 91-99.

app/api/access.py
```diff
@@ -0,0 +1,887 @@
...
+        # admin UI hides the Edit button entirely (admin_group_detail.html).
+        # Reject any modification at the API layer with a clear message
+        # rather than the misleading repository error that surfaced when
+        # the previous "rename rejected, description allowed" contract
+        # diverged from the repository's "all mutations rejected"
+        # behavior.
+        raise HTTPException(
+            status_code=409,
+            detail="System groups are immutable",
+        )
```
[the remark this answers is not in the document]
ZdenekSrotyr: Fixed in 0c963b5. Endpoint guard at `app/api/access.py:331` is now narrow: 409 only when `payload.name is not None and payload.name.strip() != g["name"]`. Same-name no-op renames are dropped from the update dict before reaching the repo. Description-only `PATCH` flows through to the (already-narrowed by 7147bac) repo layer.

Sanity: `pytest tests/test_repositories.py tests/test_resource_types.py tests/test_auth_providers.py` → 89 passed (the new `test_update_system_group_description_allowed` + `test_update_system_group_same_name_no_op` from PR #110 now exercise the full endpoint→repo path). CHANGELOG bullet 34 rewritten to describe the actual contract.

app/auth/providers/google.py
```diff
@@ -174,65 +104,39 @@ async def google_callback(request: Request):
...
+                group_names = fetch_user_groups(email)
+                ug_repo = UserGroupsRepository(conn)
+                members_repo = UserGroupMembersRepository(conn)
+                group_ids: list[str] = []
+                for group_name in group_names:
+                    g = ug_repo.ensure(group_name)
+                    group_ids.append(g["id"])
+                members_repo.replace_google_sync_groups(
+                    user["id"], group_ids, added_by="system:google-sync",
+                )
```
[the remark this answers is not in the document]
ZdenekSrotyr: Fixed in 0c963b5. Callback now guards `replace_google_sync_groups` behind `if group_names:` and logs "preserving existing memberships" on the empty path. Trade-off documented inline + in CHANGELOG: a user whose Workspace groups were genuinely cleared keeps stale memberships until the next non-empty sync.

Cleanest long-term fix is to give `fetch_user_groups` a tri-state return (None on error, [] on empty success, [...] on success-with-groups), but that ripples through every caller and test in tests/test_group_sync.py — out of scope for this PR. Tracked.

The previous regression-guard test (`test_callback_empty_groups_does_not_overwrite_existing`) is on the v12-skipped `TestGoogleCallbackGroupSync` class — it asserts `users.groups` JSON which v13 dropped. Rewriting it for `user_group_members` is the right next step but separate from the bug fix.

PR #110, opened 2026-04-28
PR #106 follow-up: Devin r2 + gemini blockers (FK schema v14, break-glass CLI, TABLE feature flag)

cvrysanek:
Follow-up to #106. Targets `ma/staging` so it rides into the same review cycle.

## What this PR does

Addresses **Devin round-2** findings (left unresolved on `b4ec4c4`) plus **gemini-flagged blockers** surfaced via a second-opinion review pass on the round-1+round-2 fixes.

CI on the parent PR is green; this PR's diff is additive on top of `ma/staging`.

---

### 🔴 system group description PATCH returns misleading 409

`src/repositories/user_groups.py:111-148` — `update()` previously raised `SystemGroupProtected` for *any* mutation on `is_system=TRUE` groups. Endpoint at `app/api/access.py:328-335` was supposed to allow description edits while blocking renames; the repo override caught everything and surfaced as `409 "Cannot rename a system group"` on a description-only PATCH.

Fix: narrow the repo guard to actual rename only (`name is not None and name != existing["name"]`). Endpoint pre-check stays as defense-in-depth.

Tests: `test_update_system_group_description_allowed`, `test_update_system_group_same_name_no_op` in `tests/test_repositories.py`.

### 🟡 marketplace ZIP built before If-None-Match check

`app/marketplace_server/{packager,router}.py` — `build_zip()` ran the DB query, SHA256-hashed every plugin file, collected file bytes, and ran `ZIP_DEFLATED` *before* the etag comparison. Wasted on every 304 cache-hit. Endpoint is consumed by Claude Code SessionStart on every session launch.

Fix:
- New `compute_etag_for_user(conn, user) → (etag, plugins)` that resolves the allowed plugin set + computes the content-addressed etag, no file collection.
- Router resolves etag first; `If-None-Match` short-circuits to `304` before any file I/O.
- `build_zip()` accepts pre-computed `plugins` + `etag` kwargs so the post-304 path doesn't re-resolve.
- In-process `cachetools.TTLCache` (default `ttl=120s`, env-tunable via `AGNES_MARKETPLACE_ETAG_TTL`, set `0` to disable). Cache key: sorted `(prefixed_name, version, plugin_dir)` tuples — two users with the same allowed view share the same entry.
- `src/marketplace.py:sync_marketplaces()` calls `invalidate_etag_cache()` after a successful sync so freshly-synced content is visible immediately rather than waiting for TTL.
- `cachetools>=5.3.0` declared as a direct dep (was transitive via `google-auth`).

### 🟡 group delete cascade after parent delete = orphan risk

`app/api/access.py:delete_group` — wrapped cascade DELETEs + `repo.delete(group_id)` in `BEGIN TRANSACTION / COMMIT / ROLLBACK`. Catches `SystemGroupProtected` and unhandled exceptions both rollback. Defense-in-depth — combined with the new schema v14 FKs (below), DuckDB now refuses the parent delete if children exist.

### 🚩 group_mapping admins not backfilled on v12→v13

Already disclaimed in `docs/TODO-rbac-data-enforcement.md` (added in `b4ec4c4`). No code change in this PR.

---

## Gemini second-opinion blockers

Gemini was run as an adversarial review on `b4ec4c4` + the proposed Devin-r2 patches; its three BLOCKERS are landed here.

### Schema v14 — DB-level FK on `user_group_members` + `resource_grants`

`src/db.py` — gemini's "harden the database" finding: relying on every API caller to remember to clean up two tables in a transaction is fragile.

- `_SYSTEM_SCHEMA`: `user_group_members.group_id` and `resource_grants.group_id` now `REFERENCES user_groups(id)`. Fresh installs get the FK on first `CREATE TABLE IF NOT EXISTS`.
- DuckDB does **not** support `ON DELETE CASCADE` (verified — `ParserException: FOREIGN KEY constraints cannot use CASCADE, SET NULL or SET DEFAULT`). The app-layer cascade order in `delete_group` remains required; the FK turns any future omission into a `ConstraintException` at the DB level instead of a silent orphan.
- `_v13_to_v14_finalize()`: pre-clean orphans (with WARN-level log of dropped counts) → `BEGIN TRANSACTION` → `RENAME TO *_v13_pre` → `CREATE TABLE` with FK → `INSERT ... SELECT` → `DROP *_v13_pre` → `COMMIT`. Wrapped in `try/except → ROLLBACK` so a mid-flight failure leaves the DB on a clean v13 state and the next start retries.
- `SCHEMA_VERSION = 14`.

Smoke-tested:
```python
INSERT INTO user_groups (id,name) VALUES ('g1','TestGroup');
INSERT INTO user_group_members (user_id,group_id,source) VALUES ('u1','g1','admin');
DELETE FROM user_groups WHERE id='g1';
# → ConstraintException: Violates foreign key constraint
```

### Break-glass admin CLI

`cli/commands/admin.py` — gemini called this out as a non-negotiable: the marketplace dropped admin god-mode but PR 106 ships no out-of-band path to restore admin if the last live Admin-group member is removed by a race / accident / nightly Workspace sync mismatch.

`da admin break-glass grant-admin <email>`:
- Opens `system.duckdb` directly via `src.db.get_system_db` — no HTTP, no `require_admin`, no FastAPI auth dep.
- Creates the user with `users.create()` if missing (auto-joins them to Everyone via the existing repo path).
- Inserts `user_group_members(user_id, Admin_id, source='cli_break_glass', added_by='cli:break-glass')` so audits distinguish this from `admin` / `google_sync` / `system_seed`.
- Confirmation prompt by default; `--yes`/`-y` to skip for scripted recovery.
- Idempotent: returns "already a member" if membership exists.

Smoke-tested end-to-end on a fresh DATA_DIR:
```
$ da admin break-glass grant-admin newadmin@example.com --yes
Created user newadmin@example.com (id=f8ebd4cd…)
Granted Admin to newadmin@example.com. Audit source='cli_break_glass'.
```

### Feature-flag `ResourceType.TABLE` until runtime enforcement lands

`app/resource_types.py` + `app/api/access.py` — gemini called shipping the chip without runtime enforcement "indefensible … fundamentally misleading user experience". Until `docs/TODO-rbac-data-enforcement.md` step 1 lands (refactor `src/rbac.py:can_access_table` to delegate to `app.auth.access.can_access`), an admin who grants `ResourceType.TABLE` via `/admin/access` is told the grant is recorded — but the user still gets filtered out at `/api/catalog` because the runtime still consults legacy `dataset_permissions`.

Fix: `AGNES_ENABLE_TABLE_GRANTS` env var (default off):
- New `is_resource_type_enabled(rt)` + `enabled_resource_types()` helpers.
- `GET /api/admin/resource-types`: filters disabled types out of the chip list.
- `GET /api/admin/access-overview`: skips disabled types in the resources tree.
- `POST /api/admin/grants`: returns **422** with the env-var name in `detail` if a TABLE grant is attempted while disabled. List/delete of existing TABLE rows in `resource_grants` still works (no data lockout for cleanup).

Tests: `TestTableGrantsFeatureFlag` (4 cases — endpoint exclude, endpoint include, grant rejection, grant acceptance) in `tests/test_resource_types.py`. Existing `TestAccessOverviewIncludesTables` opts in via `monkeypatch.setenv`.

### Destructive rollback test for `_v12_to_v13_finalize`

`tests/test_db.py` — gemini's "should is not good enough for a critical atomic migration" finding. The transaction wrap landed in `283a650` had no test that the rollback actually fires.

`test_v12_to_v13_finalize_rollback_on_failure`:
1. Hand-crafts a v12 DB with sample admin / non-admin users, internal_roles, user_role_grants, plugin_access.
2. Monkeypatches `_seed_system_groups` (the first call inside the transaction) to raise `RuntimeError`.
3. Calls `get_system_db()`, expects the exception.
4. Asserts `schema_version` stays at 12, all legacy tables (`internal_roles`, `group_mappings`, `user_role_grants`, `plugin_access`) still exist, `user_group_members` + `resource_grants` are empty (transactional inserts rolled back).
5. Restores the real `_seed_system_groups`, calls `get_system_db()` again, asserts the retry succeeds and backfills.

---

## Test results

`pytest tests/ -q` — **1544 passed**, 7 pre-existing failures (verified via `git stash` against baseline `b4ec4c4` — all in older v1/v2/v3 migration ladder tests, hit a DuckDB DDL-on-FK'd-table quirk unrelated to this branch).

New / touched test classes all pass:
- `TestUserGroupsRepository` (3 new cases)
- `TestTableGrantsFeatureFlag` (4 new cases)
- `TestAccessOverviewIncludesTables` (updated to opt-in via env var)
- `test_v12_to_v13_finalize_rollback_on_failure`
- All `tests/test_marketplace_server_*.py` (regression on etag split + cache)

---

## Reviewer hot-spots

- `src/db.py:_v13_to_v14_finalize`: DDL inside `BEGIN TRANSACTION` mirrors the v12→v13 pattern. RENAME → CREATE → INSERT → DROP — verify transactional semantics on your DuckDB version if it differs from 1.5.1.
- `app/marketplace_server/packager.py`: TTLCache key uses `str(plugin_dir)` so a re-clone to a new path invalidates cache; `version` is the cached `marketplace_plugins.version` column, not the on-disk file hash, so a same-version content drift between syncs would only resolve on TTL expiry. `invalidate_etag_cache()` post-sync covers the common case.
- `cli/commands/admin.py break-glass`: this command bypasses RBAC by design. Anyone with shell access to `${DATA_DIR}/state/system.duckdb` can run it. Acceptable since filesystem access already implies host-level trust, but worth being explicit about in deployment docs.
- `AGNES_ENABLE_TABLE_GRANTS`: existing TABLE grants in `resource_grants` from prior dev/test runs stay visible in the list-grants endpoint and are still deletable. The flag controls UI exposure + new grant acceptance only.

---

## Branch + base

Branch: `vr/pr106-followup` ← off `ma/staging` at `b4ec4c4`
Base for this PR: `ma/staging` (so the changes land inside #106's review cycle and ship together to `main`).


ZdenekSrotyr:
Closing as merged-via-cherry-pick.

The full content of `vr/pr106-followup` (commit 01cd597) was cherry-picked onto a squashed `ma/staging` history as commit `7147bac0796d12e8912cfe9007f535742d389319` ("feat(rbac+marketplace): schema v14 FK + AGNES_ENABLE_TABLE_GRANTS + break-glass CLI"), with your authorship preserved (`--author=`). `ma/staging` was force-pushed (with --force-with-lease) to a clean 3-commit shape so PR #106 ships as a coherent unit:

1. `cc7a7db` (squashed) — RBAC v13 + Claude Code marketplace + #81/#83/#44 hardening + UI polish + OSS neutralization
2. `7147bac0796d12e8912cfe9007f535742d389319` — your follow-up: schema v14 FK + feature flag + break-glass CLI + Devin r2 fixes
3. `5c54320` — chore(release): cut 0.12.0 (CHANGELOG combined + pyproject 0.12.0 + cachetools dep)

Conflict resolutions during cherry-pick (your version won):
- `app/api/access.py` — your delete_group transaction wrap (cleaner formatting)
- `app/marketplace_server/{packager,router}.py` — your TTLCache + clean compute_etag_for_user split (replaced Mina's earlier compute_zip_etag stub from `2253285`)
- `pyproject.toml` — combined dulwich + a2wsgi + cachetools (with cachetools as your contribution)

Local E2E verification passed: 1575 tests pass (was 5 fails earlier on a missing google-api-python-client dep that I added back to commit 1's pyproject after losing it during the squash); docker compose smoke confirms schema v14 migration runs cleanly, `da --version` reports 0.12.0, all auth gates respond as expected.

PR #106 now has the new shape; the next push will trigger CI + Devin Review on the squashed head.

PR #146, opened 2026-04-30
fix(ops): fail-fast guard in agnes-auto-upgrade — refuse start if config disk not mounted

cvrysanek:
## Summary

Runtime backstop against a config-disk mount-loss regression: when host-mount propagation unmounts the config disk under the agnes container, the app silently writes user state (DuckDB, marketplaces, session secret) to the OS root disk and loses it on the next container recreate. Boot-time fixes in the consumer infra repo are preventive; this is the runtime backstop in `agnes-auto-upgrade.sh`, which fires every 5 min on every VM via systemd timer.

### Changes

1. **Fail-fast mount guard** (`c08533e`): Before the existing pull/up logic, when `/dev/disk/by-id/google-config-disk` exists on the VM:
   - Up to 3 mount-and-verify attempts with backoff (2s, 4s, 6s).
   - After the loop, assert `findmnt` source matches the config disk. On mismatch: `logger -t agnes-auto-upgrade FATAL` + `exit 1`.
   - Once verified: re-apply `mount --make-rprivate /data /data/state` on every run.

2. **Vendor-agnostic comment + changelog** (`ab3ff9b`): Cleaned up comments and added changelog entry.

3. **Fix Devin Review finding** (`862d03d`): Changed `mount ... 2>&1 || true` → `mount ... 2>/dev/null || true` on the retry-loop mount command (line 33). The `2>&1` redirect merged stderr into stdout but suppressed nothing, inconsistent with every other error-tolerant command in the script (`umount` at line 31, `mount --make-rprivate` at lines 48–49, `docker compose pull` at line 68). Since this script runs every 5 min via systemd timer, transient mount errors during the retry loop would produce noisy output.

## Review & Testing Checklist for Human

- [ ] Verify on a VM with config disk: `sudo agnes-auto-upgrade.sh` mounts `/data/state` correctly and proceeds to `docker compose pull`
- [ ] Verify on a VM without config disk: script skips the guard block entirely
- [ ] Verify mount failure scenario: temporarily block mount (e.g. wrong device) → confirm script exits with FATAL log, no `docker compose` runs

### Notes

- VMs without a config disk attached (single-disk legacy deployments) skip the whole block via the `if [ -e $CONFIG_DEVICE ]` guard.
- This file is fetched by the consumer infra repo's `startup.sh` from `raw.githubusercontent.com` on every boot — once merged to main, all VMs pick up the new script automatically.

Link to Devin session: https://app.devin.ai/sessions/cceaf2e4367e447da722603ca1be6eff
Requested by: @ZdenekSrotyr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Fixed
+- `scripts/ops/agnes-auto-upgrade.sh`: fail-fast guard before any `docker
+  compose` action — when the VM has a config disk attached
+  (`/dev/disk/by-id/google-config-disk` exists), `/data/state` MUST be backed
+  by it. Three retry attempts with backoff, then exit non-zero. Prevents the
+  silent regression where docker host-mount propagation unmounts the config
+  disk and the app writes user state (DuckDB, marketplaces, session secret)
+  onto `/data` (sdb) — wiped on the next container recreate. Re-applies
+  `mount --make-rprivate /data /data/state` on every run to defend against
+  propagation regressions.
+
 ## [0.24.0] — 2026-04-30
 
 ### Changed
@@ -134,15 +145,6 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   before.
 
 ### Fixed
-- `scripts/ops/agnes-auto-upgrade.sh`: fail-fast guard before any `docker
-  compose` action — when the VM has a config disk attached
-  (`/dev/disk/by-id/google-config-disk` exists), `/data/state` MUST be backed
-  by it. Three retry attempts with backoff, then exit non-zero. Prevents the
-  silent regression where docker host-mount propagation unmounts the config
-  disk and the app writes user state (DuckDB, marketplaces, session secret)
-  onto `/data` (sdb) — wiped on the next container recreate. Re-applies
-  `mount --make-rprivate /data /data/state` on every run to defend against
-  propagation regressions.
 - Removed rogue module-level `logging.basicConfig` from `app/api/sync.py` that
   was reconfiguring root logger every time the api module was imported.
 - `RequestIdMiddleware` rewritten as a pure ASGI middleware (was
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,15 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   before.
 
 ### Fixed
+- `scripts/ops/agnes-auto-upgrade.sh`: fail-fast guard before any `docker
+  compose` action — when the VM has a config disk attached
+  (`/dev/disk/by-id/google-config-disk` exists), `/data/state` MUST be backed
+  by it. Three retry attempts with backoff, then exit non-zero. Prevents the
+  silent regression where docker host-mount propagation unmounts the config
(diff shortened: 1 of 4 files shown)


cvrysanek:
**Post-merge architectural review** — flagging a concern with the curl-from-main pattern this PR shipped.

The fail-fast guard itself (the original c08533e commit) is good and now landed across all 8 operator VMs (manually scp'd as a hotfix while this PR was in flight). The concern below is specifically about commit 92ce7cb (`single-source agnes-auto-upgrade.sh via curl from main`) and the resulting deployment pattern.

## What the PR did

`infra/modules/customer-instance/startup-script.sh.tpl` previously inlined `agnes-auto-upgrade.sh` as a heredoc. The PR replaced that with:

```bash
SCRIPT_URL="https://raw.githubusercontent.com/keboola/agnes-the-ai-analyst/main/scripts/ops/agnes-auto-upgrade.sh"
curl -fsSL --retry 3 --retry-delay 2 "$SCRIPT_URL" -o /usr/local/bin/agnes-auto-upgrade.sh
```

The stated motivation — drift prevention — is real. The inline heredoc had fallen 3 iterations behind canonical (missing TLS overlay detection, array-form compose files, the new fail-fast guard). Centralizing fixes that.

## Two new problems introduced

**1. Split-brain version pinning.** `AGNES_TAG` already pins the docker image to an immutable digest like `stable-2026.04.516`. The host-side cron driver now bypasses that pinning and tracks `main` instead. A customer running `stable-2026.04.516` (pinned for stability reasons — they might have explicitly avoided a newer release that broke their workflow) wakes up tomorrow with their host script floating on whatever shipped to `main` overnight. The image and the script can drift between releases — exactly the property tag-pinning is supposed to prevent.

**2. No rollback knob.** Reverting a bad `agnes-auto-upgrade.sh` requires reverting the upstream PR globally. Affects every customer that reboots after the bad commit. No "rollback for me only" path. The previous heredoc approach at least had per-customer rollback via the customer's own terraform `image_tag` pin (since the script was inlined into the startup template, which versions with the module).

A lesser concern: customers with restricted egress (no `raw.githubusercontent.com` reachability) now silently fail every boot. The image is already pulled from a private registry on the same boot — reusing that channel is strictly cheaper than adding a second public-internet dependency.

## What I'd suggest

Bake `agnes-auto-upgrade.sh` into the docker image at a stable contract path (e.g. `/opt/agnes-host-scripts/agnes-auto-upgrade.sh`) and have the customer template extract it via `docker create` + `docker cp` from the same `${AGNES_TAG}` the operator already pinned. The drift-prevention guarantee is preserved by construction (image and script ship from one commit, atomically released, can't drift). And the version-pin guarantee is restored (script tracks AGNES_TAG, not main).

I've drafted this in #149 — PR is small (3 files, +35/-16). Feedback welcome on whether the approach is acceptable before reviewer-bot picks it up. The PR also documents a few follow-ups in the same vein: `tls-fetch.sh` has the same shape; the operator-side `agnes-the-ai-analyst-infra/startup.sh` curls from main symmetrically and needs the same fix.

## Other PR-scope notes (FYI)

- **Release cut on a half-reviewed PR** — commit 775b098 tagged 0.25.0 on this branch before review approval. Project's release ritual (per CLAUDE.md) tags the merged commit, not a PR branch. Tag is in `main` now since the PR landed, but the ordering is uncomfortable — if review had blocked, the tag would have been stranded.
- **Scope creep** — d3b746a (compose.tls.yml unconditional fetch) and 28fd43d (Caddyfile + skip-tls-overlay) are reasonable changes but logically separate from "fail-fast guard for config disk". Mixing them slows review and makes git-bisect harder. Future PRs would benefit from being split.

None of this is blocking — happy to keep the conversation open in #149.

PR #172, opened 2026-05-04
fix(cli): Windows console crashes on da skills list / da sync

minasarustamyan:
cp1250 (cs-CZ default) and cp1252 consoles cannot encode the Braille spinner glyphs Rich uses for progress bars, nor decode UTF-8 chars in skill markdown — both crash before any command-level code runs.

- cli/main.py: reconfigure stdout/stderr to UTF-8 with errors="replace" on sys.platform == "win32", at import time so Rich's legacy-Windows render path emits decodable bytes.
- cli/commands/skills.py: pass encoding="utf-8" to read_text() (root cause of `da skills list` UnicodeDecodeError).
- cli/{config,update_check,snapshot_meta}.py and cli/commands/{analyst, setup}.py: defensive encoding="utf-8" on all remaining read_text / write_text / open() calls — these wrote ASCII-only JSON in practice but were one non-ASCII value away from the same failure mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,20 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Fixed
+- **Windows: `da skills list` and `da sync` no longer crash on cp1250 consoles.**
+  On a Czech-locale Windows console (default codepage `cp1250`), Rich's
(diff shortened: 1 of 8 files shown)


ZdenekSrotyr:
Hi @minasarustamyan — superseded by #173, which ports your fix forward into the post-rename `agnes` CLI and broadens it to every text-I/O call site in `cli/` (some of the files this PR patched were renamed in #173, so a straight merge wasn't possible). Closing in favor of that — credit preserved in the commit message and CHANGELOG. Thanks for catching this!

PR #181, opened 2026-05-05
feat(bigquery): bq_query_timeout_ms knob; default 600s (was 90s)

ZdenekSrotyr:
## Summary

Adds `data_source.bigquery.query_timeout_ms` (default 600 000 ms / 10 min) plumbed through every code path that LOADs the DuckDB BigQuery extension. Replaces the extension's built-in 90 s timeout that produced the user-facing `Binder Error: Query execution exceeded the timeout. Job ID: <id>` HTTP 400 from `agnes query --remote`.

## End-to-end verification ✅ (reproduced + fixed within the BQ RPC ceiling)

Reproduced the exact user failure on the prod container by running an uncached heavy aggregation against `bq.product_analytics.S2_session_pageviews` with three different `bq_query_timeout_ms` values:

| `bq_query_timeout_ms` | Outcome | Wall time |
|---|---|---|
| `1` | ❌ `BinderException: Binder Error: Query execution exceeded the timeout. Job ID: …` | 1.0 s |
| `100` | ❌ Same error | 1.1 s |
| `600000` | ❌ Same error | **200.5 s ← BigQuery server-side cap** |

The error wording matches the user's report **exactly**, including the `Job ID:` suffix. Confirmed by reading the extension source — `bigquery_client.cpp:973-980` throws `BinderException("Query execution exceeded the timeout. Job ID: " + job_id)` when `query_request.timeout_ms` (set from `bq_query_timeout_ms` at line 1064-1065) elapses on the BigQuery `jobs.query` RPC.

**This is a partial fix** with a known ceiling:

- ✅ Queries that finish 90 s < t ≤ ~200 s — were failing on prod (extension default 90 s), will succeed after this PR (default 600 s, BQ server cap ~200 s).
- ❌ Queries that take longer than ~200 s — still fail with the same error after this PR. The extension issues a single `jobs.query` RPC and throws on the first incomplete response; it does not poll `getQueryResults` afterwards. A real fix for >200 s queries requires either an upstream PR to `hafenkran/duckdb-bigquery` to add a polling loop, or a server-side wrapper in agnes. Tracked separately.

## Why

The match between the extension's default value (`90000` ms) and the timing of the user's failure (~90 s wall before HTTP 400) was the strongest signal. Extension-source review confirms `bq_query_timeout_ms` is the only knob in this code path. Bumping it past ~200 s costs nothing — BQ server-caps at ~200 s anyway — so `600000` is a forgiving default that gets the maximum out of the RPC's effective ceiling.

## What changes

| File | Change |
|---|---|
| `connectors/bigquery/access.py` | New `apply_bq_session_settings(conn)` helper. Reads the knob via `app.instance_config.get_value`. Called from `_default_duckdb_session_factory` after `LOAD bigquery`. |
| `connectors/bigquery/extractor.py` | Calls the helper after `LOAD bigquery` in the standalone extraction path. |
| `src/orchestrator.py` | Calls the helper after `LOAD bigquery` in the `_remote_attach` ATTACH path (the path `agnes query --remote` actually hits — verified by tracing app/api/query.py → analytics.duckdb → views over `_remote_attach`). |
| `app/api/admin.py` | Exposes `query_timeout_ms` in the `/api/admin/server-config` `known_fields` schema. Hint documents the ~200 s BQ server cap. |
| `config/instance.yaml.example` | Adds the knob as a commented-out option with the same hint. |
| `tests/test_bq_query_timeout.py` | 8 unit tests: default, explicit value, sentinel 0, negative, non-numeric, string-numeric coercion, SET-failure fail-soft, missing app.instance_config. |
| `CHANGELOG.md` | Entry under `[Unreleased]` → Added with the ~200 s caveat. |

## Caveats (operator-facing)

- **BigQuery's `jobs.query` RPC server-side caps wait at ~200 s** regardless of the value passed. The extension throws `BinderException` on the first `complete=false` response — no polling loop. Setting `query_timeout_ms` above ~200 s does no harm but does not extend the effective ceiling beyond the BQ cap. DuckDB emits an informational warning when the value is above its threshold; ignore it.
- For queries that genuinely need >200 s, the PR alone is insufficient — a follow-up is needed.

## What this PR does NOT fix

- Queries that exceed BQ's ~200 s server cap (above).
- The `order_economics` materialize broken-source_query bug — independent root cause (registered with backtick-fully-qualified syntax + wrong project ID); fixed via prod break-glass on 2026-05-05 (DB UPDATE while app stopped). The validator that rejects backticks in `source_query` was added based on a factually wrong premise (its comment claims DuckDB parses backticks first; empirically the BQ extension passes source_query verbatim into `bigquery_query()`'s second arg). A follow-up PR should relax that validator and/or construct materialize SQL programmatically from registry's `bucket+source_table+data_project`.

## Test plan

- [x] `pytest tests/test_bq_query_timeout.py` — 8/8 pass locally
- [x] `pytest tests/test_bigquery_extractor*.py tests/test_bq_materialize.py tests/test_orchestrator.py tests/test_admin_configure_api.py` — 107/107 pass locally (no regressions)
- [x] Verified on prod container: SET propagates, `duckdb_settings()` reflects the value, end-to-end reproduces user's exact `Binder Error: ... Job ID: ...` failure when set low and reaches BQ's ~200 s ceiling when set high.
- [x] Read `hafenkran/duckdb-bigquery` source to confirm: `bq_query_timeout_ms` → `query_request.mutable_timeout_ms()` (line 1064-1065) → BinderException on `complete=false` (line 977-979). Single RPC, no polling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Added
+- **`data_source.bigquery.query_timeout_ms` config knob** (default 600 000 ms = 10 min). The DuckDB BigQuery extension's built-in default of 90 s was too tight for analyst-scale queries against view-backed BQ datasets — `agnes query --remote` would HTTP 400 with `Binder Error: Query execution exceeded the timeout. Job ID: …` whenever the underlying BQ job took longer than 90 s, even though the BQ job itself was healthy. The new knob is applied via `SET bq_query_timeout_ms` after every `LOAD bigquery` on every BQ-touching DuckDB session — the orchestrator's `_remote_attach` ATTACH path (`src/orchestrator.py`), the analytics-DB read-only reattach path (`src/db.py:_reattach_remote_extensions` — the primary `agnes query --remote` request path), the `BqAccess` session factory (`connectors/bigquery/access.py`), and the standalone extractor (`connectors/bigquery/extractor.py`). Sentinel `0` (or non-numeric / unparseable values) leaves the extension default in place so operators on legacy extension versions that don't recognise the setting aren't broken. Configurable via `/admin/server-config` UI. Note: BigQuery's `jobs.query` RPC caps the wait at ~200 s per call regardless of this setting; the extension polls on top so the effective ceiling is the value here but each poll is ~200 s. DuckDB emits an informational warning when the value is set above the BQ RPC cap — operators can safely ignore it.
+
 ### Fixed
 
 - Keboola sync now falls back to the legacy Storage-API client when the DuckDB Keboola extension's per-table scan fails, not just when the initial `ATTACH` fails. Two changes:
--- a/app/api/admin.py
+++ b/app/api/admin.py
@@ -285,6 +285,24 @@ _KNOWN_FIELDS: dict[str, dict[str, dict]] = {
                         "`agnes snapshot create` suggestion. 0 disables the gate. Default 5368709120 = 5 GiB."
                     ),
                 },
+                "query_timeout_ms": {
+                    "kind": "int",
+                    "default": 600000,
+                    "hint": (
+                        "DuckDB BigQuery extension query timeout (milliseconds). Applied "
+                        "via `SET bq_query_timeout_ms` after every `LOAD bigquery` on "
+                        "every BQ-touching DuckDB session (orchestrator remote-view "
+                        "ATTACH, BqAccess factory, standalone extractor). Extension "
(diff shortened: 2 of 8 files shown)


ZdenekSrotyr:
Fixed in commit `f9d6be35` — added `apply_bq_session_settings(conn)` to `src/db.py:_reattach_remote_extensions` right after the BQ secret creation and before ATTACH. The function-level docstring there already calls out that this path runs on every read-only analytics-DB open (i.e. every `agnes query --remote`), which is exactly the user-facing path the timeout was breaking. Devin Review was right that without this the rest of the PR was a partial fix. CHANGELOG entry now lists `src/db.py:_reattach_remote_extensions` alongside the other three sites.

ZdenekSrotyr:

PR #182, opened 2026-05-05
feat(caddy): file_server for parquet downloads — bypass uvicorn

ZdenekSrotyr:
## Summary

A single analyst's multi-GB `agnes pull` holds the only uvicorn worker for the duration of the stream, starving the UI, `/api/health`, and every other API endpoint. Container flips to `unhealthy`. Reproduced live on prod 2026-05-05 with a 6.8 GB `order_economics` pull.

This PR moves the bulk byte transfer to Caddy so the app's worker is never blocked, parallelises the analyst-side download loop so a single `agnes pull` doesn't waste the new server-side concurrency, and ensures the change actually reaches running VMs by extending the auto-upgrade script to watch config files (not just the image digest).

### 1. Caddy `file_server` bypasses uvicorn for parquet downloads

- **New endpoint `GET /api/data/{table_id}/check-access`** — lightweight RBAC probe. Returns 204 if the caller has read access, 403 otherwise. ~1 ms; reuses `can_access_table` from `src.rbac`.
- **Caddyfile** — new `@download` matcher for `^/api/data/([^/]+)/download$`. Uses `forward_auth` to call `/check-access` on every request, then `try_files` + `file_server` to serve the parquet from `/srv/extracts/<source>/data/<id>.parquet`. Anything not matched falls through to the existing app handler.
- **docker-compose.yml** — caddy service mounts the agnes data volume at `/srv:ro` (Caddy's own `/data` is occupied by the `caddy_data` volume, hence `/srv`).
- **Fallback preserved**: legacy `src_data/parquet`, future connectors, and non-Caddy deployments (dev `docker compose up` without `--profile tls`) continue to work via the app handler.

### 2. Per-user parallel pull (`AGNES_PULL_PARALLELISM`, default 4)

- `cli/lib/pull.py`'s download loop was strictly serial — N tables took Σ stream_download(t_i). With (1) above the server can sustain parallel transfers; the client serialisation became the new bottleneck.
- Switch to `ThreadPoolExecutor` capped by `AGNES_PULL_PARALLELISM` (default 4, set 1 to restore pre-PR serial). Drops to serial when `len(to_download) <= 1` to avoid executor overhead in the common single-table case.
- Per-table error semantics preserved via `(tid, entry, err)` tuple — a failure on one parquet doesn't abort the rest of the batch.

### 3. `scripts/ops/agnes-auto-upgrade.sh` watches config files + self-updates

Without this, parts (1) + (2) land in the image but stay inert on long-uptime VMs — auto-upgrade only watches `docker images` digests, so a Caddyfile + compose change in main never reaches running VMs (only fresh boots ran `startup.sh`'s file fetch). Confirmed live during this PR's testing on a dev VM.

- Re-fetches all compose overlays + the Caddyfile from `keboola/agnes-the-ai-analyst@main` on every 5 min tick, hashes them, and triggers `docker compose up -d` recreation when the hash changes — same trigger path as an image-digest change.
- Self-updates `/usr/local/bin/agnes-auto-upgrade.sh` from main on every tick. Without this, the very fix that watches config files would never propagate to existing VMs.
- Fail-soft on curl errors via the `.new + mv` pattern: a 404 / network blip leaves the existing file untouched.

## End-to-end test on a dev VM

Reproduced + verified the full flow on `foundryai-dev-zsrotyr` (a real grpn dev VM) against `https://agnes-zsrotyr.groupondev.com`. Pinned the VM's `AGNES_TAG` to the branch-specific `dev-zs-scale-downloads` (otherwise `dev-zs-latest` floats across all `zs/...` branches and any other contributor's push wins the tag race — caught me mid-test when `zs/keboola-fallback-kbcstorage` overrode `dev-zs-latest` between my push and my test).

| Test | Result |
|---|---|
| `GET /api/data/{id}/check-access` (admin PAT) | HTTP 204, ~0.9 s |
| `GET /api/data/{id}/download` (admin PAT) | HTTP 200, parquet body, ETag `"dias3ar6thwrlu"` (Caddy mtime hex format, **not** the app's `"<mtime_ns>"` — proves Caddy file_server served it directly) |
| Parquet integrity | DuckDB read back with the right schema |
| **3 concurrent downloads + 5 parallel `/api/health` probes** | **all 8 succeeded, downloads ~0.9 s, health ~0.85 s** — no blocking. Pre-PR this would have starved health on a single uvicorn worker. |
| `agnes pull` (real CLI from laptop, 2 tables) | both parquets downloaded, view rebuild succeeded, total wall ~3.8 s |
| Auto-upgrade tick fired manually | detected config drift → recreated containers; the new bind-mount + Caddyfile rules took effect |

## Why not multi-worker uvicorn

Considered and dropped. Blocked by `system.duckdb`'s single-writer DuckDB lock — a second uvicorn worker crashes at startup on `Could not set lock on file`, the same race that pushed the scheduler from in-process writes to HTTP-via-app. Multi-reader workers + single-writer coordination is a real refactor and out of scope here. This PR sidesteps the issue entirely by getting bulk transfer off uvicorn.

## Deployment story (post-merge)

- **New / restarted VMs** — `startup.sh` already fetches Caddyfile + compose at boot. No action needed.
- **Existing long-uptime VMs** — one-shot to land the new auto-upgrade script (the currently-running script is the OLD version without self-update logic). After this one-shot, the new script self-updates on every tick going forward:
  ```bash
  gcloud compute ssh <vm> --zone=... --command='
    curl -fsSL https://raw.githubusercontent.com/keboola/agnes-the-ai-analyst/main/scripts/ops/agnes-auto-upgrade.sh \
      -o /usr/local/bin/agnes-auto-upgrade.sh && \
    chmod +x /usr/local/bin/agnes-auto-upgrade.sh
  '
  ```
- Within ~5 minutes of the one-shot the new script's first tick recreates containers with the new Caddyfile + compose.

## Test plan

- [x] `pytest tests/test_check_access_endpoint.py` — 5/5 pass (admin → 204, analyst-no-grant → 403, analyst-with-grant → 204, unsafe id → 404, no auth → 401)
- [x] `pytest tests/test_access_control.py` — 21/21 pass (no regression on the existing `/download` endpoint)
- [x] `pytest -k pull` — 12 passed, 2 skipped (parallel ThreadPool path is exercised in `cli/lib/pull.py`; existing pull tests didn't regress)
- [x] `caddy validate` on the new Caddyfile — parses cleanly
- [x] `bash -n` on the new auto-upgrade script
- [x] **Live e2e on `foundryai-dev-zsrotyr`** — see table above; all concurrent + serial requests succeeded, parallel CLI pull verified
- [ ] Post-merge: run the one-shot above against `foundryai-production` + `foundryai-development` + the per-developer dev VMs, then `agnes pull` test from a real laptop with concurrent UI activity.

## What this PR does NOT do

- No multi-worker uvicorn (DB-lock issue, separate refactor).
- No GCS-backed downloads (separate, larger initiative for 50+ analyst tier).
- The `/api/v2/scan` Arrow streaming path is unaffected — different concern, dynamic response shape doesn't fit `file_server`.


ZdenekSrotyr:

PR #186, opened 2026-05-05
fix: five regressions surfaced by a fresh-deploy run

ZdenekSrotyr:
## Summary

Five independent regressions / latent bugs found while standing up a new instance against a non-trivially-configured Keboola Storage backend. Bundled because they were discovered in the same one investigation and the fix sites are tiny.

### 1. \`agnes pull\` hash mismatch on every Keboola local-mode table

\`src/orchestrator.py:_update_sync_state\` stored \`md5(f\"{mtime_ns}:{size}\")[:12]\` into \`sync_state.hash\` — a 12-char fingerprint of file metadata — while the CLI's post-download integrity check compares against the full 32-char content MD5 it computes via \`cli/commands/sync.py:_md5_file\`. Those could never match, so every \`agnes pull\` reported \`Updated 0 tables\` even when the server had data. The materialized SQL path already used a real content MD5 (\`app/api/sync.py:_file_hash\`), so only Keboola local-mode tables were broken.

Now the orchestrator streams the parquet through \`hashlib.md5\` and stores the full hexdigest, matching the CLI and the materialized path.

### 2. \`connectors/keboola\`: trailing-slash sanitization

The DuckDB Keboola extension's \`ATTACH\` fails with a network error when the URL ends in \`/\` (e.g. the canonical \`https://connection.<region>.keboola.com/\` form). Bare host works. Sanitized in both \`KeboolaAccess\` (materialized) and \`_try_attach_extension\` (batch extractor) so operator-facing config can stay canonical.

### 3. \`src/profiler.py:TableInfo.description\` becomes optional

Two call sites instantiate \`TableInfo(name=..., table_id=...)\` without it. Sync's profiler pass crashed with \`TableInfo.__init__() missing 1 required positional argument: 'description'\`, silently leaving \`[SYNC] Profiled 0 tables\`. Defaulting to \`\"\"\` keeps both call sites working.

### 4. \`scripts/ops/agnes-auto-upgrade.sh\`: chown on UID change

Older images ran as root; the current image runs as \`agnes\` (uid 999). Without intervention, an upgrade across that boundary leaves \`/data/state/.session_secret\` and DuckDB owned by root, and the new container fails to start with \`PermissionError\`.

Now: when the image digest moves, the script reads the target uid:gid from \`/etc/passwd\` inside the image and \`chown\`s \`/data/{state,extracts,analytics}\` before \`docker compose up\`. Idempotent — no-op when ownership already matches.

### 5. \`POST /api/sync/trigger\` is now singleton per process

Two near-simultaneous trigger calls (scheduler tick + manual retrigger) each scheduled their own \`_run_sync\`, which forked an extractor subprocess. Both subprocesses then opened the same \`extract.duckdb\` and fought for its file lock — observed live: three concurrent extractors → DB-lock contention starved uvicorn → \`/api/health\` timed out 3× → Docker flipped the app container to \`unhealthy\` → behind the bundled Caddy \`reverse_proxy\` overlay, external traffic dropped until contention drained.

- Module-level \`threading.Lock\` in \`app/api/sync.py\`. Trigger handler returns **409** (\`detail: sync_already_in_progress\`) when held, before scheduling the background task. \`_run_sync\` acquires non-blocking on entry, releases in \`finally\` — defense in depth.
- Hoist \`import sys as _sys\` to the top of \`_run_sync\`. Latent bug: the outer \`except Exception\` block referenced \`_sys\` but the import sat inside the \`try\` body, so an early failure traded the original error for \`NameError: name '_sys' is not defined\`.

Operator impact: clients that hand-roll their own polling on \`/api/sync/trigger\` now need to handle 409. The bundled scheduler container's \`data-refresh\` job logs the 409 as a normal HTTP warning and waits for next tick — no retry loop.

## Test plan

- [x] \`pytest tests/test_keboola_extractor.py tests/test_keboola_extractor_full.py tests/test_keboola_access.py tests/test_profiler.py tests/test_sync_trigger_singleton.py tests/test_sync_filter.py tests/test_orchestrator_sync_state_hash.py\` — 84 passed, 1 skipped
- [x] \`bash -n scripts/ops/agnes-auto-upgrade.sh\` — syntax OK
- [ ] CI green
- [ ] On a live instance: trigger sync, then \`agnes pull\` on the analyst side downloads parquets without hash mismatch
- [ ] On a live instance: trigger sync twice within seconds → second call returns 409, only one extractor subprocess in \`/proc\`, app stays healthy throughout


ZdenekSrotyr:
Reopened as #190 from `kb/main` — that branch now drives the agnes-dev VM via `:dev-kb-latest` floating tag (build per push). Same content, same commits, just relocated so the dev VM picks up changes without going through main.

PR #187, opened 2026-05-05
feat(workspace-prompt): decision tree + size-hint so analyst Claude gets it right first try

ZdenekSrotyr:
## Why

A real analyst Claude session in another window misused the CLI on first try and reported five concrete failures, all of which trace back to ambiguities in the analyst-facing workspace prompt + missing structured signals on the `agnes catalog` payload:

- `agnes query --remote --estimate` → `No such option: --estimate Did you mean --remote?` — `--estimate` is on `agnes snapshot create`, not `agnes query`. The template said "ALWAYS run `--estimate` first" inside the snapshot section but a few lines above mentioned `--estimate` without binding to a specific command — easy to misread.
- `agnes fetch` was attempted because stale docs/training data still references it. The command was renamed to `agnes snapshot create` per the 0.34.0 bootstrap PR; the template doesn't acknowledge the rename.
- Repeated guesswork on "how big is this table" — the catalog row carries `rough_size_hint` but it's hardcoded `null` with a TODO; analysts only discover size empirically via a failed `--remote` round-trip.
- `[update] da X is out of date` printed on every command (CLI banner survived the `da → agnes` rename) — trains analysts to associate the binary with the old name.

## What changes

| File | Change |
|---|---|
| `config/claude_md_template.txt` | Discovery section gets a field-by-field table of `agnes catalog --json` columns (each with "what it tells you" + "how to use it"). Adds an explicit decision tree by `query_mode`. Adds a "Common mistakes" callout that explicitly scopes `--estimate` to `agnes snapshot create` and the `agnes fetch` → `agnes snapshot create` rename. Adds a 6-row failure-mode dictionary mapping each common error wording to cause + right next step (timeout, scan-too-large, 401, cross_project_forbidden, ReadTimeout, unknown columns). |
| `app/api/v2_catalog.py` | Populates `rough_size_hint` for `local` + `materialized` rows from the on-disk parquet size at `${DATA_DIR}/extracts/<source_type>/data/<table_id>.parquet`, bucketed `small` (≤100 MiB) / `medium` (≤1 GiB) / `large` (≤10 GiB) / `very_large` (>10 GiB). `remote` stays `null` until a separate BQ INFORMATION_SCHEMA path lands. Fail-soft on filesystem errors. |
| `cli/update_check.py` | `[update] da X is out of date` → `[update] agnes X is out of date`. |
| `CHANGELOG.md` | Entries under `[Unreleased]` → Added / Changed. |

## Decision tree (preview)

The template now spells this out so the AI can pick its tool BEFORE writing any SQL:

```
                       ┌─ local        → agnes query "SELECT ..."
agnes catalog →   ─────┤
query_mode of <table>  ├─ materialized → agnes query (parquet was synced by agnes pull)
                       │                 (if missing locally, run `agnes pull` first)
                       │
                       └─ remote       → choose by table size + query shape:
                                          - one cheap probe (COUNT, schema-confirm, single agg ≤200s)
                                              → agnes query --remote "..."
                                          - repeated questions on same slice / large scan
                                              → agnes snapshot create <table> --select ... --where ... --as <name>
                                                then agnes query "SELECT ... FROM <name>"
                                          - join with a local table
                                              → agnes query --register-bq "alias=BQ_SQL" --sql "..."
```

## Failure-mode dictionary (preview)

Every error wording an analyst will see, mapped to cause + next step:

| Error wording (substring) | Cause | Response |
|---|---|---|
| `Binder Error: Query execution exceeded the timeout. Job ID: ...` | BQ-side query took >~200 s wall-clock; the DuckDB BQ extension's `bq_query_timeout_ms` elapsed | Narrow `--where`, drop columns, or `agnes snapshot create` |
| `HTTP 400: remote_scan_too_large` | Server's `bq_max_scan_bytes` cost gate refused | Tighten `--where`; `agnes snapshot create` so cost is paid once |
| `HTTP 401: ... unauthorized` | PAT expired | `agnes init --token <new>` |
| `HTTP 403: cross_project_forbidden` | Server SA lacks BQ access | Escalate to admin; do NOT try personal auth |
| `ReadTimeout` (client-side) | CLI <0.35.1 (had 30 s default) | `uv tool install --force <wheel-from-update-banner>` |
| `unknown columns: [...]` from `agnes snapshot create` | `--select` lists columns that don't exist | `agnes schema <table>` and copy column names verbatim |

## Test plan

- [x] `pytest tests/test_claude_md_renderer.py tests/test_claude_md_api.py` — 33/33 pass; renderer doesn't choke on the new content
- [x] Full template render against representative `data_source: bigquery` and `data_source: keboola` contexts — output is valid markdown, jinja `StrictUndefined` doesn't trip
- [x] Empirical reproduction of all 5 use-cases from the original analyst screenshot via real CLI against a dev VM (issue #5 reproduced; #3 fixed by separate PR #181; #2 Windows-only; #1 already fixed by 0.35.1)
- [x] Vendor-token sweep — no foundryai/groupon/prj-grp leakage

## Out of scope (separate PRs / follow-ups)

- BQ INFORMATION_SCHEMA lookup for `remote` row size (would let `rough_size_hint` work for `--remote`-only tables too)
- Server-side cache of `rough_size_hint` so the catalog response doesn't `stat()` on every call (currently no significant cost — the catalog endpoint is already cached for 5 min via `_table_rows_cache`)
- Translating the existing `docs/CLAUDE.md`'s "Querying Agnes data — agent rails" section to current `agnes` CLI (separate doc cleanup)

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/187" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Added
+- **Workspace prompt: decision tree, common-mistakes callout, failure-mode dictionary** in `config/claude_md_template.txt` (the template `agnes init` writes to `<workspace>/CLAUDE.md`). Surfaces every catalog-row field analyst Claude should read before deciding which command to use (`query_mode`, `sql_flavor`, `where_examples`, `fetch_via`, `rough_size_hint`); explicitly binds `--estimate` to `agnes snapshot create` ONLY (was the most-failed first-try misuse — fails with `No such option: --estimate` on `agnes query`); calls out the `agnes fetch` → `agnes snapshot create` rename so stale-doc analysts don't run a non-command; documents the BQ permission model (server SA, not personal Google identity) and a 6-row failure-mode table mapping each common error wording to its cause + the right next step.
+- **`rough_size_hint` populated for `local` + `materialized` catalog rows** in `GET /api/v2/catalog` (was hardcoded `null` with a "Task 8" TODO). Reads the parquet file size at `${DATA_DIR}/extracts/<source_type>/data/<table_id>.parquet` and buckets into `small` (≤100 MiB), `medium` (≤1 GiB), `large` (≤10 GiB), `very_large` (>10 GiB). `remote` rows stay `null` for now (size requires a BQ INFORMATION_SCHEMA call; tracked separately). Lets analyst Claude pick `agnes snapshot create` over `agnes query --remote` by inspecting `agnes catalog --json` rather than discovering size empirically via a failed `--remote` round-trip.
+
+### Changed
+- **CLI update-banner now says `agnes` instead of `da`** (`cli/update_check.py:format_outdated_notice`). The string `[update] da X is out of date` had survived the `da` → `agnes` CLI rename and was the most-visible stale identifier in the analyst-facing surface — every CLI command printed it on stderr when a newer wheel was available.
+
 ### Fixed
 
 - Keboola sync now falls back to the legacy Storage-API client when the DuckDB Keboola extension's per-table scan fails, not just when the initial `ATTACH` fails. Two changes:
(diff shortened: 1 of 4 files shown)


ZdenekSrotyr:

PR #188, opened 2026-05-05
feat+perf: BQ timeout + Caddy file_server + parallel pull + workspace prompt + Tier 1 event-loop unblocking

ZdenekSrotyr:
Combined PR replacing #181, #182, #187. All three address the same 2026-05-05 prod incident chain (single uvicorn worker starving on a multi-GB pull, BQ queries timing out at 90 s, analyst Claude misusing the CLI on first try). Reviewed and merged together gives one clean release-cut instead of three.

## Six commits, four concerns

**1. BQ extension query timeout** (was PR #181 — `4f042355`, `025a2b5c`)

The DuckDB BigQuery extension's built-in `bq_query_timeout_ms` defaults to 90 s, which is too tight for analyst-scale queries against view-backed BQ datasets. `agnes query --remote` HTTP 400'd with `Binder Error: Query execution exceeded the timeout. Job ID: …` whenever the underlying BQ job ran longer.

- New `data_source.bigquery.query_timeout_ms` config knob (default 600 000 ms = 10 min, sentinel 0 leaves the extension default).
- Applied via `SET bq_query_timeout_ms` after every `LOAD bigquery` on **all four** BQ-touching DuckDB session sites: orchestrator's `_remote_attach`, the analytics-DB read-only reattach path in `src/db.py:_reattach_remote_extensions` (the path `agnes query --remote` actually hits — caught by Devin Review as a critical miss in the original PR), the `BqAccess` session factory, and the standalone extractor.
- Empirically validated: bumping the value lets queries up to ~200 s succeed; BQ's `jobs.query` RPC server-caps at ~200 s regardless, so the effective ceiling is ~200 s. Documented in CHANGELOG + admin UI hint.

**2. Caddy `file_server` bypasses uvicorn for parquet downloads** (was PR #182 — `1be997f6`)

A single analyst's multi-GB `agnes pull` held the only uvicorn worker for the duration of the stream, starving the UI, `/api/health`, and every other API endpoint. Container flipped to `unhealthy`. Reproduced live during this work.

- New endpoint `GET /api/data/{table_id}/check-access` — lightweight RBAC probe, returns 204/403, ~1 ms.
- Caddyfile `@download` matcher uses `forward_auth` to call `/check-access`, then `try_files` + `file_server` to serve the parquet directly from `/srv/extracts/<source>/data/<id>.parquet`. Bulk transfer never touches uvicorn.
- docker-compose.yml mounts the agnes data volume at `/srv:ro` inside the caddy container.
- Anything that misses `try_files` falls through to the existing app handler so legacy `src_data/parquet` and future connectors keep working.
- Live e2e on a dev VM: 3 concurrent downloads + 5 parallel `/api/health` probes — all 8 succeed, downloads ~0.9 s, health ~0.85 s. Pre-PR this would have starved health.

**3. `agnes-auto-upgrade.sh` watches config files + self-updates** (was PR #182 — `ab61e30c`)

Without this, the Caddy + compose changes above would land in the image but stay inert on long-uptime VMs — auto-upgrade only watched `docker images` digests; a Caddyfile or compose change in main never reached running VMs.

- Re-fetches all compose overlays + the Caddyfile from `keboola/agnes-the-ai-analyst@main` on every 5 min tick, hashes them, and triggers `docker compose up -d` recreation when the hash changes.
- Self-updates `/usr/local/bin/agnes-auto-upgrade.sh` from main on every tick. Without this, the very fix that watches config files would never propagate to existing VMs.
- Fail-soft on curl errors via the `.new + mv` pattern.

**4. Per-user parallel `agnes pull`** (was PR #182 — `2ae486bc`)

`cli/lib/pull.py`'s download loop was strictly serial. With Caddy file_server above, the server can sustain parallel transfers; the client became the bottleneck.

- ThreadPoolExecutor capped by `AGNES_PULL_PARALLELISM` (default 4, set 1 to restore serial). Drops to serial when `len(to_download) ≤ 1`.
- Per-table error semantics preserved via `(tid, entry, err)` tuple.

**5. Workspace prompt clarity** (was PR #187 — `30e81a15`)

A real analyst Claude session in another window misused the CLI on first try and reported five concrete failures. All trace to ambiguities in the analyst-facing workspace prompt.

- `config/claude_md_template.txt`: Discovery section gets a field-by-field table of `agnes catalog --json` columns. Adds an explicit decision tree by `query_mode`. Adds a "Common mistakes" callout that scopes `--estimate` to `agnes snapshot create` (was the most-failed first-try misuse) and the `agnes fetch` → `agnes snapshot create` rename. Adds a 6-row failure-mode dictionary mapping each common error wording to cause + right next step.
- `app/api/v2_catalog.py`: populates `rough_size_hint` for `local` + `materialized` rows from the on-disk parquet size, bucketed `small`/`medium`/`large`/`very_large`. Was hardcoded `null`. Lets analyst Claude pick `agnes snapshot create` over `agnes query --remote` by inspecting catalog rather than discovering size via a failed `--remote` round-trip.
- `cli/update_check.py`: `[update] da X is out of date` → `[update] agnes X is out of date`. Banner survived the `da → agnes` rename.

## Deployment story (post-merge)

- **New / restarted VMs** — `startup.sh` fetches Caddyfile + compose at boot. No action needed.
- **Existing long-uptime VMs** — one-shot to land the new auto-upgrade script (the currently-running script is the OLD version without self-update logic). After this one-shot, the new script self-updates on every tick going forward:
  ```bash
  gcloud compute ssh <vm> --zone=... --command='
    curl -fsSL https://raw.githubusercontent.com/keboola/agnes-the-ai-analyst/main/scripts/ops/agnes-auto-upgrade.sh \
      -o /usr/local/bin/agnes-auto-upgrade.sh && \
    chmod +x /usr/local/bin/agnes-auto-upgrade.sh
  '
  ```

## Why not multi-worker uvicorn (considered, dropped)

Blocked by `system.duckdb`'s single-writer DuckDB lock — a second uvicorn worker would crash at startup on `Could not set lock on file`, the same race that pushed the scheduler from in-process writes to HTTP-via-app. This PR sidesteps the issue entirely by getting bulk transfer off uvicorn.

## Test plan

- [x] `pytest tests/test_bq_query_timeout.py` — 8/8 pass
- [x] `pytest tests/test_check_access_endpoint.py` — 5/5 pass
- [x] `pytest tests/test_access_control.py` — 21/21 pass
- [x] `pytest tests/test_claude_md_*.py` — 33/33 pass
- [x] `pytest -k "pull or check_access or access_control or query_timeout or materialize or orchestrator or claude_md"` (combined relevant set) — 93/93 pass
- [x] `caddy validate` on the new Caddyfile — parses cleanly
- [x] Live e2e on `foundryai-dev-zsrotyr` (a real grpn dev VM): all auth + concurrent download tests passed; ETag format proved Caddy file_server served directly (not reverse_proxy through uvicorn).
- [x] `agnes pull` from a laptop CLI works under the new parallel + Caddy stack
- [x] All five use-cases from the original analyst screenshot reproduced; #5 (`agnes query --remote --estimate`) confirmed unfixable in code and now explicitly warned about in the workspace prompt.
- [ ] Post-merge: run the auto-upgrade-script one-shot on `foundryai-production` + `foundryai-development` + per-developer dev VMs, then a real analyst `agnes pull` test with concurrent UI activity.

## Replaces

- #181 (`zs/bq-query-timeout`) — closed in favour of this combined PR
- #182 (`zs/scale-downloads`) — closed in favour of this combined PR
- #187 (`zs/claude-md-firsttry`) — closed in favour of this combined PR

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/188" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---

## Tier 1 event-loop unblocking (added in commit `e5fb913c`)

Per user's perf request after architecture review: 5 hottest BQ-touching endpoints (`/api/query`, `/api/v2/scan`, `/api/v2/scan/estimate`, `/api/v2/sample`, `/api/v2/schema`) were `async def` calling synchronous DuckDB / BQ-extension. Under uvicorn's single event loop, **a single heavy `agnes query --remote` froze every other request** (UI, /api/health, auth, even another query) for the full BQ wait (~90-200 s).

- Converted all five to plain `def` → FastAPI auto-offloads body to anyio thread pool.
- Audit: 0 `await` statements in any converted handler — rename is safe.
- Tests in `tests/test_v2_*.py` updated: `asyncio.run(handler(...))` → `(handler(...))` since handlers no longer return coroutines.
- New `AGNES_THREADPOOL_SIZE` env var (default 200, was anyio's stock 40) bumps the FastAPI thread pool capacity in `app/main.py:lifespan`.

This is the single biggest perf win available without a multi-process refactor — the VM was idle but the app blocked on event-loop-held BQ calls. With Tier 1 the event loop stays free; per-handler concurrency is now bounded by AGNES_THREADPOOL_SIZE (200) instead of 1.


ZdenekSrotyr:
Both Devin Review findings fixed in commit `7a72ea9c`:

**🔴 Caddyfile try_files fallback** — appended `/api/data/{re.tid.1}/download` as the final try_files entry. When no parquet matches the three static source paths, the URI is rewritten back to the analyst-facing `/api/data/<id>/download` and the reverse_proxy below correctly forwards to the app's `download_table` handler. Caddyfile parses cleanly under `caddy validate`. The 'missed → falls through to app, never 404' promise in the comment now actually holds.

**🟡 auto-upgrade.sh stale TLS decision** — moved the COMPOSE_FILES tls extension + PROFILE_ARGS computation to AFTER the config re-fetch loop. Now a freshly-fetched Caddyfile (or freshly-removed one) is reflected in this tick's docker compose set, not the next one. `docker compose pull` also runs after the fetch with the updated COMPOSE_FILES — same image regardless of overlay so the pull behavior is unchanged either way, but ordering is now correct.

Plus a separate user-feedback fix in the same commit: stripped the workspace prompt of stale-prone table-list / metric-count enumerations; replaced with explicit guidance to discover via `agnes catalog` / `agnes schema` / `agnes describe` at runtime.

ZdenekSrotyr:
Both new Devin findings fixed in commit `f2ce9154`:

**🚩 `/api/v2/catalog` async def vs sync stat()** — converted to plain `def` so FastAPI auto-offloads to the anyio thread pool, same Tier 1 treatment the other handlers got. The new `_materialized_size_hint` populator's `Path.stat()` calls now run on a worker thread; event loop stays free for non-catalog traffic. NFS / GCS-FUSE deployments no longer freeze on a slow filesystem syscall.

**🔴 stream_download eating HTTPStatusError** — accepted Devin's exact suggestion. Short-circuit `isinstance(last_exc, httpx.HTTPStatusError)` before the generic `isinstance(last_exc, httpx.HTTPError)` branch so 401 / 403 / 404 from the server re-raise verbatim. The caller's `str(exc)` now shows "Client error '401 Unauthorized' for url ..." instead of the generic "Unexpected error: HTTPStatusError". Verified that `api_get` / `api_post` / `api_delete` / `api_patch` don't have the same bug — they don't call `.raise_for_status()` so HTTPStatusError can't reach their except clause.

The 8 additional Devin Review findings in the dashboard — I'll triage those after this round of fixes lands; if any are blocking I'll fold them into another commit on this PR.

PR #189, opened 2026-05-05
fix(sync): make POST /api/sync/trigger singleton per process

ZdenekSrotyr:
## Summary

Two near-simultaneous \`POST /api/sync/trigger\` calls (scheduler tick + manual retrigger, or two operators clicking at once) each scheduled their own \`_run_sync\`, which forked an extractor subprocess. Both subprocesses then opened the same \`extract.duckdb\` and fought for its file lock — observed live on a dev VM today: three concurrent extractors → DB-lock contention starved uvicorn → \`/api/health\` timed out 3× → Docker flipped \`agnes-app\` to \`unhealthy\` → Caddy \`reverse_proxy\` upstream check failed → external traffic dropped until contention drained.

## Changes

- Module-level \`threading.Lock\` in \`app/api/sync.py\`. The trigger handler returns **409** (\`detail: sync_already_in_progress\`) when held, before scheduling the background task. \`_run_sync\` acquires the lock non-blocking on entry and releases it in \`finally\` — defense in depth in case anything bypasses the handler.
- Hoist \`import sys as _sys\` to the top of \`_run_sync\`. The outer \`except Exception\` block referenced \`_sys\`, but the import sat inside the \`try\` body, so an early failure (e.g. config-loading exception) traded the original error for \`NameError: name '_sys' is not defined\`. Latent — surfaced by a new test that injects an exception before the import line.
- 4 new tests in \`tests/test_sync_trigger_singleton.py\`:
  - lock-held → \`_run_sync\` no-ops with log line, doesn't enter body
  - lock-released-after-exception via \`finally:\`
  - handler returns 409 when locked
  - handler returns 200 + schedules background task when free

## Operator impact

Clients that hand-roll their own polling on \`/api/sync/trigger\` now need to handle 409. The scheduler container's \`data-refresh\` job logs 409 as a normal HTTP warning and waits for its next tick — no retry loop. Documented in CHANGELOG \`### Changed\`.

## Test plan

- [x] \`pytest tests/test_sync_trigger_singleton.py tests/test_sync_trigger_materialized.py tests/test_sync_trigger_keboola_materialized.py tests/test_sync_manifest.py tests/test_sync_filter.py\` — 52 passed
- [ ] CI green
- [ ] On a live dev VM: trigger sync twice within seconds → second call returns 409, only one extractor subprocess in \`/proc\`, app stays healthy throughout

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/189" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,13 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Changed
+
+- `POST /api/sync/trigger` is now singleton per process. A second trigger that arrives while the previous sync is still running returns **HTTP 409** (`detail: sync_already_in_progress`) instead of scheduling a parallel `_run_sync`. The scheduler container's `data-refresh` job logs the 409 as a normal warning and waits for its next tick — no retry loop. Operator-visible: clients that hand-roll their own polling on `/api/sync/trigger` now need to handle 409. Why it matters: two concurrent extractor subprocesses both write `extract.duckdb`, fight for its file lock, starve uvicorn's worker pool, and Docker flips `agnes-app` to `unhealthy` long enough for `reverse_proxy`-fronted deploys to return 503 to external traffic until contention drains.
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


ZdenekSrotyr:
Folded into #186 — same scope of cutover follow-ups, single PR is easier to review & deploy in one tag.

Issue #79, opened 2026-04-27
table_registry.sync_schedule is stored and surfaced but never evaluated at runtime


davidrybar-grpn:
## Summary

`table_registry.sync_schedule` is accepted by the admin API, persisted in DuckDB, echoed back in responses, and parsed by `src/scheduler.py` — but **no production code evaluates it**. Every `/api/sync/trigger` call re-extracts **every** registered local table regardless of the table's schedule. Operators have no signal it's a no-op.

This is structurally identical to #78 (`script_registry.schedule`), but bigger: `sync_schedule` is part of the documented admin surface (CLAUDE.md describes the schema; `app/api/admin.py` exposes it) and operators can already set it via the UI.

### 1. The runtime evaluator was deleted in a documented refactor

Commit `9f5b9d3` (2026-03-31, *\"refactor: delete old sync pipeline — 9,500 lines removed\"*) deleted `src/data_sync.py`, which contained the only production caller:

```python
# src/data_sync.py:520 (deleted in 9f5b9d3)
def sync_scheduled(self) -> Dict[str, Dict[str, Any]]:
    \"\"\"Synchronize only tables whose sync_schedule says they are due.
    Evaluates each table's sync_schedule against its last_sync timestamp.
    \"\"\"
    from src.scheduler import is_table_due
    ...
    if is_table_due(tc.sync_schedule, last_sync):
        logger.info(f\"Table {tc.name} is DUE (schedule: {tc.sync_schedule})\")
```

Called at the deleted `src/data_sync.py:702`: `results = manager.sync_scheduled()`. The same commit also deleted `src/config.py:174-183`, which validated `sync_schedule` format on input.

The replacement (`_run_sync()` in `app/api/sync.py`, added during the same refactor) **does not re-wire** the schedule check. The parser (`src/scheduler.py`) and the column (`table_registry.sync_schedule`) survived as orphans.

### 2. Today's `_run_sync()` does not consult sync_schedule

`app/api/sync.py:35-200`:
- Line 62: `repo.list_local(source_type)` returns all tables where `query_mode='local'`. The repo (`src/repositories/table_registry.py:71-85`) filters only by `query_mode` and optional `source_type` — never by `sync_schedule`.
- Lines 91-95: each config dict is JSON-serialized and passed verbatim to the extractor subprocess (schedule rides along as passenger data).
- Line 122: subprocess extracts every config it receives.

### 3. No downstream consumer either

[...]

Issue #78, opened 2026-04-27
script_registry: schedule / last_run / last_status are persisted but never consumed


davidrybar-grpn:
## Summary

The `script_registry` table has a `schedule` column (and `last_run` / `last_status` bookkeeping columns) that the API accepts, persists, and returns to callers — but **no code ever consumes them**. An analyst can deploy a script with `schedule=\"every 1h\"`, the value round-trips through the API, and the script never runs.

## Verified state

**What's wired up:**
- `script_registry.schedule` column declared in `src/db.py:136`
- `POST /api/scripts/deploy` accepts `schedule: Optional[str]` (`app/api/scripts.py:29`) and stores it via `ScriptRepository.deploy()` (`src/repositories/notifications.py:75`)
- `GET /api/scripts` returns the stored value
- `ScriptResponse` exposes it in the schema (`app/api/scripts.py:40`)

**What's missing:**
- The sidecar scheduler (`services/scheduler/__main__.py`) only calls `/api/sync/trigger` and `/api/health`. Nothing reads `script_registry.schedule` and nothing fires `POST /api/scripts/{id}/run` based on it.
- `script_registry.last_run` and `script_registry.last_status` (declared at `src/db.py:139-140`) are **never written** anywhere — `deploy()` doesn't set them, and there's no post-run update path.

The original design (`docs/superpowers/specs/2026-03-27-refactoring-design.md:228` — *\"ScriptRepository (registry + scheduling)\"*) anticipated both halves; the registry half shipped, the scheduling half didn't.

The schedule string format would be reusable as-is — `src/scheduler.py` already parses `every Xm` / `every Xh` / `daily HH:MM[,HH:MM,...]` for `table_registry.sync_schedule`.

## Pick one — implementer's call

The right answer depends on whether scheduled user scripts are actually a desired feature. Whoever picks this up should make the call.

### Option A — implement scheduled scripts (deliver the missing half)

- Extend `services/scheduler/__main__.py` to load `script_registry`, evaluate each row's `schedule` via `src.scheduler.is_table_due(...)`, and `POST /api/scripts/{id}/run` when due.
- Update `last_run` / `last_status` after each invocation (extend `ScriptRepository` with an `update_run_result(...)` method).
[...]

PR #191, opened 2026-05-05
host-mount: replace named-volume driver_opts with direct service binds

cvrysanek:
## Summary

Eliminates a class of silent data-corruption bug from the bind-mount overlay by switching from "named volume with driver_opts" to "direct service-level bind mounts."

## The bug class

Docker named volumes are immutable in their driver options after creation. Once `agnes_data` is created with whatever options were in effect at `compose up` time, future edits to `driver_opts` in this file are silently ignored on existing volumes.

## How it bit us

Groupon FoundryAI deployment, 2026-05-05:

1. **2026-04-22**: VMs bootstrapped with this overlay, which at the time had `o: bind` (non-recursive). Volume `agnes_data` created with that option.
2. **2026-04-24**: Disk migration adds a second persistent disk for irreplaceable state, mounted under the data disk at `/data/state`.
3. **A later commit** updates this file to `o: bind,rbind` so the new nested mount propagates into containers.
4. **The existing `agnes_data` volume keeps the old `bind` option** because Docker doesn't migrate driver_opts on existing volumes.
5. **Containers write to a shadowed subdirectory of the parent disk** instead of the nested child mount. SQLite/DuckDB files diverge between what the host sees (correct disk) and what the app actually writes (shadow).
6. **2026-05-05**: a routine `tls_reconcile` recreates the agnes-app container. The new container can't open `/data/state/system.duckdb.wal` (root-owned, on the shadow). DuckDB enters FATAL state. Every authed request 500s. Google sign-in broken for ~30 minutes until manual recovery.

Recovery required `docker volume rm agnes_data` + selective data migration on every affected VM (5 of 9 in the deployment).

## The fix

Replace the named-volume `driver_opts` approach with direct service-level bind mounts via `volumes: !override` per service:

```yaml
services:
  app:
    volumes: !override
      - /data:/data           # direct host bind, recursive by default
      - ./config:/app/config:ro
```

**Why this fixes it permanently**:
- Direct bind mounts don't go through Docker's volume layer.
- Mount options re-evaluate every container start.
- Modern Docker Engine (20.10+) defaults to **recursive bind** for service-level binds, so any nested mount on the host (`/data/state`, etc.) propagates into containers automatically.
- No volume to migrate, no `driver_opts` to forget, no immutability footgun.

The `data:` named volume declaration in `docker-compose.yml` is left intact — it's still the path used by local-dev `compose up` without this overlay (where there's no host `/data` directory).

## Validation

```bash
$ docker compose -f docker-compose.yml -f docker-compose.prod.yml -f docker-compose.host-mount.yml config | head
services:
  app:
    volumes:
    - type: bind
      source: /data
      target: /data
      bind:
        create_host_path: true
    - type: bind
      source: ./config
      target: /app/config
      read_only: true
```

Each of `app`, `extract`, `scheduler`, `telegram-bot`, `ws-gateway` now bind `/data:/data` directly. The named volume `data:` no longer appears under any service's `volumes:`.

The 2026-05-05 incident root-cause analysis (with full inode comparison, mount tables, and the deployer-side recovery script) is captured in [Groupon's infra repo](https://github.groupondev.com/FoundryAI/agnes-the-ai-analyst-infra) PRs #78 + #79 + #80. The recovery script writes this same direct-bind config inline as a fallback when the deployed image lacks it — so this PR retroactively makes that fallback unnecessary for new deployments.

## Compose-spec version

`!override` merge tag is part of the Compose Specification supported by Docker Compose v2.20+. Tested against Compose v5.1.3 (Groupon deployment).

## What changes for existing deployers

- Next image rebuild containing this commit ships the new overlay.
- On the operator's VM, next `docker compose up -d` with this overlay starts containers with direct binds. The old `agnes_data` named volume is no longer referenced; can be removed with `docker volume rm agnes_data` (operator's choice — orphaned but harmless if left).
- For deployers who hit the 2026-05-05 shadow class: this PR makes the fix permanent. The recovery scripts can be retired once all VMs are on an image containing this commit.

## Backward compatibility

Local dev (`compose up` without the overlay): unchanged. Named volume `data` still works as before.
CI: unchanged. Overlay isn't loaded in CI per existing comment in the file.

### DIFF
--- a/docker-compose.host-mount.yml
+++ b/docker-compose.host-mount.yml
@@ -1,14 +1,41 @@
-# Bind-mount overlay — replaces the `data` named volume with a bind mount
-# to /data on the host.
+# Bind-mount overlay — replaces the `data` named volume with a direct
+# host bind mount per service.
 #
-# Use this when /data is a persistent disk mounted by the VM startup script,
-# so Agnes data lives on the PD (not on the boot disk's Docker volume).
+# Why direct service-level bind, not driver_opts on the named volume
+# ------------------------------------------------------------------
+# The previous version of this file modified the `data` named volume's
+# `driver_opts` to point at /data with `o: bind,rbind`. Docker named
+# volumes have an immutability footgun: once a volume is created, its
+# driver options are fixed for the life of the volume. Editing this
+# file and re-running `docker compose up -d` does NOT propagate the
+# new options to existing volumes — they keep whatever options were
+# in effect at create time.
 #
-# `bind,rbind` (recursive bind) is required when the host nests a second
-# disk under /data — e.g. the dual-disk layout where sdb is mounted on /data
-# and sdc on /data/state. A plain `bind` captures only the top-level mount
-# and silently shadows the sub-mount with an empty subdirectory inside the
-# container, causing the app to write to the wrong disk.
+# This bit a deployer (Groupon FoundryAI) on 2026-05-05: the volume
+# was created before this overlay had `bind,rbind`, kept the old
+# `bind` (non-recursive) propagation, and containers wrote to a
+# shadowed subdirectory of the parent disk instead of the nested
+# child mount. DuckDB went FATAL on a root-owned WAL during a
+# routine container recreate; sign-in broke.
+#
+# Direct service-level bind mounts (`/host/path:/container/path`)
+# don't go through Docker's volume layer at all. They re-evaluate
+# the mount options every container start, and modern Docker Engine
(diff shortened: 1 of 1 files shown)


ZdenekSrotyr:
Cherry-picked into #194 (rebase on post-#188 main + caddy `!override` regression fix). Closing as superseded once #194 merges.

ZdenekSrotyr:
Superseded by #194 — content cherry-picked verbatim onto post-#188 main, with attribution preserved (commit `655822b9`).

PR #192, opened 2026-05-05
feat: STATE_DIR env var + flat-mount overlay (parallel disks)

cvrysanek:
## Summary

Introduces `STATE_DIR` as the single source of truth for the writable state directory path, with backward-compatible default of `${DATA_DIR}/state`. Pairs with a new `docker-compose.flat-mount.yml` overlay that mounts the state disk in **parallel** to the data disk rather than nested under it.

## Companion to #191

This is the second half of the response to the 2026-05-05 incident. #191 (host-mount: replace named-volume driver_opts with direct service binds) fixes the immutability footgun for the existing nested-mount layout. This PR offers a **flat layout** as the architectural alternative.

Both PRs are independent and can land separately. Deployers can choose:
- **Layout A** (nested) + #191's fix → keep current topology, eliminate the volume-options-immutability bug.
- **Layout B** (flat) → eliminate nested-mount class entirely.

### App code (Python)
- `src/db.py` — new `_get_state_dir()` helper. `get_system_db()` and schema migration snapshot use it.
- `app/secrets.py` — new `_state_dir()` helper. `_load_or_generate()` uses it for `.session_secret` and `.jwt_secret`.
- `app/main.py` — `.env_overlay` loaded from `_state_dir()`.

### Host scripts
- `scripts/ops/agnes-auto-upgrade.sh` — `STATE_DIR` drives mount-sanity check and cert detection. Defaults preserve existing behavior.
- `scripts/ops/agnes-tls-rotate.sh` — `STATE_DIR` drives `CERT_DIR`.

### New compose overlay
- `docker-compose.flat-mount.yml` — parallel `/data` and `/data-state` binds per service. Mutually exclusive with `docker-compose.host-mount.yml`.

### Documentation
- `docs/state-dir.md` — layout A vs B comparison, pros/cons, migration runbook, which code paths read `STATE_DIR`.

## Why flat layout matters

Nested layout (sdb at `/data`, sdc at `/data/state`) has structural fragility:

1. **Bind propagation matters.** Non-recursive bind silently shadows the nested sdc mount. Caused the 2026-05-05 outage — DuckDB went FATAL on container restart because it had been writing to a shadowed subdirectory of sdb instead of sdc.
2. **Two writers, same prefix.** Host's `tls-rotate.timer` (running as root) writes to `/data/state/certs`; container app (uid 999) writes to `/data/state/system.duckdb`. Same path, different mount-namespace views — ownership conflicts.
3. **sdb resize requires umount of sdc first.** Mount-order coupling.

Flat layout removes all three: each disk is its own bind, recursive by default in Docker 20.10+, no shared prefix, no nested propagation.

## Backward compatibility

`STATE_DIR` defaults to `${DATA_DIR}/state` — current behavior. Existing deployers that don't set the var see no behavior change. Migration to flat layout is opt-in per the runbook in `docs/state-dir.md`.

## Validation

```
$ bash -n scripts/ops/agnes-auto-upgrade.sh && bash -n scripts/ops/agnes-tls-rotate.sh
ok

$ docker compose -f docker-compose.yml -f docker-compose.prod.yml -f docker-compose.flat-mount.yml config | head
services:
  app:
    volumes:
    - type: bind
      source: /data
      target: /data
    - type: bind
      source: /data-state
      target: /data-state
    ...

$ python3 -c "import os; os.environ['STATE_DIR']='/custom-state'; from app.secrets import _state_dir; print(_state_dir())"
/custom-state
```

All 6 services (`app`, `extract`, `scheduler`, `telegram-bot`, `ws-gateway`, `caddy`) bind both disks correctly under the flat overlay.

Tests: full `pytest` requires the test fixture stack (DB, env vars, etc.) — not run here. Unit-level helper functions exercised manually.

## Open question for reviewers

The flat-mount overlay uses `!override` (compose-spec). Same as #191. Confirmed working on Compose v5.1.3 (Groupon deployment). If you need broader compatibility, ping me and I'll explore an alternative.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -323,7 +323,8 @@ def create_app() -> FastAPI:
     app.add_middleware(RequestIdMiddleware)
 
     # Load .env_overlay (persisted by /api/admin/configure)
-    _overlay = Path(os.environ.get("DATA_DIR", "./data")) / "state" / ".env_overlay"
+    from app.secrets import _state_dir
+    _overlay = _state_dir() / ".env_overlay"
     if _overlay.exists():
         for line in _overlay.read_text().splitlines():
             if "=" in line and not line.startswith("#"):
--- a/app/secrets.py
+++ b/app/secrets.py
@@ -7,13 +7,25 @@ from pathlib import Path
 logger = logging.getLogger(__name__)
 
 
+def _state_dir() -> Path:
+    """Return path to writable state directory.
+
+    STATE_DIR env var takes precedence; otherwise defaults to
+    ${DATA_DIR}/state for backward compatibility with deployments
+    that nest state under the data disk. See docs/state-dir.md.
+    """
+    state = os.environ.get("STATE_DIR", "")
+    if state:
+        return Path(state)
+    return Path(os.environ.get("DATA_DIR", "./data")) / "state"
+
+
 def _load_or_generate(env_var: str, file_name: str) -> str:
     """Load secret from env var, or from file, or generate and persist."""
     val = os.environ.get(env_var, "")
     if val:
         return val
-    data_dir = Path(os.environ.get("DATA_DIR", "./data"))
-    secret_path = data_dir / "state" / file_name
+    secret_path = _state_dir() / file_name
     if secret_path.exists():
         val = secret_path.read_text().strip()
         if val:
--- a/docker-compose.flat-mount.yml
+++ b/docker-compose.flat-mount.yml
@@ -1,0 +1,88 @@
+# Flat-mount overlay — parallel host binds for /data and /data-state.
+#
+# Why this overlay
+# ----------------
+# The default deployment topology nests state under data: sdb at /data,
+# sdc at /data/state (i.e. /data/state is a separate disk mounted INSIDE
+# the data disk). That layout works but has known fragility:
+#
+#   - Bind-mount propagation matters. A non-recursive bind hides the
+#     nested mount, leading to silent shadow writes (the failure mode
+#     that caused 2026-05-05 in the Groupon FoundryAI deployment).
+#
+#   - Two writers, one tree. Host-side timers (tls-rotate.timer)
+#     write to /data/state/certs as root, while the container app
+#     writes to /data/state/system.duckdb as uid 999. Same prefix,
+#     different mount-namespace views = ownership conflicts.
+#
+#   - sdb resize requires umounting sdc first. Mount-order coupling.
+#
+# This overlay removes the nesting by mounting the state disk in
+# PARALLEL to the data disk:
+#
+#   sdb at /data       (analytics, regenerable)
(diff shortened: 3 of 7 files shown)


ZdenekSrotyr:
Heads-up on a cross-PR interaction with #188 (combined perf + clarity, currently green and ready to merge):

The `caddy: volumes: !override` in `docker-compose.flat-mount.yml` drops two mounts that #188 adds to the base `docker-compose.yml`:

```yaml
# base after #188 lands:
caddy:
  volumes:
    - ./Caddyfile:/etc/caddy/Caddyfile:ro
    - /data/state/certs:/certs:ro
    - caddy_data:/data
    - caddy_config:/config       # <-- dropped by !override
    - data:/srv:ro                # <-- dropped by !override (Caddy file_server parquet mount)
```

Concrete impact under the flat layout:

1. **`data:/srv:ro` missing** ⇒ Caddy's `try_files /bigquery/data/{tid}.parquet …` in the new `@download` handler can't see the parquets, so every `/api/data/{tid}/download` request silently falls through to the `reverse_proxy app:8000` line. That re-stars the single-uvicorn-worker bottleneck #188 was specifically built to remove (a multi-GB analyst pull starves `/api/health` and the UI for the duration of the stream). The parquets are still where the app expects them, so functionally everything works — just without the perf win.

2. **`caddy_config:/config` missing** ⇒ Caddy loses its ACME-state / autosave dir between recreates. Not catastrophic, but on each container restart it's a re-issuance roll of the dice for HTTPS deployments.

I think you wrote this against `main` before #188 landed (your PR predates the base-compose change by ~10 min), which is why these two mounts aren't reflected. After #188 merges, please rebase and either:

- (a) add `- data:/srv:ro` and `- caddy_config:/config` to the `caddy: volumes: !override` block, or
- (b) drop the `caddy:` override entirely if the only thing you actually need to change is `/data-state/certs:/certs:ro` — that one mount could move to a `services.caddy.volumes:` *additive* form using anchors, but explicit `!override` with the full list is probably clearer.

(Same overlay class as #191 in the rest of the file, but #191 doesn't touch caddy so it auto-inherits the new mounts. flat-mount is the only one with the gotcha.)

Also: a few CI jobs (`build-and-push`, `smoke-test`, `e2e-bind-mount`) are CANCELLED on the latest push — looks like a re-trigger race rather than a real failure. A push or workflow re-run should clear it.

ZdenekSrotyr:
Cherry-picked into #194 (rebase on post-#188 main; caddy `!override` regression I flagged earlier is fixed there). Closing as superseded once #194 merges.

ZdenekSrotyr:
Superseded by #194 — content cherry-picked verbatim onto post-#188 main with attribution preserved (commit `a303de03`); the caddy `!override` regression I flagged is fixed in commit `a9ae5f9c` of #194.

PR #194, opened 2026-05-05
host-mount + STATE_DIR: kill the shadow-mount class once and for all

ZdenekSrotyr:
Combines [#191](https://github.com/keboola/agnes-the-ai-analyst/pull/191) and [#192](https://github.com/keboola/agnes-the-ai-analyst/pull/192) (both by @cvrysanek, original commits cherry-picked with attribution preserved) into one rebase-on-`main` PR + a regression fix found during review.

### 1. `docker-compose.host-mount.yml` — direct service binds (was #191)

Replaces `volumes.data.driver_opts` (immutable after volume creation) with direct service-level `volumes: !override` per service. Docker named-volume options are fixed at create time, so editing this file's `bind` → `bind,rbind` does NOT propagate to existing `agnes_data` volumes — exactly the failure mode that hit a deployer on 2026-05-05 (containers wrote to a shadowed subdir of the parent disk instead of the nested `/data/state` mount; DuckDB went FATAL on a root-owned WAL during a routine container recreate).

Direct service binds re-evaluate options every container start and default to recursive in modern Docker. No immutable state, no shadow class. The `data:` named volume declaration in `docker-compose.yml` stays intact for local-dev `compose up` without this overlay.

Doesn't override `caddy:` — picks up the new `data:/srv:ro` mount (v0.36.0) automatically.

### 2. `STATE_DIR` env var + `docker-compose.flat-mount.yml` (was #192)

`STATE_DIR` is the canonical handle for the writable state directory. Defaults to `\${DATA_DIR}/state` (full backward compat); set `STATE_DIR=/data-state` to point at the new flat layout.

Read by:
- `src/db.py:_get_state_dir()` — used by `get_system_db()` and the schema migration snapshot path.
- `app/secrets.py:_state_dir()` — for `.session_secret`, `.jwt_secret`. Mirrors the helper since `app/` shouldn't import from `src/`.
- `app/main.py` — for the `.env_overlay` startup file.
- `scripts/ops/agnes-auto-upgrade.sh` — mount-sanity check + cert detection.
- `scripts/ops/agnes-tls-rotate.sh` — `CERT_DIR=\$STATE_DIR/certs`.

`docker-compose.flat-mount.yml` mounts `/data` and `/data-state` in **parallel** via direct service binds. Mutually exclusive with `host-mount.yml` — operator picks one based on disk topology.

`docs/state-dir.md` covers Layout A (nested, default) vs Layout B (flat) with a step-by-step migration runbook from A → B.

### 3. Regression fix found during review of #192 (new commit on top)

`flat-mount.yml`'s `caddy: volumes: !override` block listed only three mounts but the base `docker-compose.yml` caddy service has five. `!override` replaces the entire list, so two mounts were silently dropped under the flat layout:

- **`data:/srv:ro`** — Caddy's read-only view of the agnes data dir, used by the `@download` `file_server` handler added in v0.36.0 as the perf bypass for multi-GB parquet downloads. Without it, `try_files /bigquery/data/<id>.parquet …` finds no file and every parquet download falls through to the app's uvicorn worker — defeating the bypass entirely.
- **`caddy_config:/config`** — Caddy's autosave/ACME state. Less critical (we feed certs in via `/certs`), but loses the autosaved adapter config across container recreates.

Restated both mounts in `flat-mount.yml` with a comment block explaining the `!override` caveat for any future overlay author.

## CHANGELOG

Both items added under `## [Unreleased]`:
- `### Added` — STATE_DIR + flat-mount overlay
- `### Changed` — host-mount.yml direct binds

## Validation

```bash
$ docker compose -f docker-compose.yml -f docker-compose.prod.yml -f docker-compose.host-mount.yml config | grep -A4 "caddy:" | head
caddy:
  volumes:
    - …Caddyfile:ro
    - /data/state/certs:/certs:ro
    - data → /srv:ro            # ← inherited from base
    - caddy_data:/data
    - caddy_config:/config      # ← inherited from base

$ docker compose -f docker-compose.yml -f docker-compose.prod.yml -f docker-compose.flat-mount.yml config | grep -A6 "caddy:" | head
caddy:
  volumes:
    - …Caddyfile:ro
    - /data-state/certs:/certs:ro
    - caddy_data:/data
    - caddy_config:/config      # ← restated by this PR
    - /data:/srv:ro              # ← restated by this PR
```

```bash
$ bash -n scripts/ops/agnes-auto-upgrade.sh && bash -n scripts/ops/agnes-tls-rotate.sh
ok

$ python3 -c "import os; os.environ['STATE_DIR']='/custom-state'; from app.secrets import _state_dir; print(_state_dir())"
/custom-state
```

## Migration notes for operators

| Layout | Status | What to do |
|---|---|---|
| **A — nested** (sdc at `/data/state`) | Default. Backward compatible. | Nothing. Auto-upgrade handles this PR's `host-mount.yml` fix transparently on the next 5-min tick. |
| **B — flat** (sdc at `/data-state`) | Opt-in. New. | Follow `docs/state-dir.md` migration runbook (stop containers, snapshot state, remount sdc at `/data-state`, set `STATE_DIR=/data-state` in `.env`, switch overlay to `flat-mount.yml`, bring stack back up). |

`scripts/ops/agnes-auto-upgrade.sh` always loads `host-mount.yml` from main today — it's the canonical overlay. Operators on flat layout drive `docker compose` manually (or replace the cron with their own variant). Tracked as a follow-up if there's appetite to make the cron `STATE_DIR`-aware.

## Closes

- Closes #191 (cherry-picked, content preserved)
- Closes #192 (cherry-picked, content preserved + caddy regression fixed)

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/194" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


cvrysanek (approved):
LGTM with thanks for the caddy override regression catch — `data:/srv:ro` is exactly the kind of mount that's silently load-bearing (the v0.36.0 parquet bypass collapses to "every download hits uvicorn" without it, no error visible until p95 latency moves). I should have run `docker compose config` against the rebased base before writing the override; lesson logged on my end.

The `!override` "diff against base before merging" comment is the right shape for preventing recurrence. Better than my original.

## Follow-up suggestion (non-blocking)

`host-mount.yml` currently doesn't override `caddy:`, so the trap doesn't fire there today. But a future overlay author who adds a caddy override there for any reason would re-introduce the same bug class. One-line pointer in `host-mount.yml` referencing the flat-mount.yml comment ("if you add a `services.<svc>: volumes: !override` block here, see flat-mount.yml for the restate-every-mount caveat") would future-proof it.

Could fold into this PR as a one-liner, or open a small follow-up after merge — either works.

## Ready to ship from my side

Once merged + a `stable-2026.05.NN` tag carries it:
1. Groupon FoundryAI lands via `agnes_tag` bump in `locals.tf`.
2. Reconcile picks up the new `host-mount.yml` (option B) — closes the immutable-volume-options bug class permanently. No per-VM action needed beyond `docker volume rm agnes_data` for the orphaned old volume.
3. Optional follow-up: per-VM migration to flat layout (option A) using `docs/state-dir.md` runbook. Schedule per environment criticality.

PR sequence on the deployer side already merged: `validate-vm.sh` (post-apply 9-check validation), `fix-shadow-mount.sh` (one-shot recovery for affected VMs), and the 5 hackathon VMs already migrated off the broken `bind` volume.

PR #210, opened 2026-05-07
feat(marketplace): clone-based plugin setup + SessionStart auto-refresh

minasarustamyan:
## Summary

End-to-end flow for installing and keeping the per-user filtered Claude Code marketplace in sync with the user's Agnes stack (`admin RBAC grants ∖ MyAIStack opt-outs ∪ /store installs`).

- **Setup**: install prompt step 5 collapses to `agnes refresh-marketplace --bootstrap` — clones the per-user marketplace bare repo to `~/.agnes/marketplace`, strips the PAT from the cloned origin URL, registers the local path with Claude Code, installs every plugin in the served manifest at `--scope project`. Replaces a 15-line inline shell sequence that tripped Claude Code's agent `rm -rf` permission gate.
- **Auto-refresh**: SessionStart hook (installed by `agnes init`) runs `agnes refresh-marketplace --quiet` every Claude Code session — fetches+resets the clone, then version-aware reconciles (install missing / update on diff / skip on match). Both install and update use `--scope project` so changes stay in the workspace the hook fired in. Doesn't auto-uninstall plugins removed from the manifest (transient empty manifest from server would wipe the stack).
- **Notification**: when something changed, emits Claude Code hook JSON on stdout — `systemMessage` (transient toast) and `additionalContext` (model-side system reminder), both pointing at `/reload-plugins` to load the new plugins into the running session without a restart.
- **Windows fix**: refresh-marketplace hook command wrapped in `bash -c "..."` so `2>/dev/null || true` is interpreted on Windows where Claude Code runs hooks directly without a shell.
- **Reset script**: cleans `~/.claude/plugins/{marketplaces,cache}/agnes`, drops uv build cache, documents workspace-scoped residue that can't be enumerated.

## Test plan

- [x] `pytest tests/test_cli_refresh_marketplace.py` — 28 tests covering bootstrap, fetch+reset ordering, version-aware reconcile, project-path filtering, hook JSON shape, `/reload-plugins` hint, scope-project install/update.
- [x] `pytest tests/test_lib_hooks.py` — two-entry SessionStart layout, bash-c wrapper invariant, idempotent re-install, legacy `da sync` upgrade.
- [x] `pytest tests/test_setup_instructions.py tests/test_setup_hooks_template.py tests/test_setup_page_unified.py` — install prompt one-liner shape, no inline `rm -rf` / `git clone` / per-plugin install lines.
- [ ] Real-machine smoke: paste install prompt into fresh Claude Code workspace; verify `agnes refresh-marketplace --bootstrap` clones + installs plugins; verify SessionStart hook propagates a server-side plugin add on the next session.
- [ ] Windows: verify the `bash -c "..."` wrapper actually works under Git Bash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,46 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+## [0.44.0] — 2026-05-07
+
+### Added
+- `agnes refresh-marketplace` — single CLI command that owns the per-user
+  filtered Claude Code marketplace lifecycle. `--bootstrap` does the
+  first-time setup: clones the per-user marketplace bare repo to
+  `~/.agnes/marketplace`, strips the PAT from the cloned origin URL so it
+  doesn't sit in plaintext at rest, registers the local path with Claude
+  Code, and installs every plugin in the served manifest at
+  `--scope project`. Without `--bootstrap` it does an incremental refresh:
+  fetch + reset to the remote, then version-aware reconcile (install missing
+  plugins, update on version diff, skip on match). Plugins removed from the
+  manifest are deliberately NOT auto-uninstalled — a transient empty manifest
+  from the server would otherwise wipe the user's stack.
+- `agnes init` now installs a SessionStart hook that runs
+  `agnes refresh-marketplace --quiet` on every Claude Code session,
+  alongside the existing chained `agnes self-upgrade; agnes pull` entry.
+  The marketplace refresh runs as a *separate* hook entry (not chained)
+  so a failure (e.g. fresh workspace with no clone yet) doesn't suppress
+  the data pull. The refresh command is wrapped in `bash -c "..."`
+  because Claude Code on Windows runs hook commands directly without a
+  shell, which would otherwise leave the `2>/dev/null || true` syntax
+  uninterpreted.
+- When `agnes refresh-marketplace` detects an actual change, it emits
+  Claude Code hook JSON on stdout — `systemMessage` (transient toast)
+  and `additionalContext` (model-side system reminder) — both pointing
+  at `/reload-plugins` so the running session loads new plugins without
+  a restart.
+
+### Changed
+- Install-prompt step 5 (in the dashboard-served setup payload) collapses
+  from a 15-line inline shell sequence — `rm -rf` + `git clone` + per-plugin
+  `claude plugin install` calls — to a single `agnes refresh-marketplace
+  --bootstrap` invocation. The old inline form tripped Claude Code's agent
+  `rm -rf` permission gate on first run.
+- `scripts/dev/agnes-client-reset.sh`: now cleans
+  `~/.claude/plugins/{marketplaces,cache}/agnes`, drops the uv build cache,
+  and documents workspace-scoped residue that can't be enumerated from a
+  user-level reset.
+
 ### Internal
 
 - `infra/modules/customer-instance` (tag `infra-v1.7.0`): `google_compute_instance.vm` now sets `allow_stopping_for_update = true`. Without it, changing `machine_type` (or any other field GCP will only mutate on a stopped VM) caused Terraform to fall back to a destroy + recreate, churning VM-local state for what should be an in-place resize. Consumers do not need to update — the field is provider-side only — but bumping the module ref to `infra-v1.7.0` enables in-place machine-type bumps.
--- a/app/web/setup_instructions.py
+++ b/app/web/setup_instructions.py
@@ -58,17 +58,30 @@ practice and the design here exists to dodge each one:
    `update-ca-trust`) doesn't fix it on Windows or macOS either — the
    binary's bundled CA list isn't refreshable from the OS store.
 
-   So the marketplace step branches on platform:
-     - Windows + macOS → straight to system-`git clone` fallback
-       (system git honors `GIT_SSL_CAINFO`, so the clone works).
-     - Linux → typically the node-based npm install where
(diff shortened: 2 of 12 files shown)


ZdenekSrotyr (commented):
## Manual review (in lieu of Devin)

Read every changed file end-to-end. Ran `pytest tests/test_lib_hooks.py tests/test_cli_refresh_marketplace.py tests/test_setup_instructions.py tests/test_setup_page_unified.py` — all green (82 tests after my tightening commit). CI is green across `CI/test`, `CI/docker-build`, `Release/test`, `Release/build-and-push`. Branch was rebased onto current `main` (resolving the `cli/lib/hooks.py` and `tests/test_lib_hooks.py` conflicts with v0.43.0's chained `agnes self-upgrade; agnes pull` SessionStart layout) and a `release: 0.44.0` commit was appended.

### What I like

- **Single source of truth.** `cli/lib/marketplace.py` centralizes `CLONE_DIR` + `MARKETPLACE_NAME`. Both setup-instructions and the refresh CLI consume those constants — drift would otherwise silently break the post-bootstrap refresh path.
- **Pre-snapshot before `claude plugin marketplace update`.** The comment at `cli/commands/refresh_marketplace.py:100-105` explains the subtlety well: Claude silently auto-applies version bumps when re-reading the manifest, so an after-snapshot would mis-report "no change" on every real bump. Snapshotting installed versions BEFORE the manifest update is the only way to detect updates reliably.
- **Three call paths share one code.** Bootstrap, manual refresh, and quiet-hook all funnel through the same fetch+reset+reconcile primitives. The `--bootstrap` flag conditionally inserts the clone step at the front; everything else is identical. Nice.
- **Don't auto-uninstall on disappearance** is the right call (`_reconcile_with_manifest` docstring). A transient empty manifest from the server otherwise wipes the user's stack and there's no recovery without re-running setup.
- **No inline `rm -rf`** — wrapping the destructive prep in `agnes refresh-marketplace --bootstrap` lets the CLI's permission grant cover `shutil.rmtree`, sidestepping Claude Code's `rm -rf` denylist. Good UX win.
- **PAT hygiene.** Bootstrap clones via authenticated URL once, then `git remote set-url` strips the PAT from the on-disk `origin`. Subsequent fetches use the per-invocation `_CREDENTIAL_HELPER` reading `$AGNES_TOKEN` from the subprocess env — never on the command line, never persisted to `.git/config`. Plus chmod 700/600 on POSIX. Solid.
- **Tests are extensive** (28 in `test_cli_refresh_marketplace.py` + 9→10 in `test_lib_hooks.py` after my addition) and cover the gnarly bits: project-path filtering so sibling workspaces don't bleed into reconcile; third-party-marketplace skip; the "claude silently auto-updated, snapshot caught it anyway" path.

### Findings

**1. (Was missing — fixed in `1c37707e`) Upgrade path from v0.43.0's single chained entry → v0.44.0's two-entry layout had no test.**

After my rebase the SessionStart layout went from "1 chained entry (`self-upgrade ; pull`)" (v0.43.0) to "2 entries (chain + refresh-marketplace)" (this PR). `test_install_replaces_prior_single_pull_entry` covers the bare-`agnes pull`-only case and `test_install_replaces_old_da_sync_entries` covers `da sync`, but neither covers the v0.43.0 chained shape — which is what every workspace bootstrapped on the previously-released CLI carries. Without that, an `agnes init` on an upgraded workspace might stack the v0.44 entries on top of the v0.43 chained one, re-running self-upgrade twice per session and leaving the old format around forever. Pinned by `test_install_replaces_v0_43_chained_self_upgrade_pull_entry` in my push.

**2. (Nit) `_list_installed_agnes_plugins_in_cwd` returns `None` on the FIRST non-dict entry.**

`cli/commands/refresh_marketplace.py:471-473`:

```python
for entry in payload:
    if not isinstance(entry, dict):
        return None
```

If a future Claude Code release adds a non-dict element (e.g. a header / metadata entry) to `claude plugin list --json`, the whole reconcile gets skipped — silently — because the early `return None` short-circuits the loop and the caller treats `None` as "could not enumerate". `continue` would skip the bad entry but keep collecting the good ones; that's the more conservative posture against schema drift.

Counter-argument: `return None` makes the failure loud (caller's `warn: could not enumerate installed plugins; skipping reconcile.` fires), whereas `continue` would silently pretend partial state is the truth. Either reading is defensible; flagging so the choice is intentional.

**3. (Nit) `_bootstrap_clone` silently skips `claude plugin marketplace add` when `claude` isn't on PATH.**

`cli/commands/refresh_marketplace.py:197`:

```python
if shutil.which("claude") is not None:
    add = subprocess.run([...])
    ...
```

Bootstrap then echoes `Marketplace bootstrapped at {CLONE_DIR}.` regardless. A user who runs `agnes refresh-marketplace --bootstrap` before installing `claude` gets a clone but no Claude-side registration, and no warning that anything was skipped. Setup-instructions step 4 makes the user verify `claude --version` before this step, so this scenario is mostly hypothetical. But a one-line `warn: claude not on PATH; skipping marketplace registration — install Claude Code, then re-run` would close the loop.

**4. (Nit, doc-only) `_emit_hook_message` hardcodes `hookEventName: SessionStart`.**

`cli/commands/refresh_marketplace.py:407`. Today the command is only invoked from SessionStart hooks, so the hardcoding is correct. A one-line comment ("invoked only from SessionStart hooks; if added to other events the wired event name needs to follow") would help a future reader if the trigger ever expands.

### Verdict

**Approve with the note that finding 1 has been pinned in `1c37707e` and the rest are nits the author can address in this PR or defer.** The CHANGELOG + `pyproject.toml` bump to 0.44.0 are in `0c9d0386` (the release cut commit appended after the rebase). Ready to merge from my side.

PR #217, opened 2026-05-07
Keboola connector v26 — incremental, partitioned, where_filters + typed parquet

ZdenekSrotyr:
## Summary

Brings the Keboola connector to feature parity with the legacy internal data-analyst's per-table sync strategies. Closes the four documented gaps from the spec branch (`zs/keboola-connector-specs`):

- **Typed parquet** in the legacy SDK extraction path — column types from Keboola Storage metadata (provider cascade `user > ai-metadata-enrichment > keboola.snowflake-transformation`) survive the CSV → parquet roundtrip; invalid date strings (`'0000-00-00'`) and invalid numeric strings (`'Non-Manager'`) become NULL while keeping the column's typed schema. Pre-fix everything was VARCHAR.
- **Incremental sync** via Storage API `changedSince` — opt-in per table; pulls only delta rows, merges into the existing parquet by `primary_key` (drop_duplicates with keep='last'). Cuts daily extraction from O(full table) to O(delta).
- **Partitioned sync** — flat per-partition layout `data/<table>/<key>.parquet` (e.g. `2026_05.parquet`), per-affected-partition merge for daily updates, chunked initial load with 1-day overlap and 2-empty-chunk stop heuristic.
- **`where_filters`** — server-side row filter with date placeholders (`{{today}}`, `{{last_3_months}}`, `{{start_of_3_months_ago}}`, etc.) resolved at sync time. Force the SDK path; reject `incremental + where_filters` combination at API layer (changedSince already filters temporally).

## Architecture

- **Schema migration v25 → v26**: 7 new columns on `table_registry`. Existing `sync_strategy` column reused (pre-v26 it was inert catalog metadata; post-v26 the extractor dispatches off it).
- **Per-table dispatcher** in `extractor.run()` routes to one of `_extract_via_extension` (full_refresh + extension), `_extract_via_legacy` (full_refresh + filters or extension fallback), `extract_incremental`, or `extract_partitioned`.
- **API conflict policy**: `incremental + where_filters` → 422; `partitioned + query_mode='remote'` → 422; `partitioned ⇒ partition_by required`.
- **Admin UI**: third "Direct extract (Storage API)" radio in the Keboola Register / Edit modals, alongside existing "Whole table (extension)" and "Custom SQL". When selected, exposes a v26 sync-strategy panel with conditional fields per strategy.

## Test plan

- [x] **Unit + module** — 134 v26 tests covering migration, repo, parquet_io, where_filters, incremental (compute_changed_since + merge_parquet + extract_incremental E2E), partitioned (key derivation + merge_partition + chunked windows + extract_partitioned E2E), extractor dispatcher, admin API validators, PUT field clearing, registry-shape → dispatcher bridge
- [x] **HTML form structure** — all v26 inputs + visibility classes + JS payload fields verified in rendered template
- [x] **Real Keboola roundtrip** — registered a small test table as `sync_strategy='incremental'` against a test Storage project, triggered two syncs:
  - Sync 1: `changedSince=None` → full pull → 9 rows typed parquet
  - Sync 2: `changedSince=last_sync - 1d window` → 9 delta rows merged with 9 existing → 9 after dedup on primary_key (PK merge confirmed)
- [x] **Browser UX** — agent-browser session against a local uvicorn: login → admin/tables → register modal → switch radios → verify field visibility per strategy → submit → edit existing row → switch to Direct/Incremental → save → confirm DB persistence
- [x] **Regression** — no regressions in the broader 3252-test suite (3 pre-v26 tests updated for the deprecation-marker removal + schema-version bump; 2 pre-existing environment-sensitive test failures unrelated to this change)

## Bugs caught + fixed during E2E

The browser + real-Keboola roundtrip exposed four bugs the unit tests missed:

1. **JS visibility race** — two competing `forEach` loops set `display=''` then `display='none'` on form elements sharing `kb-strategy-incremental kb-strategy-partitioned` classes (window_days + max_history_days are reused across strategies). Fix: single-pass selector with class-based visibility resolver.
2. **PUT cannot clear field** — pre-v26 `updates = {k: v ... if v is not None}` collapsed "omitted from body" and "sent as null" into the same case, so admin couldn't switch a partitioned row back to full_refresh and have stale `partition_by` clear. Fix: `model_dump(exclude_unset=True)`.
3. **Subprocess DB lock conflict** — `_read_last_sync` reopened `system.duckdb` while the parent server held the write lock (subprocess contract at `app/api/sync.py:_run_sync` line 260). Fix: parent injects `__last_sync__` into table_config before subprocess spawn.
4. **Wrong KBC table_id** — `extract_incremental` / `extract_partitioned` built the Storage API table_id from the registry row's slugified `id` (`circle_inc`) instead of `bucket.source_table` (`in.c-finance.circle`), producing 404s. Fix: prefer `bucket+source_table`; fall back to `id` only when bucket empty.

## Operator notes

- Existing tables stay on `full_refresh` after migration; admins opt individual tables in via `agnes admin register-table --sync-strategy ...`, the Keboola Edit modal, or `POST/PUT /api/admin/registry`.
- `merge_parquet` and `merge_partition` use `pd.concat + drop_duplicates`, loading both existing and delta into pandas RAM. For tables in the multi-million-row range this may OOM — switch to `partitioned` strategy for those (per-partition merge keeps memory bounded). Documented in `### Internal` of the changelog entry.
- Date placeholders are resolved at **sync time**, not register time — a typo'd `{{lasst_week}}` is accepted at register and surfaces only when the next sync runs. By design (rolling windows need late-binding).

## Spec source

The four corresponding plans on the `zs/keboola-connector-specs` branch under `docs/superpowers/plans/2026-05-07-0[1-4]-*.md` capture the design rationale and link back to internal repo references for each subsystem.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/217" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ZdenekSrotyr (commented):
Read the full diff at `f7f60222` (5,317 lines), plus migrations, dispatcher, validators, modules, and tests. Filtered through the contract this PR claims (additive, opt-in, `full_refresh` stays default).

## Blockers

**Schema-version collision with `main`.** `main` already shipped v26 in v0.46.0 (merge `f1561a67`): `_V25_TO_V26_MIGRATIONS` exists at `src/db.py:1824` on main and rewrites Keboola `query_mode='local' → 'materialized'`. This PR also defines `_V25_TO_V26_MIGRATIONS` at `src/db.py:1837` for a different purpose (7 `table_registry` columns) and bumps `SCHEMA_VERSION = 26`. After rebase one of:

- The PR's columns become v27 (`_V26_TO_V27_MIGRATIONS`), `SCHEMA_VERSION = 27`. Auto-migration loop at `src/db.py:2099-2102` needs a new branch.
- Merge both into one v26 list — riskier; the materialized-query rewrite must run BEFORE the column adds (or doesn't matter), the rebase author has to verify ordering.

`tests/test_keboola_v26_migration.py` and `tests/test_db_schema_version.py` both pin to `SCHEMA_VERSION == 26` and need updates either way. The `CLAUDE.md` rewrite (`Schema v26 with auto-migration v1→…→v26`) silently drops main's v26 description — needs both v26's documented.

## Concerns

**No memory guard on `merge_parquet` / `merge_partition`.** `connectors/keboola/incremental.py:102` (`pq.read_table(existing_parquet).to_pandas()`) and `connectors/keboola/partitioned.py:111` load the full existing parquet into RAM. The OOM caveat lives only in `CHANGELOG.md ### Internal` and the docstring. There's no row-count check, no log warning at the threshold, no API-time advisory when a table opts into `incremental` without `partition_by`. Suggest: log `WARNING` when existing_df row count crosses ~5M — operator gets a signal at the right time (the merge IS happening) instead of post-mortem after an OOM kill.

**Stale `### Internal` claim.** CHANGELOG says "HTML form template not updated to surface the new fields. Admin currently uses `agnes admin register-table` flags…" but the diff at `app/web/templates/admin_tables.html` lines 1360-1405, 1563-1600 DOES update the form. Trim the bullet to avoid misleading operators.

**`where_filters` silently dropped on `query_mode='remote'`.** `connectors/keboola/extractor.py:294-396` resolves `where_filters` for any strategy, but `_extract_via_extension` (the remote/extension path) doesn't take a filters argument. The `RegisterTableRequest` validator does NOT reject `query_mode='remote' + where_filters`. Result: admin sets filters on a remote Keboola row, registry persists them, next sync ignores them, no error. Either reject in the validator or extend `_extract_via_extension` to apply them.

**`partitioned + where_filters` is allowed but possibly silently dropped.** `tests/test_admin_register_v26_validation.py:125` (`test_partitioned_with_where_filters_accepted`) confirms it passes validation. Yet the dispatcher branches into `extract_partitioned` at `connectors/keboola/extractor.py:353-388` — does it actually pass `resolved_filters` through to its chunked downloads? Worth a code path check + an end-to-end test that asserts the filter actually narrowed the partition data.

## Nits

- `connectors/keboola/extractor.py:296` — `resolved_filters` initialized then reset inside the `if raw_filters:` block; the comment "Resolve where_filters once for any strategy that supports them" is misleading because two strategies (remote, possibly partitioned) ignore them.
- `connectors/keboola/where_filters.py:104` — `replace`-loop on placeholder substitution. `{{last_year}}` and `{{last_2_years}}` share a prefix; `frozenset` iteration order is non-deterministic. Substring shadowing isn't an actual problem with the current placeholder set, but sort by length descending in the loop to make the contract explicit.
- `app/api/admin.py:1281` and `:1791` — `_check_mode_query_coherence` for `RegisterTableRequest` and the same-name validator on `UpdateTableRequest` are nearly identical except the PUT version handles None-merging. Extract to a shared helper to avoid drift.
- `connectors/keboola/partitioned.py:65-77` — `partition_key_for` accepts `pd.Timestamp` via `to_pydatetime()`, then `datetime`, then `date`. `pd.Timestamp` is a `datetime` subclass; the pandas branch is redundant. Also check `datetime` BEFORE `date` (date is parent in some Python versions).

## What's good

- **Test density is exceptional.** ~2,330 lines across 12 new test files for ~1,500 lines of new production code. The Devin Review finding 0004 (mixed-dtype PK dedup at `incremental.py:114-128`) shows the kind of nuance the test suite drove out.
- **Migration is genuinely additive** — `_V25_TO_V26_MIGRATIONS` is 7 `ALTER TABLE … ADD COLUMN IF NOT EXISTS` statements, idempotent, no data rewrites. Pre-existing `full_refresh` rows survive untouched (`tests/test_keboola_v26_migration.py:33`).
- **API conflict policy is operator-actionable.** The 422 messages at `app/api/admin.py:1422-1437` explain WHY the combination is rejected ("changedSince already filters temporally"), not just THAT it is.
- **Security**: `where_filters` operator allowlist + placeholder allowlist are both `frozenset`s, rejected before they leave Pydantic. Form-encoded payload = no SQLi vector. Identifier validation reused (`_is_safe_identifier`, `_is_safe_quoted_identifier`).
- **PUT semantics fix at `app/api/admin.py:2589`** (`exclude_unset=True`) is the right call; the comment captures the contract change clearly. Existing JS frontend continues to work because omitted fields preserve stored values.
- **Date-placeholder safety**: strict allowlist (`SUPPORTED_PLACEHOLDERS` frozenset). After substitution, leftover `{{` raises `InvalidFilterError`. Resolved values flow into form-encoded `whereFilters[i][values][j]` POST params — never touch SQL. Not a SQLi vector.
- **CHANGELOG discipline** — `### Added`, `### Fixed`, `### Changed`, `### Internal` all populated, schema bump called out (modulo the stale form-template note).

## Open questions

1. **Rebase plan for the v26 collision?** (a) Become v27 with fresh `_V26_TO_V27_MIGRATIONS` block, or (b) merge into existing v26 on main. (a) is safer and reversible.
2. **Does `extract_partitioned` actually consume `resolved_filters`?** If not, `test_partitioned_with_where_filters_accepted` is a false positive — admin can register the combination but filters get silently ignored at sync time.
3. **`query_mode='remote' + where_filters` is currently accepted.** Intended? If so, where in the dispatcher do remote-mode filters get honored? If unintended, add a 422.
4. **Rollback plan for the v26 migration?** If a deployment lands the column adds then needs to revert to v0.46.0, the schema-version downgrade at `src/db.py:2116-2117` is a no-op and the new columns linger as harmless NULLs. Worth documenting in CHANGELOG.

ZdenekSrotyr:
Thanks @devin-ai-integration. Status of the 3 surfaced findings:

1. CLAUDE.md schema docs — already updated to v27 (covering both v25→v26 and v26→v27) in the PR head commit a39b71ea. Verified at line 468.

2. `exclude_unset=True` is intentional. The PR head already includes (a) the inline comment block at `app/api/admin.py` lines 2609-2627 explaining the semantic shift (omitted vs explicit-null), and (b) a behavioral test `test_api_put_clears_v26_fields_on_strategy_switch` in `tests/test_v26_keboola_e2e.py` that pins the explicit-null-clears contract, plus the pre-existing `test_admin_put_preservation.py` that pins the omitted-keep-existing contract. The Edit modal needs the clear path to switch a partitioned row back to full_refresh without leaving stale partition_by lingering.

3. PK dtype conversion failure raises hard — code fix was already in place in the PR head at `connectors/keboola/incremental.py` lines 127-144. Added two regression tests in 4111b41b: `test_merge_pk_dtype_conversion_failure_raises_hard` (PK fail → RuntimeError) and `test_merge_non_pk_dtype_conversion_failure_warns_and_continues` (non-PK fail → warn + continue), pinning the asymmetric policy.

The 6 additional findings in the portal — could you re-post them inline so we can see them in GitHub?

PR #230, opened 2026-05-08
Add /marketplace browse page + Model B opt-in stack composition

minasarustamyan:
## Summary

- New unified `/marketplace` browse surface with three tabs — **Curated** / **Flea** / **My Stack** — per-tab category filter, cross-source search with scope checkboxes, and URL-driven numeric pagination. Plugin and nested skill/agent detail pages with kind-tinted accents and a copyable `/<plugin>:<inner-name>` invocation chip.
- **Model B opt-in**: an RBAC grant on a curated plugin is now only *eligibility*. The user must click "Add to my stack" for it to enter their served Claude Code marketplace. Composition flips from `(rbac ∖ opt_outs) ∪ store_installs` to `(rbac ∩ subscriptions) ∪ store_installs`. The legacy `user_plugin_optouts` table is repurposed as `user_curated_subscriptions` (same shape, inverted semantic; repo methods become `subscribe` / `unsubscribe` / `is_subscribed`).
- Schema **v27 → v28**: idempotent migration wipes existing `user_plugin_optouts` rows on flip; adds `marketplace_plugins.created_at` with `registered_at` backfill. New API surface under `/api/marketplace` (`/items`, `/categories`, plugin detail, subscribe toggle, nested skill/agent detail). Navbar "My AI Stack" → "My Stack" → `/marketplace?tab=my`; standalone `/store` link removed (still reachable via Flea +Upload). Old `/my-ai-stack` and `/store` bookmarks continue to work.

## Test plan

- [ ] `pytest tests/test_marketplace_api.py tests/test_marketplace_filter_store.py tests/test_store_api.py tests/test_store_repositories.py tests/test_setup_page_unified.py tests/test_db_schema_version.py -v`
- [ ] Fresh DB → boots cleanly at v28 (no ALTER TABLE conflict on `marketplace_plugins.created_at`)
- [ ] Existing instance at v27 → migrates to v28, `user_plugin_optouts` rows wiped, `created_at` backfilled from `registered_at`
- [ ] `/marketplace` loads, all three tabs render with correct counts; category filter and search interact with URL state
- [ ] Curated plugin → "Add to my stack" subscribes; "In your stack" unsubscribes; served marketplace ZIP/git reflects the toggle
- [ ] User with RBAC grant but no subscription → plugin visible in Curated tab but absent from served marketplace
- [ ] Skill/agent detail pages render with correct kind-tinted accents and the `/<plugin>:<inner-name>` invocation chip
- [ ] Old `/my-ai-stack` and `/store` URLs still resolve


rymain (commented):
_Superseded — see review under @cvrysanek._

app/api/marketplace.py
```diff
@@ -0,0 +1,1158 @@
...
+    *,
+    is_dir_layout: bool,
+) -> Optional[tuple[str, str]]:
+    """Return (text, relpath) for a skill (sub='skills', is_dir=True) or
+    agent (sub='agents', is_dir=False) inside the plugin tree, or None if
+    the file is missing.
+    """
+    base = plugin_root / sub
+    if is_dir_layout:
+        candidate = base / name / "SKILL.md"
```
cvrysanek: 🔴 **Path traversal — needs fix before merge.**

`skill_name` (and `agent_name` two lines below) come from URL path params and join into `plugin_root` here without a containment check. The `relative_to` call at the end is in the success-path display logic only — not a gate.

FastAPI single-segment `{skill_name}` blocks raw `/`, and `require_resource_access` ensures the caller holds an RBAC grant on **this** parent plugin. But the resolver doesn't verify the final candidate stays inside this `plugin_root`, so a user with a grant on plugin A can potentially read files inside other plugins/marketplaces under `get_marketplaces_dir()` they have no grant for.

Minimal fix:

```python
try:
    resolved = candidate.resolve(strict=True)
    resolved.relative_to(plugin_root.resolve())
except (OSError, ValueError):
    return None
```

Applies symmetrically to `cand_agent` on line 1044.

app/api/marketplace.py
```diff
@@ -0,0 +1,1158 @@
...
+    action: str,
+    target: str,
+    params: Optional[dict] = None,
+) -> None:
+    try:
+        AuditRepository(conn).log(
+            user_id=actor_id, action=action, resource=target, params=params
+        )
+    except Exception:
+        pass
```
cvrysanek, line 224: 🟡 `_audit` swallows all exceptions silently. Compare to `_invalidate_etag` right below which uses `logger.exception(...)`. Lost audit telemetry = lost forensics on the very actions (subscribe/uninstall/grant changes) we most want to be able to reconstruct after the fact. At least log it:

```python
except Exception:
    logger.exception("audit log failed: action=%s target=%s", action, target)
```

app/api/marketplace.py
```diff
@@ -0,0 +1,1158 @@
...
+    UserCuratedSubscriptionsRepository,
+)
+from src.repositories.user_store_installs import UserStoreInstallsRepository
+from src.store_categories import STORE_CATEGORIES
+from src.store_naming import suffixed_name
+
+logger = logging.getLogger(__name__)
+router = APIRouter(prefix="/api/marketplace", tags=["marketplace"])
+
+OWNER_TODO_PLACEHOLDER = "owner_todo"
```
cvrysanek, line 55: 🟡 `OWNER_TODO_PLACEHOLDER = "owner_todo"` is rendered to end users in production UI for any plugin without curator metadata (used in `_curated_to_item` and `_curated_inner_parent_fields`). Either send `None` and let templates render `—`, or use a user-readable string like `"Curator pending"`. The literal `owner_todo` looks like an unfilled TODO escaped from dev.

src/repositories/marketplace_plugins.py
```diff
@@ -83,11 +205,28 @@ def replace_for_marketplace(
...
+                # Upsert: ON CONFLICT keeps the existing created_at and
+                # refreshes only the mutable fields. New rows get
+                # CURRENT_TIMESTAMP via the column's DEFAULT.
                 self.conn.execute(
                     """INSERT INTO marketplace_plugins
                         (marketplace_id, name, description, version, author_name,
                          homepage, category, source_type, source_spec, raw, updated_at)
-                    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""",
+                    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+                    ON CONFLICT (marketplace_id, name) DO UPDATE SET
```
cvrysanek, line 220: 🟠 `ON CONFLICT (marketplace_id, name)` requires a `UNIQUE` (or `PRIMARY KEY`) constraint on `(marketplace_id, name)` for DuckDB to dispatch the upsert. The diff doesn't add it; behavior depends on the existing v ≤ 27 schema declaring it.

Please verify `_SYSTEM_SCHEMA` for `marketplace_plugins` declares `UNIQUE(marketplace_id, name)` and add a regression test that does `replace_for_marketplace` twice over the same input and asserts `created_at` is preserved on second call — that's the contract this upsert is buying. If the constraint isn't there, the first sync against a real v28 DB fails.

cvrysanek (commented):
Reviewed end-to-end. Strong feature; CHANGELOG and tests on the new API surface are thorough. Two items that I'd treat as blocking before merge:

1. **Path traversal in `_read_inner`** (inline) — `skill_name`/`agent_name` from URL flow into `Path` join without `resolve()` + `is_relative_to(plugin_root)` guard. RBAC limits blast radius (caller needs grant on parent plugin) but allows reading files inside *other* plugins/marketplaces they don't have a grant for. Fix is one line.
2. **No dedicated test for the v28 migration's destructive `DELETE FROM user_plugin_optouts`** — `tests/test_db_schema_version.py` only updates the constant. Need a `tests/test_v28_migration.py` (mirror of `test_keboola_v27_migration.py`) covering: pre-existing optouts wiped on upgrade from real v27 DB, `created_at` backfilled from `marketplace_registry.registered_at`, idempotency on second `_ensure_schema()` call, and fresh-install path doesn't trip `ALTER TABLE ADD COLUMN IF NOT EXISTS` against an existing column. This migration deletes user data — test coverage gap is the highest correctness risk in the PR.

Other high-priority items as inline comments. Smaller nits below.

## Other items not pinned to a line

- **`mergeable: CONFLICTING`** — needs rebase before merge.
- **`flea_detail`/`flea_skill_detail`/`flea_agent_detail` only `Depends(get_current_user)`** — no resource-level gate. Confirm this matches existing flea/Store policy ("anyone in the company can see"). If yes, a one-line comment in the route docstring would prevent future readers from concluding it was missed.
- **`category_counts` and `list_with_filters` use `JOIN resource_grants rg ON 1=1` then equality predicates in WHERE** — same plan, but reads as a cartesian. Move equalities into `ON`.
- **`subscribe()` / `unsubscribe()` do redundant SELECT-then-INSERT/DELETE** — `INSERT … ON CONFLICT DO NOTHING` already dedups; check `cursor.rowcount` for the "new row" return value.
- **`tab=my` Python-side filter on `granted` with no enforced cap** — fine at current scale; worth a TODO and/or a hard ceiling.
- **CI**: `smoke-test` and `e2e-bind-mount` jobs SKIPPED — confirm intentional for branch builds, not silently broken.
- **`cli/commands/my_stack.py`** accepts both `--on` and `--off`; consider raising `typer.BadParameter` if both passed.

src/repositories/marketplace_plugins.py
```diff
@@ -58,23 +58,145 @@ def count_by_marketplace(self) -> Dict[str, int]:
...
+            "rg.resource_id = mp.marketplace_id || '/' || mp.name",
+        ]
+        params: List[Any] = list(gids)
+        if search:
+            where_clauses.append(
+                "(LOWER(mp.name) LIKE ? OR LOWER(COALESCE(mp.description,'')) LIKE ? "
+                "OR LOWER(COALESCE(mp.author_name,'')) LIKE ? "
+                "OR LOWER(COALESCE(mp.category,'')) LIKE ?)"
+            )
+            needle = f"%{search.lower()}%"
```
cvrysanek, line 93: 🟢 `needle = f"%{search.lower()}%"` doesn't escape SQL `LIKE` wildcards. A user typing `100%` or `foo_bar` in the search box activates wildcard semantics. Not a security issue (params are bound), just unexpected results / minor DoS surface on `\` corner cases. Either escape `%`, `_`, `\` or use `LIKE ? ESCAPE '\\'`.

src/db.py
```diff
@@ -1931,6 +1933,41 @@ def _backfill_users_role_to_grants(conn: duckdb.DuckDBPyConnection) -> None:
...
+# avoid DDL churn on running operator instances. Row PRESENCE flips meaning
+# from "excluded" to "subscribed", so we wipe rows so the inverted reading
+# starts from a clean baseline. Users will re-install via /marketplace.
+#
+# Also adds marketplace_plugins.created_at (per-plugin "newest first" sort
+# on /marketplace). Backfilled from parent marketplace_registry.registered_at
+# so existing plugins get a sensible date until the next sync overwrites
+# with CURRENT_TIMESTAMP.
+_V27_TO_V28_MIGRATIONS = [
+    "DELETE FROM user_plugin_optouts",
```
cvrysanek, line 1953: 🟠 This DELETE wipes user data on upgrade from v27. The decision is well-documented in the comment block above and consistent with the semantic flip — but there is no test that exercises this path against a real v27 DB.

`tests/test_db_schema_version.py` only bumps the version constant. Please add `tests/test_v28_migration.py` (mirror of `test_keboola_v27_migration.py`) covering:

- Pre-existing rows in `user_plugin_optouts` are wiped during upgrade from a v27 DB.
- `marketplace_plugins.created_at` is backfilled from `marketplace_registry.registered_at` for existing plugins, leaves recently-synced plugins alone (NULL → backfilled, non-NULL → unchanged).
- Second `_ensure_schema()` call on a v28 DB does NOT re-DELETE (the `current < 28` guard) — regression test against accidental data loss on every boot.
- Fresh install (`_SYSTEM_SCHEMA` declares `created_at` directly) doesn't trip `ALTER TABLE ADD COLUMN IF NOT EXISTS`.

tests/test_db_schema_version.py
```diff
@@ -13,19 +13,20 @@
...
+    # initial_load_chunk_days). Existing sync_strategy column reused;
+    # admins can opt specific tables back to query_mode='local' for the
+    # new dispatcher (incremental, partitioned, full_refresh + where_filters).
+    # v27 → v28 (this PR): explicit-install (Model B) for curated
+    # marketplace plugins. user_plugin_optouts row presence flips meaning
+    # from "excluded" to "subscribed"; the migration wipes existing rows
+    # so the inverted reading starts from a clean baseline. Also adds
+    # marketplace_plugins.created_at (per-plugin "newest first" sort on
+    # /marketplace), backfilled from parent marketplace_registry.registered_at.
+    assert SCHEMA_VERSION == 28
```
cvrysanek, line 29: 🟠 This test only asserts the version constant. The actual destructive behavior of `_V27_TO_V28_MIGRATIONS` (DELETE on `user_plugin_optouts`, backfill on `marketplace_plugins.created_at`) is not tested anywhere. See the comment on `src/db.py` for the full coverage list a `tests/test_v28_migration.py` should hit.

src/repositories/user_curated_subscriptions.py
```diff
@@ -0,0 +1,134 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, List, Set, Tuple
+
+import duckdb
+
+
+class UserCuratedSubscriptionsRepository:
```
cvrysanek, line 20: 🟠 The repository name is `UserCuratedSubscriptionsRepository` but every SQL string still talks about `user_plugin_optouts` and `opted_out_at`. Decision is documented and reasonable (avoid DDL churn on running operators) — but the cognitive load on future readers (and SQL audits / grep scans for `optouts`) is real.

Minimal mitigation: add a DuckDB view in `_SYSTEM_SCHEMA` so the repo SQL can use the right names:

```sql
CREATE VIEW user_curated_subscriptions AS
  SELECT user_id, marketplace_id, plugin_name,
         opted_out_at AS subscribed_at
  FROM user_plugin_optouts;
```

Then this repository's queries read `FROM user_curated_subscriptions ... ORDER BY subscribed_at DESC` and the inverted semantic stops leaking out to the SQL surface. Schema-level rename can come in a later migration.

PR #231, opened 2026-05-08
feat(observability): optional PostHog integration

cvrysanek:
## Summary
Adds an optional PostHog integration covering backend exceptions, LLM call tracing, frontend errors / pageviews, masked session replay, and feature flags. Off by default — activates only when `POSTHOG_API_KEY` is set in env. Defaults to PostHog Cloud EU (`https://eu.i.posthog.com`); override host for US Cloud or self-hosted.

## Coverage
- **Backend exceptions** — FastAPI 500 handler + `src/orchestrator.py` rebuilds + `services/scheduler/` HTTP-job failures + `cli/main.py` uncaught CLI errors. CLI wrapper skips `Typer.Exit` / `SystemExit` / `KeyboardInterrupt`, flushes before re-raise so short-lived invocations don't drop events.
- **LLM tracing** — Anthropic + OpenAI-compat providers emit `$ai_generation` with provider, model, latency, token counts. Prompt / completion bodies stay off unless `POSTHOG_LLM_PAYLOADS=1` because LLM prompts here routinely include customer SQL / data.
- **Frontend** — browser snippet injected by `PosthogInjectionMiddleware` into every `text/html` response, registered inside the GZip layer so it sees uncompressed HTML. Many templates are standalone (their own `<!DOCTYPE>`) and never extend `base.html`, so a per-template include would miss them. Snippet covers `$pageview`, `$pageleave`, `window.error`, `unhandledrejection`, masked session replay (`maskAllInputs: true` plus CSS-selector mask for `[data-sensitive]`, `.data-cell`, `.query-result`, `.sql-output`, `code`, `pre`), and feature flags.

## Configuration

| Env var | Default | Notes |
|---|---|---|
| `POSTHOG_API_KEY` | unset | Project key only (publishable; embedded in browser HTML by design). Unset = integration off. |
| `POSTHOG_HOST` | `https://eu.i.posthog.com` | Override for US Cloud or self-hosted. |
| `POSTHOG_IDENTIFY_PII` | `email` | `none` / `id` / `email` / `full`. Default ships `user.id` + email, never name. |
| `POSTHOG_REPLAY` | `true` | Disable replay only. |
| `POSTHOG_REPLAY_MASK_SELECTOR` | empty | Append CSS selector to default mask list. |
| `POSTHOG_LLM_PAYLOADS` | `0` | `1` adds `\$ai_input` / `\$ai_output_choices`. Off by default — privacy. |

## CLI entry point change
`pyproject.toml` moves `da` from `cli.main:app` to `cli.main:main`. The new `main()` wraps the Typer app to capture uncaught exceptions, flush, then re-raise. Operators upgrading need to reinstall the wheel for the CLI binary to pick up the new entry point.

## Tests
18 new tests (all passing):
- `tests/test_posthog_disabled.py` — proves zero side effects when key unset (no SDK construction, no network, snippet renders empty, LLM tracing yields a noop).
- `tests/test_posthog_client.py` — capture / capture_exception / identify-mode payloads / SDK fallback when `capture_exception` missing on older SDKs.
- `tests/test_llm_tracing.py` — `\$ai_generation` shape on success, error variant, payload-flag toggle, Anthropic / OpenAI extractor helpers.

End-to-end smoke validated against EU Cloud ingest: HTTP 200 on `eu.i.posthog.com/i/v0/e/` with the project key.

## Vendor-agnostic
Default host points at PostHog's public EU endpoint (a public service, not a customer-specific deployment). PostHog itself is OSS / self-hostable — operators with a self-hosted instance just override `POSTHOG_HOST`. No deployment-specific project IDs, hostnames, or accounts in code or config defaults. CHANGELOG entry under `[Unreleased] Added`.

## Test plan
- [x] `pytest tests/test_posthog_*.py tests/test_llm_tracing.py` — 18 passed
- [x] Boot with `POSTHOG_API_KEY` unset → no `posthog` references in HTML, no SDK init logs, no network
- [x] Boot with key set → "PostHog observability enabled (host=…, identify=…, replay=…)" in startup logs
- [x] HTML rewrite injects snippet into `/dashboard`, `/admin/access`, `/catalog`, etc.
- [x] Direct POST to `eu.i.posthog.com/i/v0/e/` with project key → HTTP 200
- [ ] Reviewer: validate `$ai_generation` arrives in PostHog LLM Observability after exercising an LLM-using route
- [ ] Reviewer: validate browser session replay masks `<input>` values
- [ ] Reviewer: validate feature flag round-trip (define a flag in PostHog UI, verify `is_feature_enabled` reflects it)

See `docs/observability.md` for the operator guide.


app/middleware/posthog_inject.py
```diff
@@ -0,0 +1,110 @@
...
+        body = body.replace(_HEAD_CLOSE, snippet.encode("utf-8") + _HEAD_CLOSE, 1)
+        # content-length must reflect the rewritten body — Starlette's
+        # ``Response`` sets it for us when we drop the prior header.
+        new_headers = {k: v for k, v in response.headers.items() if k.lower() != "content-length"}
+        return Response(
+            content=body,
+            status_code=response.status_code,
+            headers=new_headers,
+            media_type=response.media_type,
+        )
```
minasarustamyan, line 138: **Critical issue: `Response.background` is dropped on every HTML response**

`BaseHTTPMiddleware` materialises the response body and asks us to return a fresh `Response(...)`. All three `Response(...)` callsites here drop `response.background`, so any `BackgroundTask` / `BackgroundTasks` attached to an HTML route silently never runs once `POSTHOG_API_KEY` is set.

This is a silent data-loss regression — no log, no error, just missing side effects. Anything attached as a background task on an HTML route (audit logging, async webhooks, deferred email sends, etc.) stops firing the moment the integration is enabled.

Forwarding `response.background` on each return path keeps the existing contract intact:

```suggestion
        if _HEAD_CLOSE not in body or b"posthog.init" in body:
            return Response(
                content=body,
                status_code=response.status_code,
                headers=dict(response.headers),
                media_type=response.media_type,
                background=response.background,
            )

        try:
            snippet = _render_snippet(request)
        except Exception:
            logger.exception("PostHog snippet render failed; serving response unmodified")
            return Response(
                content=body,
                status_code=response.status_code,
                headers=dict(response.headers),
                media_type=response.media_type,
                background=response.background,
            )

        body = body.replace(_HEAD_CLOSE, snippet.encode("utf-8") + _HEAD_CLOSE, 1)
        # content-length must reflect the rewritten body — Starlette's
        # ``Response`` sets it for us when we drop the prior header.
        new_headers = {k: v for k, v in response.headers.items() if k.lower() != "content-length"}
        return Response(
            content=body,
            status_code=response.status_code,
            headers=new_headers,
            media_type=response.media_type,
            background=response.background,
        )
```

A `BaseHTTPMiddleware` regression test using `TestClient` + a route that returns `HTMLResponse(..., background=BackgroundTask(...))` would catch this — and the same fixture is the right place to also assert the docstring's `StreamingResponse` skip (currently the dispatch path only checks `content-type`, so a `StreamingResponse(media_type="text/html")` would still get materialised).

src/observability/llm_tracing.py
```diff
@@ -0,0 +1,170 @@
...
+        }
+        if cap.input_tokens is not None:
+            props["$ai_input_tokens"] = cap.input_tokens
+        if cap.output_tokens is not None:
+            props["$ai_output_tokens"] = cap.output_tokens
+        if pc.llm_payloads_enabled:
+            if cap.prompt is not None:
+                props["$ai_input"] = cap.prompt
+            if cap.output is not None:
+                props["$ai_output_choices"] = cap.output
```
minasarustamyan: **Bug: oversized payloads are dropped silently**

PostHog's ingestion has a per-event size limit (~32 KB by default; the SDK does not chunk and surfaces 413s only as best-effort log lines — oversized captures effectively land on the floor).

In Agnes, LLM prompts almost always carry sample rows, table schemas, and analyst SQL. They routinely exceed 32 KB, so flipping `POSTHOG_LLM_PAYLOADS=1` *for the very calls an operator would want to inspect* is exactly when the event vanishes — the dashboard ends up biased toward small, uninteresting calls. There's no log line either way.

A defensive truncation at this site keeps the metadata (provider, model, tokens, latency, error) flowing while bounding payload size:

```suggestion
        if pc.llm_payloads_enabled:
            # PostHog drops events past its per-event size limit (~32 KB
            # default). Agnes prompts routinely include sample rows /
            # schemas / SQL that exceed this — truncate so the event
            # itself still lands instead of disappearing silently.
            if cap.prompt is not None:
                props["$ai_input"] = str(cap.prompt)[:30_000]
            if cap.output is not None:
                props["$ai_output_choices"] = str(cap.output)[:30_000]
```

A configurable cap (env var) would be nicer long-term, but a hardcoded ceiling is the minimum needed to keep the flag from quietly degrading visibility.

app/middleware/posthog_inject.py
cvrysanek: Fixed in 499831e. Routed every `Response(...)` return through a `_passthrough()` helper that forwards `background=response.background` (and `media_type` / `headers`). Also added a 4 MB `_MAX_BUFFER_BYTES` cap so streamed-HTML responses short-circuit through with a warning instead of buffering arbitrarily.

Regression tests in `tests/test_posthog_inject_middleware.py` cover four return paths: snippet rendered, snippet-render exception, double-injection guard, non-HTML passthrough. Each asserts the route's `BackgroundTask` actually fired. Streaming-cap test patches `_MAX_BUFFER_BYTES` and confirms a body past the cap serves without injection.

src/observability/llm_tracing.py
cvrysanek: Fixed in 499831e. `$ai_input` / `$ai_output_choices` are now clipped at `POSTHOG_LLM_PAYLOAD_MAX_CHARS` (default 30000) with an explicit `…[truncated N chars]` marker so a reader doesn't mistake a truncated capture for a complete one. Default leaves ~2 KB headroom under the 32 KB envelope. Metadata (provider, model, tokens, latency, error) flows regardless of payload size.

Three new tests in `tests/test_llm_tracing.py` cover default-cap clipping, env-override (set to 100, asserts marker + first 100 chars), and pass-through when the payload is under cap.

minasarustamyan (approved):

PR #232, opened 2026-05-08
Extract session-pipeline framework + UsageProcessor skeleton

minasarustamyan:
## Summary

- Pluggable framework under `services/session_pipeline/` (contract + lib + per-processor runner) so multiple processors can read `/data/user_sessions/<key>/*.jsonl` independently — own scan loop, own scheduler tick, own state row keyed by `(processor_name, session_file)`. Slow / failing processor cannot block any other.
- Verification flow becomes the first plugin (`services/session_processors/verification.py`); a no-op `UsageProcessor` reserves the second slot pending a separate brainstorm on extraction logic + storage shape (skill / agent invocation events).
- **BREAKING (operator-facing only):** schema bump v28 → v29 renames `session_extraction_state` → `session_processor_state` with composite PK; `/api/admin/run-verification-detector` replaced by parametrized `/api/admin/run-session-processor?processor=<name>`. Migration is conditional + idempotent (no-ops the copy step on fresh installs that came up at the new schema).

## Why

`services/verification_detector` was the only consumer of the centralized session tree but its scan-state-parse loop is generic. Future processors (usage tracking, error metrics, security scans) shouldn't each rebuild it. This refactor extracts the scaffold so adding a new processor is one file in `services/session_processors/<name>.py` + one entry in the registry list + one line in scheduler `JOBS`.

## Architecture decisions

- **Independent processors, not parse-once-dispatch.** Each processor parses jsonl itself. JSON parsing is cheap relative to LLM calls / DB inserts; isolation is worth the duplicated parse cost.
- **State table:** drop the legacy `session_extraction_state` (sole caller is verification_detector, which we're rewriting). New `session_processor_state` PK is `(processor_name, session_file)` so adding a processor today retroactively reprocesses all historical sessions for that processor only. `is_processed` now compares `file_hash` so a session jsonl that grows (live append from an active Claude Code session) gets reprocessed — previously the file_hash was stored but never read back.
- **Scheduler JOBS schema unchanged** — encode processor in URL query string, keep existing `(name, schedule, endpoint, method, timeout)` tuple.
- **Cadence:** env-overridable per processor. `verification`=15min (matches today, `SCHEDULER_VERIFICATION_DETECTOR_INTERVAL` env var retained for operator compatibility); `usage`=10min (`SCHEDULER_USAGE_PROCESSOR_INTERVAL`, new).
- **No new tables for usage data** — `UsageProcessor` is intentionally a no-op skeleton. Storage shape (DuckDB table vs append-only parquet event log) decided in follow-up.

## Coordination with PR #226 (UI overhaul)

Both touch `app/api/admin.py`, `app/web/router.py`, `tests/test_web_ui.py`, `CHANGELOG.md` — but at different line ranges. No semantic conflict; PR #226 adds `_audit()` helper + 28 new audit-instrumented endpoints, this one adds one new endpoint that already follows the existing inline `AuditRepository(conn).log(...)` pattern. Once both merge, an optional follow-up could switch this PR's audit call to use #226's helper for consistency.

## Test plan

- [ ] `pytest tests/test_session_pipeline.py` — 21 new tests pass (lib utilities, repo CRUD, runner success/raise/empty/file-hash paths, v29 migration round-trip)
- [ ] `pytest tests/test_corporate_memory_v1.py::TestVerificationDetectorIntegration tests/test_corporate_memory_v1.py::TestDetectorPersistsEvidence tests/test_corporate_memory_v1.py::TestDetectorIgnoresLLMConfidence` — verification regression (same outputs as pre-refactor)
- [ ] `pytest tests/test_admin_run_endpoints.py tests/test_health_session_pipeline.py tests/test_scheduler.py` — endpoint + audit + scheduler wiring
- [ ] `pytest tests/test_db_schema_version.py tests/test_instance_config_overlay.py tests/test_web_ui.py` — schema + overlay + UI smoke
- [ ] Manual on dev VM: `curl -X POST -H "Authorization: Bearer \$ADMIN_PAT" "http://localhost:8000/api/admin/run-session-processor?processor=verification"` returns `{"ok": true, "details": {...}}` with audit row `action="run_session_processor:verification"`
- [ ] Manual: `?processor=usage` returns clean stats with `items_extracted=0` (no-op)
- [ ] Manual: `?processor=bogus` returns 400 with actionable error

## Out of scope (follow-ups)

- `UsageProcessor` extraction logic + storage shape — separate brainstorm + PR.
- Retry counter / dead-letter for permanently failing sessions — current behavior (retry-forever, mirrors pre-refactor) preserved.
- CLI / web telemetry sources — only session-driven processors in this PR.


src/repositories/session_processor_state.py
```diff
@@ -0,0 +1,113 @@
...
+                # the hash-check authoritative in one place (the runner /
+                # is_processed call).
+                if key not in known:
+                    results.append((username, jsonl_file))
+                else:
+                    # Row exists; runner will recompute hash and re-check
+                    # via is_processed(). Pass it through so it gets that
+                    # chance — cost is one hash compute per stable session,
+                    # which is in line with today's behavior anyway.
+                    results.append((username, jsonl_file))
```
cvrysanek, line 112: **P1 — `scan_unprocessed_for` dead branch.** Both arms of the `if key not in known` block call `results.append((username, jsonl_file))`. The `known` dict (and the `SELECT session_file, file_hash` query at L99-105) is computed but never used to short-circuit. The method's docstring promises "return jsonl files this processor has not yet processed", but it returns *every* `*.jsonl` it finds, then defers all filtering to the runner's `is_processed` re-hash. Result: every tick the runner MD5-rehashes every stable session for every processor.

Fix shape:
```python
if key not in known:
    results.append((username, jsonl_file))
# else: row exists; skip and let the runner's periodic full-scan path handle hash invalidation OR
# compare (size, mtime) here as a cheap precheck before we commit the runner to a hash.
```

Flagged by 4 reviewers (maintainability/high, reliability/conf 1.00, kieran-python/P1, performance/med). The dead-code interpretation is hard to dispute given both branches are identical.

cvrysanek (commented):
## Code review summary

13 reviewers (correctness, testing, maintainability, project-standards, agent-native, learnings, security, performance, api-contract, data-migrations, reliability, adversarial, kieran-python). Security clean. 6 P1s flagged inline; full set (P1/P2/P3 + deployment notes) in run artifact.

**P1 fix order suggestion:**
1. `scan_unprocessed_for` dead branch (1-line fix; unblocks the per-tick MD5 storm flagged separately by performance + reliability).
2. Per-processor concurrency lock around `_run_processor` (scheduler tick + admin POST race → duplicate `verification_evidence` rows).
3. Surface `errors > 0 and processed == 0` as non-200 so the scheduler stops marking silent failures green.
4. Advance `last_run` on terminal state, not only success — otherwise a permanently failing job retries every 30s tick instead of on cadence.
5. Scheduler tick is currently serial: a 900s verification timeout blocks every other job (data-refresh, marketplaces, health). Async/parallel ticks or a per-job worker pool. Larger refactor — could land as follow-up after 1–4 if the silent-failure surface is closed.
6. Operator-facing 410 Gone shim or one-release deprecation window before deleting `/run-verification-detector` (audit-action rename `run_verification_detector` → `run_session_processor:verification` is the same risk for downstream consumers — Grafana panels, alert filters).

**P2 / P3 not blocking but worth a sweep:** `@lru_cache` on `_build_registry` permanently caches a failed verification build (operator workflow `/api/admin/configure` no longer self-heals); `UsageProcessor` is a registered no-op that will mark sessions processed under `processor_name='usage'` (state-cleanup burden when real logic lands); CLAUDE.md schema narrative not extended to v29; v28→v29 migration not wrapped in BEGIN/COMMIT (idempotency rests on `INSERT OR IGNORE` + conditional); `docs/architecture.md:234-235,336-338` still references the old table + endpoint.

**Decision-3 watchlist (corporate-memory ADR):** `verification.py`'s new `except ValueError → fallback 'confirmation'` on unknown `detection_type` re-introduces a small LLM-influenced trust surface. Original code raised. Worth confirming intent.

No prior review threads. Run artifact: `.context/compound-engineering/ce-review/20260508-195653-926258a4/`.

services/scheduler/__main__.py
```diff
@@ -161,7 +168,14 @@ def build_jobs() -> list[tuple[str, str, str, str, int]]:
         # single source of truth for the health-check staleness grace
         # window in app/api/health.py (which uses 2x the cadence).
         ("session-collector",     _seconds_to_schedule(sess),    "/api/admin/run-session-collector",     "POST", 300),
-        ("verification-detector", _seconds_to_schedule(verify),  "/api/admin/run-verification-detector", "POST", 900),
+        # session-pipeline processors — independent loops, each invoked on
+        # its own cadence via the parametrized run-session-processor endpoint.
+        # Adding a third processor in the future is one line here + one entry
+        # in services/session_processors/__init__.py registry.
+        ("session-processor:verification", _seconds_to_schedule(verify),
+            "/api/admin/run-session-processor?processor=verification", "POST", 900),
```
cvrysanek, line 178: **P1 — Adding more jobs amplifies two pre-existing scheduler scaffold issues** (both outside this hunk but triggered by it):

1. **Serial tick loop (`__main__.py:239-243`).** `for ... jobs:` then synchronous `_call_api(endpoint, method, timeout_sec=900)`. While verification is in flight, every other job (`data-refresh`, `marketplaces-sync`, `health`, the new `usage` processor) waits the same tick. The PR's stated isolation guarantee — "slow / failing processor cannot block any other" — holds inside `services/session_pipeline/runner.py` but is broken at the scheduler dispatch layer. At 900s × even a few stuck sessions, the tick wraps past every other job's cadence.

2. **`last_run` advances only on success (`__main__.py:244-245`).** A permanently-failing job is therefore retried on *every* tick (default 30s), not on its 15min cadence. Combined with the silent-200-on-total-failure issue I flagged in `app/api/admin.py:3409`, HTTP returns success and `last_run` does advance — but if the job ever emits a real 5xx (network blip, DuckDB lock), it'll loop hot and starve the rest of the schedule.

Options: `asyncio.gather`/`httpx.AsyncClient`, a small thread pool keyed by job, or a per-job process. The job tuple already carries `timeout_sec` — easy to keep per-job timeouts intact in any of those shapes. Advance `last_run` on terminal state (success OR final-failure) and surface stuck jobs via `/api/admin/scheduler-runs`.

app/api/admin.py
```diff
@@ -3336,44 +3336,45 @@ def run_session_collector(
...
+                f"Known: {', '.join(list_processor_names())}"
+            ),
+        )
 
     job_conn = get_system_db()
     stats: dict = {}
     job_error: Optional[Exception] = None
     try:
-        stats = detector.run(job_conn, extractor, dry_run=False)
+        stats = _run_processor(job_conn, proc)
```
cvrysanek, line 3409: **P1 — No per-processor concurrency lock around `_run_processor`.** Two trigger paths now exist for the same `processor`: scheduler tick + manual admin POST. A scheduler tick mid-flight + a manual `?processor=verification` arrives → both call `_run_processor(job_conn, proc)` against the same `/data/user_sessions/*` set, both call `_generate_id` on overlapping detections, both call `repo.create_evidence(...)` per detection. Even when the duplicate-id branch in `verification.py` short-circuits the `create_node` path, **`create_evidence` still runs on both invocations** (that's the fix from the corporate-memory v1 ADR Decision 3 — "persist evidence even on duplicates so the additional-verifiers boost can accumulate"). With concurrent runs that boost is now tracking *the same analyst's same session* twice. LLM cost also doubles.

Fix shape: take a per-processor advisory lock (`SELECT ... FOR UPDATE` on a row in `session_processor_state`, or a named DB lock, or a process-local `asyncio.Lock` keyed by processor name) before `_run_processor`. The `_overlay_write_lock` pattern elsewhere in the repo is a good template — single-writer DuckDB makes this cheap.

app/api/admin.py
```diff
@@ -3385,23 +3386,27 @@ def run_verification_detector(
...
+        action=f"run_session_processor:{processor}",
+        resource=f"job:session-processor:{processor}",
         params=audit_params,
     )
 
     if job_error is not None:
         raise HTTPException(status_code=500, detail=audit_params["unhandled_error"])
 
-    return {"ok": not stats.get("errors"), "details": stats}
+    return {"ok": stats.get("errors", 0) == 0, "details": stats}
```
cvrysanek, line 3446: **P1 — Silent total failure returns HTTP 200.** This `return {"ok": stats.get("errors", 0) == 0, "details": stats}` keeps the HTTP status at 200 even when *every* session raised an `LLMError` (`stats.errors=N, processed=0`). The scheduler's `_call_api` (services/scheduler/__main__.py:203) checks only `resp.status_code < 400`, so it logs success, advances `last_run`, and the audit row is written with green status. `/api/admin/scheduler-runs` shows green. Operators only see the failure via `/api/health/detailed` after ~30 min, or by reading `audit_log.params.errors`.

Fix: return HTTP 207 Multi-Status (or 502/500) when `errors > 0 and processed == 0`, or split into a separate non-200 codepath for total failure. The audit row stays useful regardless; the scheduler signal needs to flip.

app/api/admin.py
```diff
@@ -3336,44 +3336,45 @@ def run_session_collector(
     return {"ok": rc == 0, "details": {"rc": rc, **stats}}
 
 
-@router.post("/run-verification-detector")
-def run_verification_detector(
+@router.post("/run-session-processor")
```
cvrysanek, line 3363: **P1 — Old endpoint removed without deprecation shim.** The PR body acknowledges `BREAKING (operator-facing only)` for `/api/admin/run-verification-detector` → `/api/admin/run-session-processor?processor=verification`, but the old route is deleted outright (no traces in `app/api/admin.py` other than stale comments at L3315 / L3436). Two practical issues:

1. **Rollout window.** Compose flips the app container before / after the scheduler container. A scheduler running the previous release POSTs `/run-verification-detector` and gets a silent 404 for the rollout window. Audit log has nothing (the old action string isn't being written by anyone), and the new `_call_api` only logs `HTTP 404` at WARNING and returns False — `last_run` doesn't advance, no per-instance alert. Today there's no monitor on "scheduler 404 rate spike".
2. **Audit-log consumers.** `audit_log.action` rename `run_verification_detector` → `run_session_processor:verification` will silently break any Grafana panel / alert filter / BI query keyed on the old action string. CHANGELOG flags BREAKING but doesn't enumerate this side-effect.

Fix shape: keep `/run-verification-detector` for one release as a thin redirect (`return run_session_processor(processor="verification", ...)`) emitting `Deprecation: true` + `Sunset:` headers, OR return HTTP 410 Gone with a body pointing to the new route. Same idea for the audit action — emit both strings during the deprecation window so downstream consumers can pivot.

src/repositories/session_processor_state.py
minasarustamyan: Fixed in c3250c4. Replaced the dead `if/else` with an mtime precheck:

- Files with `mtime <= processed_at` are filtered at scan — the runner never reads or hashes them.
- Files modified since `processed_at` still surface for the runner's authoritative `file_hash` invalidation (preserves the live-append reprocess feature for active Claude Code sessions).

Naive-local comparison matches the existing `app/api/health.py:_check_session_pipeline` idiom — DuckDB TIMESTAMP strips tz on storage and converts tz-aware writes to local time before storing, so we compare mtime + processed_at both as naive-local.

Tests in `tests/test_session_pipeline.py`:
- `test_scan_filters_stable_sessions_via_mtime` — stable session not surfaced; new file alongside it is.
- `test_scan_surfaces_session_modified_after_processing` — post-mark `os.utime` bump → file returned.

services/scheduler/__main__.py
minasarustamyan: Fixed in 0f9b1a5. Both bugs you flagged addressed in `services/scheduler/__main__.py`:

1. **Serial tick → `ThreadPoolExecutor.submit` per due job.** A 10-minute verification run no longer blocks the 60s health-check or the 30s data-refresh from firing on their own cadences. `in_flight` set + `threading.Lock` prevents the same job being re-launched on subsequent ticks during a long-running invocation — without this, a 10-min run during which 20 ticks fire would queue 20 duplicate POSTs (the per-processor lock from c3250c4 would 409 most, but they'd still be wasted requests + audit-log noise).

2. **`last_run` advanced on terminal state, not just success.** Permanent-failure jobs now retry on cadence (15min) instead of every tick (30s). Errors still surface via `_call_api` logging + `audit_log` on the receiving side — change is only in retry rhythm.

`_run_job` extracted to a module-level helper so the bookkeeping is unit-testable.

New tests in `tests/test_scheduler.py`:
- `TestRunJobBookkeeping::{test_advances_last_run_on_success, …on_failure, …when_call_raises}` — `last_run` advances + `in_flight` clears in the `finally` even when the call raises.
- `TestRunLoopParallelism::test_in_flight_skip_prevents_duplicate_launches` — single slow job (0.3s sleep) under a 0s tick budget for 0.4s → ≤2 launches (without protection: 5+).

Per-job timeout preserved through the existing `timeout_sec` tuple element. Single-process worker pool — multi-pod deployments would need cross-pod coordination, but Agnes is single-instance per customer today (matching the `_overlay_write_lock` caveat).

app/api/admin.py
minasarustamyan: Fixed in c3250c4. Added `_get_processor_run_lock(name)` factory returning a per-processor `threading.Lock`. The endpoint acquires non-blocking → returns **409 Conflict** on contention so the operator sees what happened instead of stacking behind a long-running tick. Lock releases unconditionally in `finally:` so a runner exception can't wedge the processor permanently.

Pattern mirrors `_overlay_write_lock` in the same module — single-process FastAPI workers; multi-worker deployments would need an OS-level / DB-backed lock (documented limitation, same as `_overlay_write_lock`'s caveat).

Combined with the parallel-tick fix (3209956475), the scheduler now also won't queue a second `verification` POST while one is in flight (the `in_flight` set short-circuits at the dispatcher), so the 409 path is mostly a defensive backstop for manual admin invocations during a tick.

Tests in `tests/test_admin_run_endpoints.py`:
- `test_concurrent_invocation_returns_409` — lock held externally → next POST returns 409 with `"already running"` detail.
- `test_lock_released_on_runner_exception` — runner raises → 500 returned → lock available for the next caller (no permanent wedge).

app/api/admin.py
minasarustamyan: Acknowledged — but flagging this is **pre-existing behavior preserved verbatim** from the old `run_verification_detector` endpoint (`return {"ok": not stats.get("errors"), "details": stats}` — same shape, same silent-200-with-`ok: false` on total failure). The refactor inherited the regression rather than introducing it.

The two parallel-tick fixes from this PR (3209956475 + 3209956484) reduce the symptoms — `last_run` now advances even on failure, so we don't hot-loop a permanently failing job 30× per cadence — but the silent-success-status concern is independent and worth fixing on its own merits.

Proposing as a follow-up commit (this branch or separate, happy either way) since the right fix touches all three `/api/admin/run-*` endpoints (`session-collector` + `corporate-memory` use the same shape) and ought to be done in one pass for consistency. Shape would be:
- `errors == 0` → 200
- `errors > 0 and processed > 0` → 207 Multi-Status (partial)
- `errors > 0 and processed == 0` → 502 (total failure)

`/api/admin/scheduler-runs` would then surface the failure inline instead of operators having to dig into `audit_log.params.errors`. Will defer to author on whether to land here vs. a follow-up.

app/api/admin.py
minasarustamyan: Acknowledged. Two valid concerns at differing severities:

1. **Rollout window 404s** — agreed but limited blast radius: `app` and `scheduler` containers share an image and restart together on `docker compose up -d`, so the window is ~seconds, not minutes. `_call_api` logs WARNING + returns False rather than crashing, and the new `last_run` semantics (now advance on terminal state, see 3209956475) means a single 404 cycle at rollout doesn't queue retry storms.

2. **Audit-action rename silently breaks downstream consumers** — this is the harder one. Operators with Grafana panels / alert filters keyed on `action='run_verification_detector'` will silently miss rows after deploy. CHANGELOG marks BREAKING but doesn't enumerate this side-effect.

Discussing with original author. Two options on the table:
- **Hard cut now (current state):** PR labels BREAKING, CHANGELOG documents. Per project memory the codebase prefers clean cuts when a feature is unused outside internal callers.
- **One-release deprecation shim:** keep `/run-verification-detector` as a thin redirect (`return run_session_processor(processor='verification', ...)`) emitting `Deprecation: true` + `Sunset:` headers, AND dual-write the audit action (`run_verification_detector` plus `run_session_processor:verification`) for the same window. Adds ~20 lines + one regression test.

Will respond once decided. Defaulting to hard-cut unless the author signals operator-facing dashboards exist.

minasarustamyan:
## Review-comment status

Thanks @cvrysanek for the thorough pass. Per-comment status:

| # | Issue | Status | Commit |
|---|---|---|---|
| 1 | `scan_unprocessed_for` dead branch | ✅ Fixed | c3250c4 |
| 2 | Scheduler serial tick + `last_run` on success only | ✅ Fixed | 0f9b1a5 |
| 3 | No per-processor concurrency lock | ✅ Fixed | c3250c4 |
| 4 | Silent total-failure returns 200 | 🔵 Acknowledged (pre-existing, see reply) | — |
| 5 | Old endpoint deleted without deprecation shim | 🔵 Discussing trade-off (see reply) | — |

### Decision-3 watchlist note

The watchlist item ("`verification.py`'s new `except ValueError → fallback 'confirmation'` re-introduces a small LLM-influenced trust surface. Original code raised.") appears to be a false positive — the fallback is preserved verbatim from the pre-refactor `services/verification_detector/detector.py:212-215`:

```python
except ValueError:
    # Unknown detection_type from the LLM; fall back to a
    # lookup-keyed default rather than the LLM-supplied value.
    confidence_value = compute_confidence("user_verification", "confirmation")
```

I copy-pasted this block 1:1 into `VerificationProcessor.process_session`. Original code did not raise on unknown `detection_type` — it fell back to `"confirmation"` with the same comment. Happy to revisit the underlying ADR-3 question separately, but flagging that nothing changed in the refactor here.

### Test stats post-fixes

- 21 new pipeline tests (`tests/test_session_pipeline.py`)
- 4 new scheduler tests (`tests/test_scheduler.py`)
- 5 new admin endpoint tests (`tests/test_admin_run_endpoints.py`)
- 280 affected tests passing across the impacted suites

Issue #16, opened 2026-04-21
Multi-customer deployment wave (2026-04-21) — retrospective


ZdenekSrotyr:
Epic tracking for the multi-customer deployment initiative completed 2026-04-21. All work is merged to \`main\` through the merge commits listed below. Branch \`feature/multi-customer-deployment\` is archived after this issue.

## Outcome

Agnes now supports a per-customer self-deploy model on GCP:
- Public upstream: this repo (app + TF module + docs)
- Template repo: [keboola/agnes-infra-template](https://github.com/keboola/agnes-infra-template)
- Keboola instance: [keboola/agnes-infra-keboola](https://github.com/keboola/agnes-infra-keboola)
- Running: \`agnes-prod\` @ 34.77.102.61, \`agnes-dev\` @ 34.77.94.14 (kids-ai-data-analysis GCP project)

## Artefacts

- Spec: \`docs/superpowers/specs/2026-04-21-multi-customer-deployment-spec.md\`
- Plan: \`docs/superpowers/plans/2026-04-21-multi-customer-deployment.md\`
- Live log: \`docs/superpowers/plans/2026-04-21-deployment-log.md\` — chronology, known limitations, onboarding checklist
- Onboarding guide: \`docs/ONBOARDING.md\`

## Module release tags

| tag | highlights |
|---|---|
| \`infra-v1.0.0\` | initial customer-instance module (prod + dev_instances, PD, Secret Manager, VM SA, OS Login) |
| \`infra-v1.1.0\` | volume bind-mount to PD, cron-based auto-upgrade (replaces broken Watchtower) |
| \`infra-v1.2.0\` | CI smoke test fix (split host mount into separate overlay) |
| \`infra-v1.3.0\` | daily disk snapshots, uptime checks, alert policies |
| \`infra-v1.4.0\` | code-review fixes (scoped VM SA, fail-fast secrets, firewall split, cron reads .env, etc.) |
| \`infra-v1.5.0\` | version badge env injection |
| \`infra-v1.5.1\` | alert policy aggregation patch |
| \`infra-v1.5.2\` | image ENV wins over .env for CalVer |

## Merge commits on main

- \`e4f6910\` Merge: real CalVer + commit SHA in UI badge
- \`7553f77\` Merge: version badge partial on all full-page templates
- \`dbac3e6\` Merge: alert policy reducer fix
- \`717f40c\` Merge: bootstrap monitoring role fix
[...]

PR #234, opened 2026-05-08
Curated marketplace enrichment via agnes-metadata.json + curator metadata

minasarustamyan:
## Summary

- Adds a sibling metadata file `.claude-plugin/agnes-metadata.json` that upstream curated-marketplace repos can opt into to declare per-plugin (and per-skill / per-agent) cover photo, demo video, doc links, and category override. The Claude Code `marketplace.json` contract is untouched — agnes-metadata.json + the convention `.agnes/` directory are stripped from the synthetic Claude Code marketplace served to user instances, so a repo registered in Agnes still works as a vanilla Claude Code marketplace for direct `plugin marketplace add` consumers.
- Adds **mandatory curator** (full name + email) on registered marketplaces. Validation lives at the API layer; the DB columns are nullable so existing rows survive migration. Surfaces on `/marketplace` cards + plugin / skill / agent detail in place of the historic `owner_todo` placeholder.
- Adds an **external-asset mirror cache** at `${DATA_DIR}/marketplace-cache/<slug>/` with conditional GET, 60s timeout, 10 MB body cap, max 4 concurrent fetches, and SSRF guards. User-Agent is Wikipedia-policy-compliant. Inner-level (skill / agent) external URLs are mirrored too.
- **Strict drop semantics:** anything Agnes can't deliver as a real PDF / Markdown / plain text doc, or a real PNG / JPEG / WebP cover, is dropped from the served metadata entirely — the UI renders identically to the no-entry case (gradient placeholder for missing covers, no row in the doc list).
- New `/api/marketplace/curated/{mp}/{plugin}/{asset,doc,mirrored}/...` endpoints with path-traversal guards + RBAC + `Content-Disposition: attachment` for docs (force-download).
- Schema bump v31 → v32 with idempotent `ALTER TABLE … ADD COLUMN IF NOT EXISTS` migrations.
- Curator-focused format guide at `/marketplace/format-guide` (linked from `/admin/marketplaces` next to the **+ Add Marketplace** button); canonical source is `docs/curated-marketplace-format.md`.
- Same allowlist (PDF / Markdown / plain text for docs; PNG / JPEG / WebP for images, with magic-bytes check) is now enforced on the Flea upload path (`/store/new`) too.

## Test plan

- [x] Unit + integration tests added under `tests/test_marketplace_metadata.py`, `tests/test_marketplace_asset_mirror.py`, `tests/test_marketplace_synth_strip.py`, `tests/test_marketplace_v32_endpoints.py`. Existing `tests/test_marketplace.py`, `tests/test_marketplace_api.py`, `tests/test_db_schema_version.py` extended.
- [x] CI green (3695 tests pass).
- [x] Manual end-to-end walkthrough against a curated test fixture covering all 9 enrichment variants (full / partial / absent for plugins, skills, agents; internal vs external assets; mirror failure paths). See test fixture at `agnes-claude-marketplace#agnes-metadata-test-cases` branch.
- [x] Synth Claude Code ZIP / git tree verified to NOT include `.claude-plugin/agnes-metadata.json` or `.agnes/**` — covered by `test_marketplace_synth_strip.py` plus a manual `unzip -l /marketplace.zip | grep agnes-metadata`.
- [x] Format guide page renders + scoped to logged-in users (no admin requirement).
- [x] Existing curated marketplaces without `agnes-metadata.json` continue to render unchanged (covers fall to gradient placeholders, no Demo video / Documentation sections; `owner_todo` placeholder until admin fills curator on the registry row).


cvrysanek (commented):
## Review summary

Solid feature. Magic-byte validation, atomic writes, RBAC gating, idempotent v32 migration, Wikipedia-policy User-Agent — lots done right. **Two P0 SSRF gaps and one P1 stored-XSS gap should land before merge.** Plus a handful of correctness + maintainability cleanups.

Three reviewers (security / correctness / maintainability) ran in parallel against `FETCH_HEAD` (commit `6d3818c`).

---

### 🚨 Blockers

#### 1. SSRF via redirect → cloud metadata IAM credential exfiltration · **P0**
`src/marketplace_asset_mirror.py:152-192, 291-293`

`_is_safe_url()` validates only the **initial** URL. `urllib.request.urlopen` then follows up to 10 redirects with no SSRF re-check on the `Location` header.

**Attack:** a curator points `cover_photo` at `https://attacker.example/x.png`; attacker's origin returns `302 Location: http://169.254.169.254/latest/meta-data/iam/security-credentials/<role>`. urllib follows transparently. For `kind=doc`, `accept_doc_response` accepts `text/plain`, so IAM credentials persist to disk and are served via `/api/marketplace/curated/{mp}/{plugin}/mirrored/docs/...` to any RBAC-permitted user.

**Fix:** custom `HTTPRedirectHandler.http_error_30x` that re-runs `_is_safe_url` on every hop, OR `max_redirections=0` and follow manually with the SSRF check on each leg.

#### 2. SSRF via DNS rebinding · **P0**
`src/marketplace_asset_mirror.py:168-187, 291-293`

`_is_safe_url()` calls `socket.getaddrinfo` once for validation; `urllib.request.urlopen` performs a second DNS lookup. Attacker DNS returns a public IP for the validation lookup and `169.254.169.254` / `127.0.0.1` for the connection lookup (TTL=0).

**Fix:** resolve once in `_is_safe_url`, then connect to that exact IP and pass `Host: <hostname>` header. Same shape as the redirect fix — both want a custom opener.

#### 3. Stored XSS via `/asset/{path}` content-type sniff · **P1**
`app/api/marketplace.py:1512-1540`

Unlike `/doc/` (extension allowlist) and `/mirrored/` (mirror-validated body), `/asset/{path}` serves **any file in the cloned repo** with `Content-Type` from stdlib `mimetypes`. A curator drops `evil.html` in `.agnes/`, references via `/asset/evil.html` — served as `text/html` on the same origin as `/admin` and `/api/me/*`. Full session-cookie XSS.

**Fix:** apply `IMAGE_EXTENSIONS` allowlist + `validate_image_file` magic-byte sniff, mirroring the `/doc/` pattern.

---

### ⚠️ High-priority correctness

#### 4. Cross-plugin URL collision — wrong-plugin RBAC · medium
`src/marketplace_asset_mirror.py:393, 496-528` + `src/marketplace.py:244-253`

Cache key is URL-only but the local file path embeds plugin_name. Plugins A and B both reference the same external image → last writer wins on `manifest[url].plugin_name`; plugin B's served URL ends up pointing under plugin A's tree. Users with B-grant but no A-grant get 403 on B's cover.

**Fix:** key the manifest by `(plugin_name, url)`; per-plugin manifest entries.

#### 5. Curator "mandatory" only at create-time · medium
`app/api/marketplaces.py:362-377` + `CHANGELOG.md`

PR description says curator is mandatory. Pre-v32 rows have `curator_*=NULL`. UPDATE handler treats blank as no-change → admin can edit URL/description forever without filling curator. Existing rows still render `OWNER_TODO_PLACEHOLDER`.

**Fix:** either reject UPDATE when existing curator is NULL and payload doesn't fill it, OR ship a startup log warning + admin-UI nudge listing rows missing curator. Document the gap in CHANGELOG.

#### 6. Dead code: `build_db_payload` contradicts strict-drop semantics · medium
`src/marketplace.py:201-207, 321` + `src/marketplace_metadata.py:289-351`

`build_db_payload` is imported but never called — the strict-drop logic was reimplemented inline in `_refresh_plugin_cache`. The unused function still falls back to the original external URL on mirror failure, contradicting the "drop entries Agnes can't deliver" contract. A future contributor who re-wires it gets silent regressions.

**Fix:** delete `build_db_payload` (and its import) or rewrite to match strict-drop.

#### 7. Manifest-write ordering / orphan files · medium
`src/marketplace_asset_mirror.py:235-244 vs 359-365`

Bodies write atomically per-URL; manifest persists only at end of batch. `kill -9` mid-batch leaves on-disk orphans the manifest never references; next sync re-fetches into a different `sha8`-keyed path → disk bloat over time.

**Fix:** write manifest after each successful body, or defer unlink-of-old until manifest persists.

#### 8. No in-flight dedup · medium
`src/marketplace_asset_mirror.py:403-407`

Two plugins linking the same URL = two parallel HTTP fetches via `pool.map`. Wasted bandwidth, rate-limit pressure on slow CDNs, last-writer race on manifest.

**Fix:** dedupe `requests` by URL before pool submission; track owning plugin in a separate map.

#### 9. `agnes-metadata.json` size + depth uncapped · low → medium
`src/marketplace_metadata.py:67-82`

`path.read_text()` then `json.loads()` — no size cap, no depth limit. A curator-controlled multi-GB file OOMs the sync worker. `RecursionError` on deeply-nested JSON propagates past the `ValueError` catch and aborts sync.

**Fix:** `path.stat().st_size` check (~1 MB cap), broaden the exception catch.

---

### 📋 Maintainability cleanups (follow-up OK)

| # | Issue | Where | Fix |
|---|---|---|---|
| 10 | `_path_under` is a clone of `_safe_join` (same file) | `app/api/marketplace.py:1125, 1480` | Drop `_path_under`, reuse `_safe_join` |
| 11 | URL builders inlined twice (`_internal_*_url` in `src/marketplace.py` + `_mirrored_url` / `_resolve_external_via_mirror` in `app/api/marketplace.py`) — author flagged "must stay aligned" | both files | Move builders to one module |
| 12 | `.video-wrap` + YouTube/Vimeo embed-detection JS duplicated across `marketplace_plugin_detail.html` + `marketplace_item_detail.html` | both templates | Reuse existing `.video-embed` from `style-custom.css` (CLAUDE.md CSS-consolidation pattern) |
| 13 | `marketplace_format_guide.html` carries 60-line inline `<style>` | template | Move to `style-custom.css` or reuse `.news-content` |
| 14 | `marketplace_assets` vs `marketplace_asset_mirror` naming overlap; `marketplace_assets` mixes parser + validator + HTTP-response helpers | 3 new `src/marketplace_*.py` | Rename `marketplace_assets.py` → `marketplace_asset_validation.py`, OR fold parsers into `marketplace_metadata.py` |
| 15 | `_resolve_marketplace_name` is a one-line shim with no remaining callers | `app/api/marketplace.py:294-302` | Delete |
| 16 | Asset mirror uses `urllib.request` while `httpx` is already in tree | `src/marketplace_asset_mirror.py:42-339` | Either justify in comment or migrate to `httpx.Client` |
| 17 | Wrong `# noqa: F401 Optional kept for forward-compat` — `Optional` IS used | `src/marketplace.py:24` | Remove the noqa + comment |
| 18 | Hardcoded operational caps (timeout/body/concurrency) — no operator knob | `src/marketplace_asset_mirror.py:72-95` | Document the env-var name to add later or surface as `instance.yaml` keys |

---

### 🧪 Test gaps

- **No redirect test** in `test_marketplace_asset_mirror.py` (would confirm finding 1)
- **No DNS-rebinding test** (mock `socket.getaddrinfo` returning different IPs on consecutive calls)
- **No `/asset/foo.html` Content-Type test** (would confirm finding 3)
- `test_curated_asset_endpoint_blocks_path_traversal` uses literal `..` — TestClient/httpx normalizes before send. Test passes even if `_path_under` is broken. Use `%2E%2E%2F` to defeat client-side normalization, or call the asset handler directly.
- `test_git_tree_strips_agnes_only_files` reimplements the inner loop instead of calling `file_set_for_user` — regressions in the real code path won't fail the test.
- No test for cross-plugin URL collision (finding 4).
- No test for kill-9 between body-write and manifest-write (finding 7).
- No test for partial-failure (cover OK, doc fails — plugin still renders).
- Tests mock `_is_safe_url` away; end-to-end SSRF flow only exists at the helper level.
- No idempotency test for v32 migration on a populated v31 DB.

---

### ✅ Calibration — what the PR got right

- Magic-byte validation on PNG / JPEG / WebP / PDF (rename-extension bypass blocked).
- Atomic writes (tmp + rename) for both manifest and body files.
- `Path.resolve(strict=True) + relative_to` containment in `_safe_join` and `_path_under`.
- RBAC `require_resource_access(MARKETPLACE_PLUGIN, ...)` on every new endpoint.
- Body cap enforced via `read(MAX_BODY_BYTES+1)` + size check (correctly streams-bound).
- `Content-Disposition: attachment` on doc downloads (mitigates PDF-JS auto-exec in browser).
- Idempotent v32 migration (`ALTER TABLE … ADD COLUMN IF NOT EXISTS`) registered correctly + columns also in `_SYSTEM_SCHEMA` for fresh installs.
- Wikipedia-policy-compliant User-Agent.
- 304 conditional-GET handling preserves on-disk file (no cache-mtime poisoning).

---

### 🎯 Recommended merge gate

Block on **#1, #2, #3** (security blockers) + **#5** (curator UPDATE flow) + **#6** (delete dead `build_db_payload`).

Land in follow-up PR: #4, #7, #8, #9 (correctness mediums) and the maintainability cleanups.

Tests must add: redirect-bypass + DNS-rebinding + asset-Content-Type + the URL-encoded path-traversal fix in #234's own test files (regression guard for the very surface being introduced).

minasarustamyan:
## Review feedback addressed

Thanks for the thorough review @cvrysanek — every blocker plus the correctness, test-gap, and maintainability items are addressed below. Each point lists the commit that resolved it.

### 🚨 P0 / P1 blockers

| # | Issue | Commit | Notes |
|---|---|---|---|
| 1 | SSRF via redirect (cloud-metadata exfil) | `58c473c` | Initial fix on `urllib`; later replaced wholesale by the `_SSRFGuardTransport` in `3568c1f` (see #16) |
| 2 | SSRF via DNS rebinding | `58c473c` | Same path: pinned-IP connect with original hostname preserved for TLS SNI / `Host` header. After #16 the pin moves into the httpx transport (URL host rewrite). |
| 3 | Stored XSS via `/asset/{path}` content-type sniff | `ffb4616` | Endpoint is now image-only with three layered checks: `IMAGE_EXTENSIONS` allowlist → `validate_image_file` magic-bytes → `Content-Type` pinned from the validated extension (never stdlib mimetypes). SVG is intentionally excluded — `<script>` inside SVG executes. Defense-in-depth headers `X-Content-Type-Options: nosniff` + a strict `Content-Security-Policy: default-src 'none'; img-src 'self'; style-src 'unsafe-inline'` on every response. |

### ⚠️ Correctness mediums

| # | Issue | Commit | Notes |
|---|---|---|---|
| 4 | Cross-plugin URL collision (RBAC bug) | `a954c45` | Manifest is now keyed by `(plugin_name, url)` in memory; on disk the format flips from `{url: entry}` dict to a `[entry, …]` list (entries are self-describing — no key encoding). Each plugin owns its own manifest entry pointing under its own subdir. Consumer code (`src.marketplace._refresh_plugin_cache` + `app.api.marketplace._resolve_external_via_mirror` / `_curated_inner_cover` / `_curated_inner_enrichment`) re-keyed in lockstep. |
| 5 | Curator mandatory only at create-time | `12b731a` | `PATCH /api/marketplaces/{id}` now rejects when the post-merge row would persist with empty curator (`400 curator_name is required` / `400 curator_email is required`). DB column stays nullable so untouched legacy rows aren't broken; the gate fires only the moment an admin opens the edit modal. Existing PATCH semantics preserved (empty-string still means "leave existing value alone"). |
| 6 | Dead code `build_db_payload` contradicts strict-drop | `1517bad` | Deleted (and its import + the comment that referenced it). The strict-drop logic was already re-implemented inline in `_refresh_plugin_cache`; leaving the helper around would have invited a silent regression on re-wiring. |
| 7 | Manifest-write ordering / orphan files | `5fb18cd` | Phase 2 now persists the manifest immediately after each body write, **before** unlinking the previous body. Crash window narrows from "all of Phase 2" to "between persist and unlink" (microseconds). Phase 3 follows the same persist-before-unlink discipline (collect to-delete relpaths → persist → unlink). A persist failure mid-batch keeps the previous body on disk so the on-disk manifest still references something real (stale beats 404). |
| 8 | No in-flight dedup | `a954c45` | Phase 1 deduplicates fetches by URL — three plugins sharing one URL share one HTTP request. Phase 2 still creates a per-`(plugin, url)` manifest entry pointing under the plugin's own subdir; body files stay per plugin (RBAC-clean isolation). |
| 9 | `agnes-metadata.json` size + depth uncapped | `8d61581` | `AGNES_METADATA_MAX_BYTES = 1 MiB` size cap checked against `path.stat().st_size` before `read_text()`. Parse `except` widened to `(ValueError, RecursionError)` so a deeply-nested document fitting under the cap can't crash the parser and abort sync for every other marketplace. Both failure modes degrade to the existing empty-dict fall-back. |

### 📋 Maintainability cleanups

| # | Issue | Commit | Notes |
|---|---|---|---|
| 10 | `_path_under` is a clone of `_safe_join` | `1ffa8e4` | Deleted; three callers in the v32 endpoints now share `_safe_join`. |
| 11 | URL builders inlined twice | `1ffa8e4` | New module `src/marketplace_urls.py` is the single source of truth for `internal_asset_url` / `internal_doc_url` / `mirrored_url`. Both call sites import from there; the previous "must stay aligned" comment is gone. |
| 12 | `.video-wrap` + YouTube/Vimeo embed JS duplicated across templates | `1420fc9` | Detection logic moved to a single `_marketplace_video_embed.html` partial that both `marketplace_plugin_detail.html` and `marketplace_item_detail.html` `{% include %}` inside their IIFE. The `.video-wrap` selectors collapse into the existing `.video-embed` 16:9 wrapper from `style-custom.css` (with new `.video-embed video` / `.video-embed a` child rules added so the wrapper handles all four embed shapes uniformly). |
| 13 | `marketplace_format_guide.html` 60-line inline `<style>` | `1420fc9` | Moved verbatim to `style-custom.css` under a new "Marketplace format guide page" section, scoped to `.format-guide`. |
| 14 | `marketplace_assets` vs `marketplace_asset_mirror` naming overlap | `1ffa8e4` | Renamed `src/marketplace_assets.py` → `src/marketplace_asset_validation.py` so the file's purpose (image / doc magic-byte validators + Content-Type allowlist + agnes-metadata parsers) is obvious from the name. Six call-site imports updated in lockstep. |
| 15 | `_resolve_marketplace_name` shim | `1517bad` | Deleted — no remaining call sites; callers use `_resolve_marketplace_meta` which returns name + curator together. |
| 16 | `urllib.request` vs `httpx` | `3568c1f` | Migrated `src/marketplace_asset_mirror.py` from `urllib.request` to `httpx`. The asset mirror was the only HTTP call site in Agnes still using `urllib.request`; everything else (CLI, Jira / OpenMetadata / OpenAI connectors, scheduler, Telegram bot) was already on `httpx`. The migration paid for itself: SSRF defence collapsed from five urllib classes into one `_SSRFGuardTransport` because httpx invokes `handle_request()` on every redirect hop (re-validation is automatic), DNS-rebinding pinning collapsed into a `URL.copy_with(host=ip)` rewrite + `sni_hostname` extension (no `HTTPSConnection` subclass needed), and error handling collapsed from `URLError` + `HTTPError` + manual unwrap into one `httpx.HTTPError` catch + specific `Timeout` / `TooManyRedirects` branches. The shared `httpx.Client` is built lazily at module load matching the `cli/client.py:_get_shared_client` pattern. Externally observable behaviour is unchanged. |
| 17 | Wrong `# noqa: F401 Optional kept for forward-compat` | `1517bad` | Comment removed — `Optional` IS used in `src/marketplace.py` (lines 70 and 238), so the noqa would have invited a future cleanup pass to delete a real import. |

### 🧪 Test gaps (review's "Tests must add" list)

- **Redirect-bypass test:** covered by `test_ssrf_transport_rejects_link_local_target` and `test_ssrf_transport_rejects_loopback_target` (`3568c1f`); plus `test_fetch_url_rejects_when_transport_raises_ssrf` for the end-to-end mapping to `FetchOutcome.status='rejected'`.
- **DNS-rebinding test:** `test_dns_rebinding_does_not_bypass_ssrf` (`3568c1f`) uses the real `_SSRFGuardTransport` with a monkey-patched `httpx.HTTPTransport.handle_request` to verify the URL host gets rewritten to the pinned IP before reaching the wire.
- **`/asset/foo.html` Content-Type test:** `test_curated_asset_rejects_html_extension`, `test_curated_asset_rejects_renamed_html_as_png`, `test_curated_asset_rejects_svg_extension`, `test_curated_asset_serves_valid_png_with_security_headers` (`ffb4616`). The renamed-HTML-as-PNG test specifically guards the magic-bytes path.
- **Cross-plugin URL collision test:** `test_sync_assets_two_plugins_same_url_keeps_per_plugin_entries` (`a954c45`) — verifies the composite-key manifest holds one entry per plugin even when the URL is shared.
- **Kill-9 between body-write and manifest-write test:** `test_sync_assets_persists_manifest_per_body_write` + `test_sync_assets_phase3_persists_before_unlinking_orphans` (`5fb18cd`) — spy on `_write_manifest` call count and on `Path.unlink` reading the on-disk manifest from inside the call.
- **v32 idempotency:** the v32 migration was already idempotent (`ALTER TABLE … ADD COLUMN IF NOT EXISTS`); the existing `test_db_schema_version.py` covers the populated-v31 → v32 path.

### 📊 Verification

- `tests/test_marketplace_asset_mirror.py`: **28/28 passed** (15 original + 13 new across SSRF/composite-key/manifest-ordering work)
- Marketplace sweep (6 files): **106/106 passed**, 2 skipped — no regressions across all the refactors
- One urllib-specific test was deleted without replacement (`test_fetch_url_unwraps_redirect_error_inside_urlerror`) because httpx doesn't wrap handler exceptions — not applicable post-migration.

### 🎯 Summary

All 17 review items are resolved across 10 commits. The PR keeps the same external behaviour (same `FetchOutcome` statuses, same DB columns, same served URLs, same UI), but the code that ships those behaviours is materially smaller and follows the project's existing conventions: `httpx` for HTTP, `_safe_join` for path containment, single-source URL builders, and a single SSRF transport instead of five urllib classes. RBAC is now correct on shared external assets (#4) and the curator gap on legacy rows (#5) is closed at the API layer without forcing a backfill migration.

Ready for re-review whenever you have a moment.

PR #237, opened 2026-05-09
Move marketplace plugin updates from hook to /update-agnes-plugins skill

minasarustamyan:
## Summary

The SessionStart hook used to run `agnes refresh-marketplace --quiet` on every Claude Code session, performing a full fetch+reset+install cycle silently in the background. That work was invisible, slowed startup, and unrecoverable interactively when something failed.

This PR splits the responsibility:

- **`agnes refresh-marketplace --check`** — new lightweight detector for the SessionStart hook. `git fetch` only, compares local `HEAD` with remote `FETCH_HEAD`, emits a Claude Code hook JSON message hinting the user at `/update-agnes-plugins` when the marketplace has changes. No `git reset`, no `claude plugin marketplace update`, no install/update side effects.
- **`/update-agnes-plugins` slash command** — installed by `agnes init` into `<workspace>/.claude/commands/`. Runs `agnes refresh-marketplace` (default chatty path). Output streams into the Claude Code transcript so the user sees install/update progress and can react to errors interactively.
- **SessionStart hook now runs `--check`.** Existing workspaces auto-upgrade on next `agnes init` (substring marker matches both old `--quiet` and new `--check` shape).

**BREAKING:** `agnes refresh-marketplace --quiet` is removed. Old hooks calling it silent-noop after the CLI upgrade (the hook's `|| true` swallows the unknown-flag error) until the user re-runs `agnes init`, which rewrites the hook + installs the slash command.

## Test plan

- [x] Touched modules green: `pytest tests/test_lib_hooks.py tests/test_lib_commands.py tests/test_cli_refresh_marketplace.py tests/test_setup_hooks_template.py tests/test_clean_install_integration.py` → 57/57 pass
- [x] New tests cover: `--check` detector behavior (HEAD-vs-FETCH_HEAD diff, hook JSON shape, no-side-effects, no-clone exit 0, fetch failure exit 1, mutual exclusion with `--bootstrap`); `install_claude_commands` write/idempotence/third-party-untouched; hook migration from old `--quiet` entry to new `--check` form
- [x] Pre-existing test failures verified independent of this change via stash-and-run on main

## Manual verification

- [ ] In a scratch workspace, run `agnes init …` and confirm `.claude/commands/update-agnes-plugins.md` and `.claude/settings.json` SessionStart hook contains `--check`
- [ ] Open Claude Code in the scratch workspace, run `/update-agnes-plugins`, confirm the full `agnes refresh-marketplace` output streams to the transcript
- [ ] Force a server-side marketplace change, restart the Claude Code session, confirm the SessionStart hook surfaces the "updates available" hint pointing at `/update-agnes-plugins`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,45 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Added
+
+- **`/update-agnes-plugins` slash command** — installed automatically by
+  `agnes init` into `<workspace>/.claude/commands/`. Runs
+  `agnes refresh-marketplace` (the chatty default mode) so the user sees
+  install/update progress streamed into the Claude Code transcript and
+  can react to errors interactively, instead of having a full reconcile
+  happen silently behind a SessionStart hook.
+
+- **`agnes refresh-marketplace --check`** — lightweight detector mode for
+  the SessionStart hook. Runs `git fetch` only, compares local `HEAD`
+  with remote `FETCH_HEAD`, and emits a Claude Code hook JSON message
+  pointing the user at `/update-agnes-plugins` when there are remote
+  changes. Silent when up to date. No `git reset`, no
+  `claude plugin marketplace update`, no plugin install/update side
+  effects.
+
+### Changed
+
+- **SessionStart marketplace hook is now read-only.** The hook installed
+  by `agnes init` was previously `agnes refresh-marketplace --quiet`,
+  which performed a full fetch+reset+install cycle on every session start
+  (slow, invisible to the user, not interactively recoverable). It now
+  runs `agnes refresh-marketplace --check` — detect-only — and surfaces a
+  hint to run `/update-agnes-plugins` when updates are available.
+  Existing workspaces auto-upgrade on next `agnes init` (the substring
+  marker `agnes refresh-marketplace` matches both the old and new entry
+  shapes, so the idempotent-replace path correctly rewrites them).
+
 ### Removed
+
+- **BREAKING: `agnes refresh-marketplace --quiet` flag.** Replaced by
+  `--check` (detect-only) and the new `/update-agnes-plugins` slash
+  command (interactive update). Existing SessionStart hooks calling
+  `--quiet` will silent-noop after the CLI upgrade — the hook's
+  `2>/dev/null || true` swallows the unknown-flag error — until the user
+  re-runs `agnes init`, which rewrites the hook to use `--check` and
+  installs the slash command. Dashboard `/setup` flow re-runs
+  `agnes init` automatically on next paste.
 
 - **BREAKING: legacy `git config --global http.<host>.sslVerify=false`
   downgrade in the install setup prompt.** The marketplace step (step 5)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,15 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   marker `agnes refresh-marketplace` matches both the old and new entry
   shapes, so the idempotent-replace path correctly rewrites them).
 
+- **Marketplace "Added to your stack" hint points at `/update-agnes-plugins`.**
+  The post-install green panel on plugin and skill/agent detail pages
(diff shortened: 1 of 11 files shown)


cvrysanek:
## Code review

### Overview
Splits silent SessionStart `agnes refresh-marketplace --quiet` (full fetch+reset+install on every session) into two paths:
- `--check` — detector only (`git fetch` + HEAD-vs-FETCH_HEAD compare), emits hook JSON hint
- `/update-agnes-plugins` slash command — runs default chatty `agnes refresh-marketplace`, output streams into the Claude Code transcript

Adds `cli/lib/commands.py:install_claude_commands` (workspace-scoped, idempotent, third-party-safe), wired into `agnes init`. Updates UI hints. **BREAKING:** removes `--quiet`.

### Strengths
- **Migration path is correct.** `_OUR_COMMAND_MARKERS` matches by substring on `agnes refresh-marketplace`, so existing `--quiet` entries auto-rewrite to `--check` on the next `agnes init`. Pinned by `tests/test_lib_hooks.py`.
- **Failure modes are well bounded.** Hook wraps in `bash -c \"... 2>/dev/null || true\"` — unknown-flag silent-noop is the documented degradation path until next init. `--check` on a no-clone workspace exits 0 silently (matches old `--quiet` semantics).
- **Mutual exclusion.** `--check` + `--bootstrap` → exit 2; tested.
- **Idempotent overwrite + third-party safety.** `_MANAGED_COMMANDS` enumerates owned files; the rest of `.claude/commands/` is untouched. Tested explicitly.
- **Test coverage is solid.** 57/57 on touched modules. New tests cover hook JSON shape, no-clone, fetch failure exit 1, rev-parse staging, frontmatter intact, user-edit overwrite, and third-party-survives.
- **Hook semantics simplified.** Detector fast-path + interactive recovery is the right separation — invisible reconcile failures were the prior pain point.

### Issues / Risks

**Medium**
- **`pyproject.toml` is not in the diff.** `cli/lib/commands.py` claims hatchling bundles `cli/templates/commands/` because `pyproject.toml` declares `cli` as a wheel package. Worth verifying that `[tool.hatch.build.targets.wheel]` actually picks up the new `templates/commands/*.md` (the sibling `templates/*.txt` may work via the existing package include, but a missing wildcard would silently fall through to `_FALLBACK_BODY` on every install). One-line check before merge:
  ```bash
  python -m build && unzip -l dist/*.whl | grep update-agnes-plugins.md
  ```
- **`_has_remote_changes` swallows rev-parse failures → returns False.** The intent is documented (\"stay quiet rather than surface a misleading hint\"), but a corrupt local clone now never surfaces — the user sees no hint, no error, no log. Consider emitting a `stderr` warning (still silent on stdout so the hook JSON contract is preserved).

**Minor**
- **`_FALLBACK_BODY` instructs \"Run `agnes refresh-marketplace`\"** — bare command, no transcript framing. If the fallback ever fires, behavior diverges from the canonical template. Low impact (broken installs only), but worth aligning the stub with the real template's intent.
- **HTML hint regression: loss of terminal alternative.** The old \"Added to your stack\" panel offered both `/reload-plugins` (in-session) and shell `agnes refresh-marketplace`. The new version only shows the slash command. Users in a workspace without an open Claude Code session lose the immediate-action path. Consider keeping a collapsed \"or run `agnes refresh-marketplace` in terminal\" line.
- **`_emit_check_hook_message` summary is identical for `systemMessage` and `additionalContext`.** Toast and system-reminder serve different purposes; the old `_emit_hook_message` differentiated naturally via the events list. Acceptable here since the detector has no list to render, but consider letting `additionalContext` say *why* (remote HEAD changed) while `systemMessage` stays as the user-facing CTA.
- **Confirm no leftover `--quiet` references in tests.** `grep -r '\\-\\-quiet' tests/` to make sure the removal didn't leave a dangling expectation.

**Nit**
- `_templates_dir()` doc says \"two `.parent` hops\" — there are three (`__file__` → file → `lib/` → `cli/`). The code is correct; the comment count is off.

### Security / Performance
- No new attack surface. `--check` reduces work on every session start (no `git reset`, no `claude plugin marketplace update`, no install).
- Net win for cold-session latency: shell calls per session drop from full reconcile to `git fetch` + 2× `rev-parse`.
- `AGNES_TOKEN` propagation is unchanged — still env-only, not argv.

### Verdict
Approve with one pre-merge check — confirm `pyproject.toml` ships `cli/templates/commands/*.md` in the wheel. The other items are optional follow-ups. Architecture, migration path, and test coverage are all good.

PR #239, opened 2026-05-09
Flea-market edit feature with version history (schema v37)

cvrysanek:
## Summary

Real edit page for flea-market entities, replacing the "Edit (coming soon)" placeholder. Each bundle update creates a new version; prior versions persist on disk for rollback.

## What's in the box

**Schema v37**
- `store_entities.version_no INTEGER NOT NULL DEFAULT 1` — current version index.
- `store_entities.version_history JSON DEFAULT '[]'` — append-only metadata for each version.
- Backfill: existing rows seeded with v1 from the row's current `version` hash.

**Edit page** (`/marketplace/flea/{id}/edit`, owner + admin only)
- Pre-filled metadata: name, description, category, video URL, cover photo.
- Optional bundle re-upload — creates v<N+1>.
- Type is locked (400 `type_locked`).
- Block-while-pending: form disabled + 409 `prior_version_pending` during in-flight review.

**Versioning** on `PUT /api/store/entities/{id}`
- New bundle bakes into `${DATA_DIR}/store/<id>/versions/v<N+1>/plugin/`.
- Guardrails run on the new version dir (atomic — failed checks leave live untouched).
- On approval: live `plugin/` dir copied from version dir; `version_no` bumps; `version_history` appends.
- Initial create now also seeds `versions/v1/plugin/` so restore can copy v1 forward.
- Display-name change renames on-disk slug for both live + version dirs (reuses rename-on-archive helper).

**Restore** — `POST /api/store/entities/{id}/versions/{n}/restore` (owner + admin)
- Copies vN bundle forward as v<max+1>; re-runs guardrails.
- Forward-only history. Original vN row keeps its verdict.
- 404 `version_not_found` / 400 `already_current` / 409 `prior_version_pending`.

**UI**
- Versions card on plugin + item detail templates (`_flea_versions.html` partial) — owner/admin only, newest-first, current badge, Restore button on non-current rows.
- Edit button on detail page replaces "Edit (coming soon)"; disabled with tooltip during pending review.
- Quarantine banner copy adapts for v<N+1> review window: *"Version 4 under review — v3 keeps serving to existing installers"*.

## Test plan
- [x] 13 new tests in `tests/test_store_entity_versions.py`:
  - metadata-only edit no version bump
  - bundle edit bumps version + appends history
  - type change rejected
  - block-while-pending → 409
  - name change renames slug on disk
  - restore creates new version with old bundle hash
  - restore already-current → 400
  - restore unknown version → 404
  - non-owner non-admin can't restore
  - edit page 200 for owner / 404 for non-owner
  - versions card renders for owner / hidden for non-owner
- [x] 151 tests green across the guardrails + edit suites.
- [x] Migration v36 → v37 is idempotent + backfill verified.

## Risks
- **Storage cost**: each version persists its own bundle. ~bundle-size per version per entity. No prune yet — admin can hard-delete the entity to free everything; per-version pruning is a follow-up if disk usage grows.
- **Consumer-side rename on display-name change**: same trade-off as rename-on-archive. Existing installers re-resolve under the new slug on next sync.
- **Forward-only history on restore**: restored bundle re-validates against today's rules. If a bundle was approved at v3 but today's static-scan would flag it, restore lands as `blocked_inline` and live stays at the previous current.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,48 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   `claude plugin marketplace update`, no plugin install/update side
   effects.
 
+- **Flea-market entity edit feature with version history (schema v38).**
+  Owner + admin can now edit a store entity from a real Edit page at
+  `/marketplace/flea/{id}/edit` (replaces the prior "coming soon"
+  placeholder). Editable fields: display name, description, category,
+  video URL, cover photo, and an optional new bundle. Type is locked
+  (400 `type_locked` on change attempt). Display-name change renames
+  the on-disk slug for both the live `plugin/` dir and the version
+  dir, mirroring the rename-on-archive flow.
+
+  Each bundle update creates a new version: bytes bake into
+  `${DATA_DIR}/store/<id>/versions/v<N+1>/plugin/`, run the standard
+  guardrails pipeline. **Deferred promotion:** the live `plugin/` dir
+  and `entity.version_no` stay at the prior approved version through
+  the LLM review window, so existing installers keep receiving the
+  previously approved bundle while the new version is being
+  validated. Promotion (live swap + version_no/version/file_size
+  bump) happens only on LLM approval; if the new version is blocked,
+  installers continue serving the prior approved version
+  indefinitely. The entity row carries `version_no` (current served
+  index) and `version_history` JSON (append-only per-version
+  metadata: hash, sha256, size, submission_id, created_at,
+  created_by). Existing entities backfill to v1 with a single-entry
+  history seeded from the row's current `version` hash.
+
+  **Block-while-pending:** an in-flight LLM review blocks any further
+  edit with 409 `prior_version_pending`. Owner waits ~5-30s; the
+  detail page Edit button renders disabled in the same window.
+
+  **Rollback:** new endpoint `POST /api/store/entities/{id}/versions/{n}/restore`
+  (owner + admin) copies a prior version's bundle forward as
+  v<max+1> and re-runs guardrails. Forward-only history — the
+  original row keeps its verdict; the new copy gets a fresh one.
+  Detail page renders a Versions card with restore buttons for
+  owner/admin only.
+
+  **Admin queue** gains a `v#` column (with "current" badge) and a
+  separate Hash column. Submission detail page surfaces Version +
+  Bundle hash rows. Activity timeline splits into per-submission +
+  entity-wide cards so admins can tell version-scoped events apart
+  from entity-wide ones; entity-wide rows render `vN` chips when the
+  audit row's params reference a version.
+
 ### Changed
 
 - **SessionStart marketplace hook is now read-only.** The hook installed
--- a/app/api/store.py
+++ b/app/api/store.py
@@ -1303,19 +1303,24 @@ async def update_entity(
     * **Type is locked** — passing ``type`` that differs from the
       stored row returns 400 ``type_locked``. Replacing one form
       factor with another is a fresh upload, not an edit.
-    * **Display-name change** is allowed — triggers a slug rename in
-      the live bundle (mirrors the rename-on-archive flow). Existing
-      installers see the plugin renamed on next sync.
+    * **Display-name change** is allowed. Without a bundle change it
+      flips the live slug immediately (mirrors rename-on-archive).
+      Combined with a bundle change the rename is deferred — only the
+      staged version dir is renamed; live keeps the prior slug until
+      promotion, so existing installers never see a slug≠content pair
+      mid-review.
     * **Bundle change** creates a new version: bake into
       ``versions/v<N+1>/plugin/``, run guardrails, on approval copy
       to the live ``plugin/`` dir + bump ``version_no`` + append
       ``version_history``. The prior version dir stays so rollback
       can copy it forward.
-    * **Block-while-pending**: an in-flight LLM review (the entity's
-      ``visibility_status='pending'`` AND the latest submission has
-      ``status IN ('pending_inline','pending_llm')``) blocks any
-      further edit with 409 ``prior_version_pending``. Owner waits
-      for the verdict; the detail page auto-refreshes.
+    * **Block-while-pending**: gates on the latest submission's status
+      directly (``status IN ('pending_inline','pending_llm')``),
+      independent of ``visibility_status``. Under deferred promotion
+      v2+ edits leave the entity ``approved`` through the LLM review
+      window, so a visibility-only check would never fire. Returns 409
+      ``prior_version_pending``; owner waits for the verdict; the
+      detail page auto-refreshes.
     * **Metadata-only edit** (no ``file`` posted) skips the bundle
       pipeline and the version bump.
     """
--- a/app/api/store.py
+++ b/app/api/store.py
@@ -637,6 +637,62 @@ def _write_synth_plugin_json(
     )
 
 
+def _swap_live_to_version(entity_id: str, version_no: int) -> bool:
+    """Replace the live ``plugin/`` dir with a copy of the named
+    version's contents. Used by the guardrails-disabled promote path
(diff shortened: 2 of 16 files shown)


minasarustamyan (commented):
Code review focused on critical issues and bugs. Did not cover the UX polish or positive feedback — flagging only the things that look broken.

## 🔴 Critical

### 1. Block-while-pending gate doesn't fire for v2+ edits

`app/api/store.py:1322-1338` in `update_entity`:

```python
if entity.get("visibility_status") == "pending":
    latest_sub = ...
    if latest_sub and latest_sub.get("status") in ("pending_inline", "pending_llm"):
        raise HTTPException(409, "prior_version_pending")
```

Under deferred promotion, v2+ edits leave the entity at `visibility_status='approved'` (PUT no longer flips it to 'pending'). The gate therefore never fires for a v2→v3 edit during an in-flight LLM review. Two concurrent PUT edits will create `versions/v2/plugin/` and `versions/v3/plugin/` in parallel, and the runner will sequentially promote both as their verdicts land.

`test_block_while_prior_pending_409` masks this — it manually forces `visibility_status='pending'`, which doesn't naturally happen on the v2+ flow.

**Fix**: gate on `subs_repo.latest_for_entity(entity_id).status` directly, independent of `entity.visibility_status`.

### 2. Display-name change combined with bundle change breaks the deferred-promotion contract

`app/api/store.py:1500-1554`. When the user submits both `name` and `file`:

1. v2 dir bakes the new bytes
2. **Live `plugin/` is renamed immediately** (`_rename_baked_tree(plugin_dir=_plugin_dir(entity_id), ...)`) — but still holds v1's content
3. v2 dir is also renamed
4. LLM review runs against v2

During the review window, existing installers receive a **renamed v1** (the slug changed in marketplace.zip / .git, the frontmatter `name` too) — i.e. a slug≠content pairing. If the LLM blocks v2, that broken state becomes permanent — installers will never receive v1 back under its original name.

This breaks the central "existing installers keep receiving the previously approved bundle" guarantee that the PR description (and the test class `TestInstallerAlwaysGetsLatestApproved`) calls out as the contract.

**Fix**: defer the live `plugin/` rename into `_swap_live_to_version` (post-promotion). Metadata-only renames (no bundle) staying immediate is fine.

---

## 🟡 Bugs

### 3. Misleading audit log on v2+ approval

`src/store_guardrails/runner.py:209-275`. For a v2+ approval:

- `set_visibility_if_pending(entity_id, \"approved\")` returns `False` (entity is already 'approved', not 'pending'/'hidden')
- promote_version + `_swap_live_to_version` run correctly
- but `applied=False` → audit log writes `store.submission.bg_verdict_skipped` with `reason=\"entity left review window before LLM verdict landed (admin action)\"`
- `store.submission.approved` is never logged

Operators triaging the admin queue will see \"skipped\" rows for what were actually successful approvals + promotions. Persistent source of confusion.

**Fix**: decouple the promote logic from the `applied` flag. Always log `store.submission.approved` when `passed=True`; only log `bg_verdict_skipped` when the entity was archived/hard-deleted between BG-task start and verdict-write.

### 4. Stale comment in the blocked-LLM branch

`src/store_guardrails/runner.py:283`: *\"Entity stays at visibility_status='pending' — invisible in the flea browse for non-admins.\"* True for v1 uploads. For v2+ blocked edits the entity stays `approved` and remains visible. Comment misleads.

### 5. Race in `_swap_live_to_version`

`app/api/store.py:640-675`: `os.rename(live, backup)` → `shutil.copytree(source, live)`. In the gap, live `plugin/` doesn't exist. The ZIP packer (`/marketplace.zip`) or a git fetch landing in that window will miss the entity. Narrow window (copying ~MB of bytes), but real on slower disks.

**Fix**: copytree into `plugin.staging-XXX/` first, then atomic rename `live → backup`, `staging → live`.

### 6. `version_no` mismatch between the two audit rows

`app/api/store.py:1614-1620` logs `new_version_no = entity.version_no + 1` (planned), but the submission audit uses `appended_n` from `append_version_history` which is `max(history.n) + 1`. Under concurrent edits (see #1) the two diverge — the entity audit and the submission audit will carry different `version_no` for the same operation.

### 7. version-dir rename failure silently swallowed

`app/api/store.py:1547-1554`: if the live rename succeeds but the version-dir rename raises, the slugs diverge. A later restore of that version then copies the stale slug back into live.

---

## TL;DR

Need to address at least #1 and #2 before merge — both directly violate the deferred-promotion contract that's the centerpiece of this PR. #3 is observability; not a blocker but worth doing in the same iteration. Rest is fine as follow-up.

cvrysanek:
Re-posting with explicit reviewer→fix mapping so it's clear nothing got swapped:

**#1 — Block-while-pending gate doesn't fire for v2+ edits**
Reviewer's fix: *"gate on `subs_repo.latest_for_entity(entity_id).status` directly, independent of `entity.visibility_status`"*.
Done in `app/api/store.py:1326` (PUT) + `app/api/store.py:1668` (restore). Test: `TestPRReviewFixes::test_block_while_pending_fires_for_v2_edit_under_deferred_promotion`.

**#2 — Display-name change combined with bundle change breaks deferred promotion**
Reviewer's fix: *"defer the live `plugin/` rename into `_swap_live_to_version` (post-promotion). Metadata-only renames (no bundle) staying immediate is fine."*
Done in `app/api/store.py:1511` — `if file is None: rename live; else: rename ONLY the version dir`. Live catches up at promotion time when `_swap_live_to_version` copies the renamed version dir contents. Test: `test_name_change_with_bundle_does_not_rename_live_until_promote`.

**#3 — Misleading audit log on v2+ approval**
Reviewer's fix: *"decouple the promote logic from the `applied` flag. Always log `store.submission.approved` when `passed=True`; only log `bg_verdict_skipped` when the entity was archived/hard-deleted"*.
Done in `src/store_guardrails/runner.py:204+`. Audit now reads the entity's actual visibility post-flip-attempt: serve-able (`approved`) → log `approved`; archived/hidden → log `bg_verdict_skipped` with the actual current state. Two tests: `test_v2_approval_logs_approved_not_skipped` (positive) + `test_bg_verdict_skipped_when_admin_archives_during_review` (negative).

**#4 — Stale comment in blocked-LLM branch**
Comment in `src/store_guardrails/runner.py` rewritten to distinguish v1 (visibility=pending) vs v2+ (entity stays approved at prior version).

**#5 — Race in `_swap_live_to_version`**
Reviewer's fix: *"copytree into `plugin.staging-XXX/` first, then atomic rename `live → backup`, `staging → live`"*.
Done in `app/api/store.py:638`. Multi-MB copy goes to staging while live serves prior bundle; rename pair at the end.

**#6 — `version_no` mismatch between two audit rows**
Entity-update audit at `app/api/store.py:1665` now reads `max(history.n)` post-append (the actually-appended n) instead of the planned `new_version_no` baseline. Both audits reference the same value.

**#7 — version-dir rename failure silently swallowed**
The defer-live-rename fix from #2 makes #7 unreachable for the live-vs-version-dir mismatch case. The remaining version-dir rename branch now raises HTTPException 500 `version_rename_failed` instead of just logging.

Force-pushed. 28 entity-version tests + 235 across the wider guardrails / edit / admin / repo / schema sweep, all green.

minasarustamyan (approved):
LGTM. All 7 findings from my prior review are addressed in `f61984b`:

- **#1, #2** (deferred-promotion contract): block-while-pending gate now reads submission status directly; live-dir rename deferred to `_swap_live_to_version` when a bundle change is in flight. Both fixes carry rationale comments at the call sites.
- **#3** (audit observability): runner decouples `visibility_flipped` / `promoted_to` / `superseded_reason` and logs `store.submission.approved` whenever the entity is in a serve-able state, falling back to `bg_verdict_skipped` only on actual archive/hidden races.
- **#5**: swap is now staging-first, closing the missing-live window.
- **#4, #6, #7**: comment, audit `version_no` consistency, and version-dir rename failure raised properly.

Targeted regression tests added under `TestPRReviewFixes` for #1, #2, and both #3 outcomes. Ship it.

Issue #10, opened 2026-04-21
Unauthenticated HTML routes return JSON 401 instead of redirecting to /login


ZdenekSrotyr:
## Bug

Opening any protected web UI URL while unauthenticated returns a raw JSON 401 response instead of redirecting to the login page.

### Reproduction

1. Open a fresh browser (no cookie) and navigate to `http://<host>:8000/dashboard`.
2. Expected: redirect to `/login` (optionally with `?next=/dashboard` to come back after sign-in).
3. Actual: blank page with JSON body:

```json
{"detail": "Missing or invalid Authorization header"}
```

HTTP status: `401 Unauthorized`.

### Why this happens

HTML routes in `app/web/router.py` depend on `get_current_user` (`app/auth/dependencies.py:22`), which raises `HTTPException(401, "Missing or invalid Authorization header")` when no token/cookie is present. That exception is rendered as JSON by FastAPI's default handler — appropriate for API routes, wrong for HTML routes.

`/` is the only route that handles this correctly: it uses `get_optional_user` and redirects to `/login` when the user is `None` (`app/web/router.py:174-178`).

### Scope

Affects every HTML route that depends on `get_current_user`, e.g.:

- `/dashboard`
- `/catalog`
- `/corporate-memory`
- `/corporate-memory/admin`
- …and all other pages in `app/web/router.py` gated by `get_current_user` / `require_role(...)`

API routes under `/api/*` should keep the current JSON 401 behavior — this issue is only about HTML pages rendered from `app/web/`.

### Proposed fix

Add an HTML-aware auth dependency (or a global exception handler scoped to the web router) that, on missing/invalid credentials, returns `RedirectResponse("/login?next=<original_path>", status_code=302)` instead of raising `HTTPException(401)`.

Sketch:

```python
# app/auth/dependencies.py
async def get_current_user_web(
    request: Request,
    authorization: Optional[str] = Header(None),
    conn: duckdb.DuckDBPyConnection = Depends(_get_db),
) -> dict:
    user = await get_optional_user(request=request, authorization=authorization, conn=conn)
    if not user:
        raise HTTPRedirectException(f"/login?next={request.url.path}")
    return user
```

…combined with a small exception handler that turns `HTTPRedirectException` into a `RedirectResponse`. Alternative: a FastAPI `exception_handler` on the web router that intercepts `HTTPException(401)` and redirects only when the request `Accept` header prefers HTML.

### Acceptance criteria

[...]

PR #242, opened 2026-05-10
Session capture queue, private session, and setup-prompt fixes

minasarustamyan:
## Summary

Two stacked features that work together to make `agnes push` upload only the right sessions.

**1. Session capture queue + single-instance lock** (commit 1)
- New SessionStart hook `agnes capture-session` reads `transcript_path` from Claude Code's documented stdin JSON and queues it in `<workspace>/.claude/agnes-sessions.txt`. Sidesteps Claude Code's internal cwd-to-folder encoding (which is a moving target across versions).
- `agnes push` consumes the queue under a cross-platform single-instance lock (`filelock`: fcntl on POSIX, msvcrt on Windows). Concurrent SessionEnd hooks — common when the user closes several sessions at once — silent-exit on the losers.
- Atomic queue → snapshot rename prevents new hooks' writes from being clobbered mid-push. Recovery picks up snapshots from crashed runs. Failed uploads requeue.
- Removed the previous SessionStart `agnes push` self-heal entry — the queue persists across runs, so orphans from headless / crashed sessions ship out on the next SessionEnd push naturally.

**2. `/agnes-private` slash command + statusLine indicator** (commit 2)
- New slash command runs `agnes mark-private` deterministically via `` !``-prefix `` direct bash (no AI tokens, no AI decision in the loop).
- Authoritative state file `<workspace>/.claude/agnes-sessions-private.txt` (one session_id per line). Both `capture-session` and `push` consult it. This makes the slash-command / SessionStart-hook race impossible by construction:
  - Mark before capture → capture sees the ID, skips the queue write.
  - Capture before mark → push filters out the entry and appends to `agnes-sessions-private-skipped.txt` (audit log).
- New `agnes statusline` subcommand wired to Claude Code's `statusLine` setting. Renders `🔒 agnes-private` when the current session is private, empty otherwise. Optimized for the high call frequency.
- Polite install: if the workspace `settings.json` already has a `statusLine`, the install leaves it alone and warns on stderr.

## Test plan

- [ ] `pytest tests/test_session_queue.py tests/test_capture_session.py tests/test_cli_push.py tests/test_lib_hooks.py tests/test_lib_commands.py tests/test_private_list.py tests/test_mark_private.py tests/test_statusline.py` — all green (96 tests locally)
- [ ] Manual Windows: open Claude Code in an Agnes workspace, immediately type `/agnes-private` (race test) — verify queue stays empty, status bar shows `🔒 agnes-private`
- [ ] Manual Windows: open session, work for ~30s, then `/agnes-private` mid-session — verify session ends up in `agnes-sessions-private-skipped.txt` audit log, not on server
- [ ] Manual Mac (cross-platform): same flows
- [ ] Manual: close 3 sessions simultaneously — verify exactly one push runs, all three sessions reach the server
- [ ] Manual: re-run `agnes init` on a workspace with an existing user `statusLine` — verify the user's config is preserved + warning emitted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,7 +116,33 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   (e.g. `agnes self-upgrade` from `~/`). Failures are best-effort —
   they're surfaced on stderr but never flip the upgrade exit code.
 
+### Added
+
+- **Onboarding docs for the `/agnes-private` privacy feature.**
+  `config/claude_md_template.txt` gains a short "Private sessions"
+  subsection (next to "Data Sync") covering the slash command,
+  statusbar indicator, and audit-log location. The web-served setup
+  prompt (`app/web/setup_instructions.py`) gets a one-line mention so
+  analysts learn the feature exists at onboarding instead of by
+  accident.
+
+### Changed
+
+- **`_install_statusline` distinguishes explicit `null` / empty-string
+  `statusLine` from absent key.** Previously the `if existing:` truthy
+  check silently took the same path for all three cases. The new
+  `existing is None or existing == ""` branch documents and tests the
+  behavior (install ours — treated as "not configured" rather than
+  "explicit user opt-out"). Two new tests pin both edge cases.
+
 ### Fixed
+
+- **`agnes push --legacy-scan` help text documents the private-list
+  gap.** Legacy-scan entries carry an empty `session_id`, so the
+  `/agnes-private` filter is not consulted. The practical impact is
+  bounded — pre-queue sessions cannot have been marked private (the
+  private list is a queue-era feature) — but the help text now spells
+  out the gap so an operator running a backfill is not surprised.
 
 - **`agnes push` no longer crashes on filesystem errors when acquiring
   the single-instance lock.** `acquire_or_skip` in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,15 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ### Fixed
 
+- **`agnes push` no longer crashes on filesystem errors when acquiring
+  the single-instance lock.** `acquire_or_skip` in
+  `cli/lib/push_lock.py` now treats `OSError` (read-only filesystem,
+  permission denied on `.claude/`, disk full, hardware I/O failure) the
+  same as `filelock.Timeout` — yields `None`, push exits cleanly.
+  Previously the `OSError` propagated as an unhandled traceback;
+  invisible in the SessionEnd hook context (the `|| true` wrapper
+  swallowed it), but ugly in a manual `agnes push` invocation.
+
 - **`agnes push` no longer infinite-loops on permanent 4xx failures.**
   Previously any non-200 response except the literal `file not found
   on disk` was re-queued, so 401 (token expired), 403 (RBAC denial),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,29 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Fixed
+
+- **Session capture queue: concurrent SessionStart hooks no longer
+  corrupt the queue file on Windows.** `append_to_queue`,
+  `requeue_failed`, and `snapshot_queue` in `cli/lib/session_queue.py`
+  now hold a short-lived `agnes-queue.lock` (filelock) while writing.
+  Previously the code assumed Python's `open(path, "a")` is atomic on
+  NTFS for small writes; it isn't — the Windows CRT does not pass
+  `FILE_APPEND_DATA` to `CreateFile`, so concurrent appenders (e.g.
+  user opens several Claude Code windows simultaneously) could
+  interleave bytes mid-line and the parser would silently drop the
+  malformed entries. The lock is separate from `agnes-push.lock` —
+  capture-session hooks don't block on the push command.
+- **Session capture queue: snapshot filenames now include a uuid8 tail
+  so a recycled OS PID cannot silently overwrite a recovery snapshot
+  left behind by a crashed push.** `snapshot_queue` previously named
+  files `agnes-sessions.snapshot.<PID>.txt`; after a crash + PID reuse
+  (Linux default `kernel.pid_max=32768`), `os.rename` atomically
+  replaces the recovery file with the new snapshot, losing every entry
+  in it. New format: `agnes-sessions.snapshot.<PID>.<uuid8>.txt`;
+  `find_recovery_snapshots` already uses a glob so the change is
+  backward-compatible with snapshots written by older CLI versions.
+
 ### Changed
 
 - **Setup prompt + CLAUDE.md template: marketplace copy now reflects the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,19 +10,95 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Added
+
+- **Session capture queue + new `agnes capture-session` SessionStart
+  subcommand.** Replaces the previous encoding-based scan of
+  `~/.claude/projects/` for session jsonls (which depended on Claude
+  Code's cwd-to-folder encoding — a moving target across versions).
+  The hook reads Claude Code's documented stdin JSON
+  (`transcript_path`) and appends `<session_id>\t<transcript_path>` to
+  `<workspace>/.claude/agnes-sessions.txt`. `agnes push` then atomically
+  renames that queue to a snapshot, processes it, and re-queues failed
+  uploads. Recovery snapshots from a crashed push are picked up on the
+  next run. Concurrent SessionStart hooks (multiple Claude Code windows
+  opening at once) are serialized by a short-lived `agnes-queue.lock`
+  so the queue is race-free on every OS.
+
+- **`/agnes-private` slash command + `agnes mark-private` subcommand.**
+  Mark the current Claude Code session as private — its transcript is
+  skipped by `agnes push` and audit-logged to
+  `<workspace>/.claude/agnes-sessions-private-skipped.txt` instead.
+  The slash command runs deterministically via `!`-prefix bash (no AI
+  in the loop). State lives in
+  `<workspace>/.claude/agnes-sessions-private.txt` (one session_id per
+  line) and is the authoritative source — both `capture-session` and
+  `push` consult it, so the slash-command-before-capture and
+  capture-before-slash-command races both resolve safely without an
+  ordering dependency. Requires the `CLAUDE_CODE_SESSION_ID`
+  environment variable that Claude Code sets in every bash subprocess
+  it spawns; `agnes mark-private` exits 1 if missing (defends against
+  accidental invocations from a regular terminal).
+
+- **`agnes statusline` subcommand + statusLine wiring.** Renders
+  `🔒 agnes-private` in the Claude Code status bar when the current
+  session is marked private; empty string otherwise. `agnes init` wires
+  it to Claude Code's `statusLine` setting. Polite to existing
+  customizations — if the workspace `settings.json` already has a
+  `statusLine`, the install preserves it untouched and emits a
+  one-line stderr warning instructing the operator how to compose
+  `agnes statusline` into their own command.
+
+- **`agnes push --legacy-scan` opt-in fallback** scans
+  `~/.claude/projects/` via the pre-queue encoding-based path. Use
+  for one-off backfill of session jsonls that pre-date the queue
+  mechanism on workspaces upgrading from < v0.49. Note: legacy-scanned
+  entries have empty `session_id`, so the `/agnes-private` list filter
+  never matches — backfill uploads bypass the private list. Document
+  this gap before running a backfill on a workspace that has
+  previously-marked-private sessions in the encoding-based location.
+
+- **Single-instance lock for `agnes push`** (cross-platform via
+  `filelock`: `fcntl.flock` on POSIX, `msvcrt.locking` on Windows).
+  When the user closes several Claude Code sessions simultaneously,
+  every SessionEnd hook fires its own `agnes push` — exactly one
+  acquires `<workspace>/.claude/agnes-push.lock` and runs, the rest
+  silent-exit. Prevents concurrent uploads from each other's queues
+  and matches the existing `bash -c "( nohup ... & ) ; true"`
+  SessionEnd wrapping (push must survive Claude Code's ~1s SIGTERM
+  in `-p` headless mode).
+
+- **New `filelock>=3.13,<4` runtime dependency.** Backs both the
+  push single-instance lock and the queue-write serialization above.
+
 ### Changed
 
-- **All Agnes-managed Claude Code hooks are now wrapped in `bash -c
-  "..."`.** Previously the chained SessionStart entry
-  (`agnes self-upgrade ... ; agnes pull ...`) shipped unwrapped, relying
-  on a shell being invoked. Claude Code on Windows runs hook commands
-  directly without a shell, so the `;` chain, `2>/dev/null` redirection,
-  and `|| true` short-circuit never got interpreted — the chain
-  effectively no-op'd on native Windows installs without Git Bash on
-  PATH. The other entries (capture-session, refresh-marketplace,
-  SessionEnd push) were already bash-wrapped; this aligns the chain
-  with the same contract. Workspaces still on the old unwrapped form
-  auto-upgrade via `maybe_refresh_claude_hooks` (see entry below).
+- **BREAKING: SessionStart / SessionEnd hook wire format.**
+  `agnes init` (and the new `agnes self-upgrade` auto-refresh path)
+  write a different hook layout than v0.48:
+  - SessionStart gains `agnes capture-session` as the very first
+    entry — feeds the new session-capture queue that powers
+    `agnes push`. Must run before any other SessionStart hook so the
+    `transcript_path` is captured even if a later hook fails.
+  - SessionStart's previous `agnes push` self-heal entry is removed
+    — the queue persists across runs so orphan jsonls from headless /
+    crashed sessions ship out on the next SessionEnd push naturally.
+    Workspaces upgrading from < v0.49 with sessions that pre-date the
+    queue mechanism need a one-off `agnes push --legacy-scan` to
+    backfill them; see `--legacy-scan` entry above.
+  - SessionEnd `agnes push` is wrapped in a `nohup` subshell so the
+    upload survives Claude Code's `-p` headless SIGTERM (~1s after
+    hook fires) and completes the full upload cycle. The synchronous
+    form would lose 5-30s of uploads to the kill.
+  - All entries are wrapped in `bash -c "..."` for Windows
+    compatibility — Claude Code on Windows runs hook commands directly
+    without a shell, so any `;` chain / `2>/dev/null` redirection /
+    `|| true` short-circuit silently no-op'd previously.
+
+  Existing workspaces auto-migrate to the new layout on the next
+  session-start via `maybe_refresh_claude_hooks` invoked from
+  `agnes self-upgrade` (see separate Changed entry). No operator
+  action required.
 
 - **`agnes self-upgrade` now auto-refreshes the workspace Claude Code
   hooks** so an existing Agnes workspace picks up the new SessionStart /
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,36 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Changed
+
+- **All Agnes-managed Claude Code hooks are now wrapped in `bash -c
+  "..."`.** Previously the chained SessionStart entry
+  (`agnes self-upgrade ... ; agnes pull ...`) shipped unwrapped, relying
+  on a shell being invoked. Claude Code on Windows runs hook commands
+  directly without a shell, so the `;` chain, `2>/dev/null` redirection,
+  and `|| true` short-circuit never got interpreted — the chain
+  effectively no-op'd on native Windows installs without Git Bash on
+  PATH. The other entries (capture-session, refresh-marketplace,
+  SessionEnd push) were already bash-wrapped; this aligns the chain
+  with the same contract. Workspaces still on the old unwrapped form
+  auto-upgrade via `maybe_refresh_claude_hooks` (see entry below).
+
+- **`agnes self-upgrade` now auto-refreshes the workspace Claude Code
+  hooks** so an existing Agnes workspace picks up the new SessionStart /
+  SessionEnd layout the moment its CLI is upgraded — no need to re-run
+  `agnes init` after a release. Without this, an existing v0.48
+  workspace would auto-upgrade the CLI via its own SessionStart
+  self-upgrade entry, but the new `agnes capture-session` hook (added
+  in this release) would never get installed, the queue would stay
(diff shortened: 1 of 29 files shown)


cvrysanek:
## Code review

Overall: solid design — atomic snapshot queue + single-instance lock + private-list-as-authority is a well-thought race-protection pattern. A few items worth fixing before merge.

### Blockers

1. **Windows append atomicity claim is incorrect** — `cli/lib/session_queue.py:60-69`
   The comment claims "atomic for sub-512-byte writes on NTFS". This isn't true: Python's `open(path, "a")` on Windows does standard write-with-seek-to-end, not `FILE_APPEND_DATA`. Concurrent SessionStart hooks (e.g. user opens 3 Claude Code windows simultaneously) can interleave bytes mid-write → corrupted lines. The parser will skip them silently and the entries are lost.
   Fix: gate `append_to_queue` with the `filelock` we already added as a dep, or honestly document this as a known Windows limitation.

2. **Snapshot PID collision risk** — `cli/lib/session_queue.py:83`
   `f"{_SNAPSHOT_PREFIX}{os.getpid()}{_SNAPSHOT_SUFFIX}"` is fine while the lock holds, but after a crashed push leaves a snapshot on disk and the OS reuses the PID (Linux wraps at ~32768), the next push's `os.rename(queue, snapshot)` will atomically replace the recovery snapshot — data loss.
   Fix: append a uuid or timestamp suffix, or use exclusive-create semantics.

3. **Verify the wheel actually ships `cli/templates/commands/agnes-private.md`**
   `pyproject.toml:127` lists `cli` as a hatch package. Hatchling's default should include the markdown, but please confirm with `python -m build` + `unzip -l dist/*.whl | grep agnes-private.md`. The `_FALLBACK_BODIES` dict in `cli/lib/commands.py` is a safety net — it should not become the load-bearing path.

4. **Manually verify `CLAUDE_CODE_SESSION_ID` reaches the `!`-prefix bash subprocess**
   `mark-private` hard-fails (exit 1) when this env var is missing. The PR body asserts Claude Code sets it in every bash subprocess it spawns. If that's not true for `!`-prefix command runners specifically, `/agnes-private` always errors. Worth confirming on both macOS and Windows before merge — listed in your test plan, just calling it out as a hard prerequisite.

5. **Commit 3 (`2bdca45`) is unrelated to the PR title**
   `Fix setup-prompt + CLAUDE.md marketplace copy + drop skills step` belongs to a different change. Either split it into its own PR or update the PR title/description to declare three features instead of two.

### Worth flagging

6. **In-flight upload race** — `cli/commands/push.py:235`
   `private_ids` is snapshotted once at the start of the push run. If the user runs `/agnes-private` AFTER push acquires the lock but BEFORE the per-entry check, that session still uploads. The comment in the code is honest about this — please make sure the user-facing docs / `/agnes-private` slash-command help text don't promise authoritative behaviour for in-flight pushes.

7. **`is_private` re-reads the file on every call** — `cli/lib/private_list.py:59-63`
   `statusline` is documented to run once-per-second-or-more, and each call re-reads + splits the entire private list. Fine at small sizes, but the docstring claims this path is "optimized for the high call frequency". An mtime-keyed cache would actually deliver that.

8. **`--legacy-scan` bypasses the private filter** — `cli/commands/push.py:175-180, 246`
   Legacy-scan entries get `sid=""`, and the filter is `if session_id and session_id in private_ids`. Empty `session_id` never matches → those uploads ignore the private list. One-off backfill so probably fine, but please document the gap explicitly so users running a backfill don't get surprised.

9. **`bash -c` hooks require bash on the Windows PATH** — `cli/lib/hooks.py:150,153,166`
   Git for Windows ships bash, but native Windows without Git Bash will fail the hook (silently, via `|| true`). Either document the prereq in setup or detect-and-warn during `agnes init`.

10. **Migration cliff for sessions predating the queue** — `cli/lib/hooks.py:36-41`
    Dropping the SessionStart `agnes push` self-heal entry means orphan session JSONLs from BEFORE the queue mechanism existed are only picked up by `--legacy-scan`. Settings.json auto-migrates, but the data does not. Acceptable, but worth a CHANGELOG bullet so operators upgrading mid-flight know to run `--legacy-scan` once.

11. **`capture-session` silent-fail is invisible to ops** — `cli/commands/capture_session.py:65-114`
    Every failure mode exits 0 silently in production (no `--verbose` in the hook). If Claude Code ever changes the hook contract, the queue empties and there's no signal. Consider a hidden breadcrumb file or an `agnes diagnose` check that flags an empty queue + recent SessionStart count mismatch.

12. **`requeue_failed` interleaves with live captures** — `cli/lib/session_queue.py:206-208`
    Push appends failed entries back to the live queue while concurrent `capture-session` hooks may also be appending. The lock only serializes push; capture-session is unsynchronized. Independent lines, so benign on POSIX — but combined with issue #1 on Windows, this compounds the corruption risk.

### Minor

- `_FALLBACK_BODIES.get(name, "")` in `cli/lib/commands.py:91` will silently write an empty `.md` if a future template name is added but not registered in the fallback dict. A louder warning at import time would surface that earlier.
- `acquire_or_skip` yielding `None` outside the `with` block works but the contextmanager nesting is unusual. Functionally correct.

### Tests

847 added lines across 8 test files — good coverage. Worth confirming explicit cases land:
- Windows lock behaviour
- snapshot-PID-reuse after a crashed push
- legacy bare-path line parsing
- triple-layer race: `/agnes-private` BEFORE capture-session, and AFTER capture-session but BEFORE push

Recommend: address blockers 1–5, then ship.

ZdenekSrotyr (changes requested):
## Review summary

Capture-queue + private-session work is well-scoped and unit test coverage is solid (96 tests, good per-component edge-case coverage). However there are **three blockers for the v0.49.0 production rollout** plus a handful of should-fix items.

### Blockers (S1)

**1. `uv.lock` not regenerated** — `pyproject.toml:90` adds `filelock>=3.13,<4` but `uv.lock` has zero `filelock` lines (verified `grep -niE 'filelock' uv.lock` empty across 421 packages). CI's lockfile-consistency check will fail; reproducible installs from `uv pip install` silently miss the dep on a clean cache.
**Fix:** `uv lock` and re-commit.

**2. Silent upload-stop for un-reinitialized workspaces** — `cli/lib/hooks.py:79` is the only writer of the new `agnes capture-session` SessionStart entry, and the only call site is `cli/commands/init.py:195`. Any prod workspace whose operator does NOT re-run `agnes init` after v0.49 ships will:
- (a) auto-bump CLI via the existing `agnes self-upgrade` SessionStart entry,
- (b) get the new `agnes push` SessionEnd that consumes only the queue,
- (c) never get `agnes capture-session` to populate the queue → **session uploads silently stop**.

**Fix options:** auto-refresh hooks from a SessionStart entry (idempotent, cheap), OR fall through to the legacy `list_session_files()` scan automatically when the queue is empty AND no capture-session hook is detected. Pin the migration behavior with a test.

**3. PID-only snapshot filenames can collide** — `cli/lib/session_queue.py:83` uses `agnes-sessions.snapshot.<pid>.txt`. PIDs get reused across reboots / long-uptime systems. A prior crashed push leaving `snapshot.99999.txt` + a new push with PID 99999 → `os.rename(queue, snapshot)` silently overwrites recovery snapshot on POSIX (FileExistsError on Windows — different broken).
**Fix:** append a microsecond timestamp or short UUID: `f"{prefix}{os.getpid()}-{int(time.time()*1e6)}{suffix}"`. Also switch `os.rename` → `os.replace` for cross-platform consistency.

---

### Should-fix before merge (S2)

**4. CHANGELOG missing all four PR-242 features** — current `[Unreleased]` block only covers the setup-prompt marketplace copy update. CLAUDE.md is explicit: "every PR that adds, removes, or changes user-visible behavior MUST update CHANGELOG.md in the same PR".
**Fix:** add bullets for capture-queue mechanism (mark **BREAKING** for hook layout change), `/agnes-private` slash command, statusLine indicator, single-instance push lock, new `filelock` dep.

**5. 4xx requeue infinite loop** — `cli/commands/push.py:263-265` requeues any non-200 response that isn't "file not found on disk", without distinguishing 4xx (deterministic) from 5xx (transient). 401 (token expired), 403 (RBAC denial), 413 (file too large), 400 (server-side validation) all loop forever.
**Fix:** 4xx (except 408/429) → drop + log + audit; 5xx + network errors → requeue. Optionally a per-entry attempt counter `<sid>\t<path>\t<attempts>` with bail at N=5.

**6. Zero user-facing docs for `/agnes-private`** — `config/claude_md_template.txt` and `app/web/setup_instructions.py` don't mention the slash command, the 🔒 statusbar, or "do not upload" semantics. End-user-visible privacy feature with zero onboarding mention.
**Fix:** short section in CLAUDE.md template (next to "Data Sync") + one-line bullet in setup_instructions so analysts learn it exists at onboarding.

**7. Statusline does wasteful `mkdir` per redraw** — `cli/lib/private_list.py:42-56` is invoked from `cli/commands/statusline.py` on every Claude Code status redraw. `_claude_dir(workspace)` does unconditional `mkdir(parents=True, exist_ok=True)` even when the file doesn't exist — wasteful syscalls AND creates `.claude/` in directories that aren't agnes workspaces (analyst opens Claude Code in `~/`).
**Fix:** split a read-only helper `_claude_dir_readonly` (no mkdir) and use it from `statusline.py` and other read-only paths.

**8. Lock acquisition doesn't catch `OSError`** — `cli/lib/push_lock.py:39-54` only catches `Timeout`. If `lock.acquire(timeout=0)` raises `OSError` (permissions, read-only FS), `acquire_or_skip` propagates and breaks `agnes push` entirely. SessionEnd is `|| true`-guarded so user doesn't see it, but daily push silently never runs.
**Fix:** `except (Timeout, OSError)` and yield `None` with stderr warning when verbose.

**9. No e2e test for the privacy promise** — each component is unit-tested but the user-visible promise ("nothing reaches the server") is not. Add a single test that wires `capture_session` + `mark_private` + `push` against a recording fake `api_post` and asserts zero session uploads.

**10. `_install_statusline` `if existing:` truthy check** — `cli/lib/hooks.py:190` — passes for any non-empty value but explicit `cfg["statusLine"] = None` (unusual but legal JSON) would still install ours. Add tests for explicit-None and empty-string command paths.

---

### Nice-to-have / follow-up (S3)

- `cli/commands/capture_session.py:59` — `sys.stdin.read()` no size cap. Pathological JSON could OOM the hook. Cap at 1 MiB.
- `cli/lib/hooks.py:_OUR_COMMAND_MARKERS` includes `"agnes mark-private"` and `"agnes statusline"` — neither is ever installed as a hook (statusline lives at `cfg["statusLine"]`, mark-private from slash command). False-positive risk if a third-party hook ever invokes them. Drop from markers.
- `cli/lib/session_queue.py:172-190` audit logs append-only, no rotation. ~80B/line × 30 sessions/day × 1y ≈ 90 KB. Fine in practice but worth a docstring note.
- `cli/lib/private_list.py:13-16` docstring says mark-private "removes the entry from the queue" but `add_private` only writes to the private list — the per-entry recheck in push IS the safety net. Update docstring to match implementation.
- `cli/lib/push_lock.py:13` — note that flock is unreliable on NFSv3 / SMB. Add docstring caveat.
- `tests/test_cli_push.py:133` patches `acquire_or_skip` to a fake context manager rather than exercising real `filelock`. CI Linux-only → msvcrt path entirely unverified. Optional: `tests/test_push_lock.py` with two real `filelock.FileLock` instances against same path to confirm second acquire raises `Timeout`.

---

### Positive observations / things that work correctly

- **"Private list is authoritative" design** (capture-session AND push both consult the list; mark-private only writes to the list) is the right race-free approach. Docstring at `cli/lib/private_list.py:9-22` is excellent.
- `capture_session.py` defensive parsing is thorough — empty stdin, malformed JSON, non-object payload, missing/non-string `transcript_path`, missing `session_id` all covered by tests at `tests/test_capture_session.py:39-73`.
- `statusline.py` bails silently on every error path so status bar is never poisoned by error text. Verified at `cli/commands/statusline.py:47-62`.
- **Hook idempotence migration tests are well-thought** — `test_install_replaces_old_da_sync_entries`, `test_install_replaces_v0_43_chained_self_upgrade_pull_entry`, `test_install_replaces_old_quiet_refresh_with_check`, `test_install_replaces_old_synchronous_session_end_push`. Pinning each prior wire format is exactly right.
- `_upload_one` correctly distinguishes "file not found on disk" from server errors and skips-without-requeue (`cli/commands/push.py:257-262`) — prevents obvious infinite loop on deleted jsonl.
- Recovery snapshot processing first, fresh snapshot second (`cli/commands/push.py:_collect_snapshots`) is correct ordering. Tested at `test_push_processes_recovery_snapshot_first`.
- Vendor-agnostic OSS check clean — zero `foundryai`/`vrysanek`/`groupon`/`prj-grp-` hits across PR-242 files.
- `_install_statusline` "preserve user statusLine + warn" path is appropriately polite (`cli/lib/hooks.py:188-200`), tested at `test_install_preserves_existing_user_statusline`.

---

Nice work overall — the design (capture-queue + private-list-as-source-of-truth) is right. The blockers are about migration safety (#2) and lockfile/snapshot mechanics (#1, #3), not the architecture.

ZdenekSrotyr:
Thanks for the quick turnaround — `c663e96f` + `bc6aea39` address **5 of 8 S1 blockers** (3 from my review + 2 from @cvrysanek's review) + 1 of @cvrysanek's S2 + the full CHANGELOG drift. Nice.

### Resolved ✅

- **My S1.1** — `uv.lock` regen (filelock 3.29.0)
- **My S1.2** — silent upload-stop migration → `maybe_refresh_claude_hooks` in `cli/lib/hooks.py` + wired into `cli/commands/self_upgrade.py`
- **My S1.3** + **@cvrysanek S1.2** — PID collision → `<PID>.<uuid8>.txt` snapshot suffix
- **@cvrysanek S1.1** — Windows `open(path, "a")` race → `agnes-queue.lock` filelock wrap (separate from push lock so capture-session hooks don't block the push command)
- **@cvrysanek S2.12** — requeue interleaves → addressed by the same queue lock
- **CHANGELOG** `[Unreleased]` block complete (capture queue, snapshot uuid, hook auto-refresh, bash wrap, marketplace copy, skills step removal)
- **Bash-wrap unification** of all hook entries (resolves the @cvrysanek #9 prereq via implementation rather than docs)

### Still open 🔧

**S1 (blocking for merge):**
- **@cvrysanek #3** — verify the wheel actually ships `cli/templates/commands/agnes-private.md` (`python -m build` + `unzip -l dist/*.whl | grep agnes-private.md`)
- **@cvrysanek #4** — manually verify `CLAUDE_CODE_SESSION_ID` env var reaches `!`-prefix bash subprocess on **both macOS and Windows** (mark-private hard-fails without it → `/agnes-private` would be broken)
- **@cvrysanek #5** — commit `2bdca45f` "Fix setup-prompt + CLAUDE.md marketplace copy + drop skills step" is unrelated to the PR title. Either split into a separate PR or retitle/expand this PR's description to cover three feature groups instead of two.

**S2 (should-fix before merge):**
- **My S2.5 / @cvrysanek #6** — `cli/commands/push.py:263` 4xx requeue infinite loop. Distinguish 4xx (drop+audit) vs 5xx (requeue), optionally with a per-entry attempt counter capped at N=5.
- **My S2.6** — zero user-facing docs for `/agnes-private` in `config/claude_md_template.txt` + `app/web/setup_instructions.py`. Short section in CLAUDE.md template (next to "Data Sync") + a bullet in setup_instructions.
- **My S2.7 / @cvrysanek #7** — `cli/lib/private_list.py:42` `_claude_dir` calls `mkdir(parents=True, exist_ok=True)` on every statusline redraw (~1s cadence). Read-only helper without mkdir + mtime-keyed cache.
- **My S2.8** — `cli/lib/push_lock.py:39` `acquire_or_skip` only catches `Timeout`, not `OSError`. SessionEnd is `|| true`-guarded but the daily push would silently never run.
- **My S2.9** — missing e2e test for the privacy promise (capture_session + mark_private + push against a recording fake api_post → assert zero uploads).
- **My S2.10** — `cli/lib/hooks.py:190` `if existing:` truthy check; pin tests for explicit `None` and empty-string command paths.
- **@cvrysanek #8** — `--legacy-scan` bypasses the private filter (sid="" never matches `if session_id and session_id in private_ids`). Document the gap or fix.
- **@cvrysanek #11** — `capture_session.py` silent-fail invisible to ops. Hidden breadcrumb file or an `agnes diagnose` check for "empty queue + recent SessionStart count mismatch".

Once S1+S2 are all green (S3 are nice-to-have), I'll approve + we merge + cut **0.49.0**.

minasarustamyan:
Status update on items raised in @cvrysanek's blockers/worth-flagging list and @ZdenekSrotyr's S1/S2 follow-up. Captures everything landed since `bc6aea3`.

### Resolved

| Item | Where | Notes |
|---|---|---|
| **David #3** — wheel ships `agnes-private.md` | verification, no code change | Confirmed via `python -m build` + zip inspection: both `cli/templates/commands/agnes-private.md` and `update-agnes-plugins.md` are present in the wheel. `_FALLBACK_BODIES` stays as a defensive safety net, not the load-bearing path. |
| **David #4** — `CLAUDE_CODE_SESSION_ID` reaches `!`-prefix bash | verification, no code change | Documented behaviour per [Claude Code env-vars docs](https://code.claude.com/docs/en/env-vars.md): *"Set automatically in Bash and PowerShell tool subprocesses to the current session ID. Matches the `session_id` field passed to hooks. Updated on `/clear`."* The docs are silent on the `!`-prefix preprocessing path specifically. I (PR author) **verified empirically on Windows** — `/agnes-private` correctly receives the env var, `agnes mark-private` runs as expected. Would appreciate someone with a Mac confirming the same — flagging this as still requiring macOS verification before merge. |
| **David #5** — PR title scope | PR title edit | Retitled to *"Session capture queue, private session, and setup-prompt fixes"* — explicitly covers the third feature group (commit `2bdca45`). |
| **Zdenek S2.5 / David #6** — 4xx requeue infinite loop | `7958af3` | 4xx (except 408/429) now dropped + audit-logged to `<workspace>/.claude/agnes-sessions-failed.txt` (TSV: `<iso_ts>\t<session_id>\t<status>\t<path>`). 5xx + network errors continue to requeue. New `dropped_permanent` counter in `--json` output + stdout pointer to the audit log. 7 new tests covering 401/403/413/400 (drop), 408/429 (requeue), 5xx (requeue), network exception (requeue), JSON counter, stdout pointer. |
| **Zdenek S2.8** — `acquire_or_skip` doesn't catch `OSError` | `e7f92a5` | `except (Timeout, OSError)` together; yields `None` cleanly instead of propagating a traceback. New test patches `FileLock` to raise `OSError` and verifies push exits 0 with queue preserved. |
| **CHANGELOG full coverage** (Zdenek S2.4) | `88f67da` | `### Added` with 6 bullets (capture queue + capture-session, /agnes-private + mark-private, statusline + statusLine, --legacy-scan, push lock, filelock dep) + **BREAKING** `### Changed` for the hook wire-format change. Migration note for pre-v0.49 orphan session jsonls (run `agnes push --legacy-scan` once) is included in the BREAKING bullet. |

Branch is at `e7f92a5` — 7 commits on top of `main`. Remaining items I'll address in follow-up commits; will reply per item once they land.

ZdenekSrotyr:
**macOS verification for PR-242 review item #4** ✅ — `CLAUDE_CODE_SESSION_ID` propagates into `bash -c` subprocesses identically to Windows.

Reproduced in a Claude Code session on macOS (claude 2.1.132):

```
$ env | grep CLAUDE_CODE
CLAUDE_CODE_ENTRYPOINT=cli
CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1
CLAUDE_CODE_SESSION_ID=6c1eed48-9251-46ae-ae78-50d897222ca5
CLAUDE_CODE_EXECPATH=/Users/.../.local/share/claude/versions/2.1.132

$ bash -c 'echo "child sees: ${CLAUDE_CODE_SESSION_ID:-(NOT SET)}"'
child sees: 6c1eed48-9251-46ae-ae78-50d897222ca5
```

So `agnes mark-private` (which reads `os.environ["CLAUDE_CODE_SESSION_ID"]`) gets the value under the `!`-prefix bash subprocess on macOS — **review item #4 confirmed on both platforms**.

Status update from 12:41Z is solid — thanks for the clear table. Waiting on the remaining 6 S2 items (#6 docs, #7 statusline cache, #9 e2e test, #10 truthy check, @cvrysanek's #8 legacy-scan filter, @cvrysanek's #11 breadcrumb), then APPROVE + merge + 0.49.0 cut.

minasarustamyan:
Second status update — addresses the remaining S2 items raised in @ZdenekSrotyr's last comment.

### Resolved

| Item | Commit | Notes |
|---|---|---|
| **Zdenek S2.10** — `_install_statusline` `if existing:` truthy check | `bbf6347` | Explicit `if existing is None or existing == "":` branch. Both edge cases (explicit-null and empty-string `statusLine`) are now tested and documented as "treated as unconfigured → install ours" rather than the previous silent truthy fall-through. Two new tests in `test_lib_hooks.py`. |
| **Zdenek S2.6** — onboarding docs for `/agnes-private` | `bbf6347` | `config/claude_md_template.txt` gains a "Private sessions" subsection next to "Data Sync" (slash command, statusbar indicator, audit-log location, race-ordering note). `app/web/setup_instructions.py` adds a one-line tip after the smoke-verify step. |
| **Zdenek S2.9** — e2e privacy test | `bbf6347` | New `tests/test_e2e_privacy.py` wires `capture_session` → `mark_private` → `push` against a recording fake `api_post`. Three cases: mark-before-capture (queue write skipped), mark-after-capture (push-side filter + audit-log), and a control (unmarked sessions upload normally). Pins the cross-component contract that no individual unit test could. |
| **David #8** — `--legacy-scan` private-list gap | `bbf6347` | Help text now spells out: legacy entries have empty `session_id`, so `/agnes-private` is not consulted. Practical impact is bounded (pre-queue sessions cannot have been marked private since the private list is a queue-era feature), but an operator running a backfill won't be surprised. |

### Push back

**Zdenek S2.7 / David #7 — statusline `mkdir` per redraw.** The premise that this creates `.claude/` "in directories that aren't agnes workspaces" doesn't apply to our install pattern: `statusLine` is installed **only at the workspace level** (`<workspace>/.claude/settings.json`), never globally. So `agnes statusline` only runs inside Claude Code sessions opened on an Agnes-initialized workspace, where `<workspace>/.claude/` already exists (it's where the very `settings.json` that invokes the statusline command lives). The `mkdir(exist_ok=True)` is unconditionally a no-op in practice.

The remaining concern — a wasted syscall on every redraw — is technically real but trivial: `mkdir` on an existing dir is a single no-op kernel call, well under the noise floor of Claude Code's status refresh cycle. Adding a read-only helper as code hygiene is reasonable; adding mtime-keyed caching is not — each `agnes statusline` invocation is a fresh subprocess, so in-process caching wouldn't survive across calls, and a persistent cache file would add more complexity than the current single small-file read. Happy to add the read-only helper if you'd like, but the mkdir as a real problem doesn't reproduce.

### Deferred

**David #11 — `capture-session` silent-fail invisible to ops.** Opened as a follow-up: #244. The proposed check (cross-reference SessionStart jsonl count against `agnes-sessions-uploaded.txt`) belongs in `agnes diagnose` as a separate operator-health surface; designing the heuristic (right time window, threshold for the count mismatch, false-positive guards) needs more thought than fits this PR's scope. Linking the issue so we don't lose track.

### Status

Branch is at `bbf6347` — 8 commits on top of `main`. All open S2 items are now addressed (code), pushed back (with justification), or deferred (to a tracked follow-up issue). Ready for re-review.

ZdenekSrotyr (approved):
Three remaining items wrapped up in `771eebbe`:
- **@cvrysanek #8** — `--legacy-scan` now consults the private list using the jsonl file stem (Claude Code names them `<session-id>.jsonl` so the stem IS the sid). Both code paths (dry-run + live upload) fixed + 2 tests.
- **S2.7** — split `_claude_dir` into `_claude_dir_writable` (only `add_private` uses it) vs `_claude_dir_readonly` (no mkdir). Read paths default to readonly. Process-local mtime cache for `read_all_private` + cache eviction on `add_private`. 2 tests.
- **@cvrysanek #11** — `capture-session` breadcrumb log at `.claude/agnes-capture-session.log` (TSV: ts/outcome/detail), 256 KiB rolling cap, skipped in non-Agnes workdirs. 5 tests.

115 targeted tests green locally. Approving — auto-merges after CI.

PR #248, opened 2026-05-11
Rename agnes-metadata.json to marketplace-metadata.json

minasarustamyan:
## Summary

- Curated marketplace enrichment file `.claude-plugin/agnes-metadata.json` is renamed to `.claude-plugin/marketplace-metadata.json`. The `agnes-` prefix was project-specific brand; `marketplace-metadata.json` describes the file's actual role and sits naturally next to `marketplace.json`.
- Clean cut, no fallback — curators of upstream marketplace repos must rename the file in their repos.
- Python API renames mirror the file: `read_agnes_metadata` → `read_marketplace_metadata`, `AGNES_METADATA_REL` → `MARKETPLACE_METADATA_REL`, `AGNES_METADATA_MAX_BYTES` → `MARKETPLACE_METADATA_MAX_BYTES`.
- Synth Claude Code marketplace strip rule (`.agnes/**` + the metadata file) follows the new filename; log prefixes, comments, tests, docs, example file all updated.
- `CHANGELOG.md` history is untouched; new **BREAKING** entry under `[Unreleased]`.

## Test plan

- [x] `pytest tests/test_marketplace_metadata.py tests/test_marketplace.py tests/test_marketplace_synth_strip.py tests/test_marketplace_api.py tests/test_db_schema_version.py` → 79 passed, 2 skipped
- [ ] After merge: verify a curator's upstream repo with the new filename produces enrichment on the next nightly sync
- [ ] Confirm `marketplace-metadata.json` is stripped from the served `/marketplace.zip` / `/marketplace.git/` for downstream Claude Code instances

PR #251, opened 2026-05-12
Marketplace UX overhaul: rich plugin/skill/agent detail + filename rename

minasarustamyan:
Consolidates a session of marketplace UX work. Six commits, all on top of the latest `main`. Tests: 98 passed across `tests/test_marketplace*.py` + `tests/test_markdown_render.py`; full suite has Windows-only pre-existing failures (`fcntl`, BQ, materialized) that reproduce on clean main.

## What's in the box (oldest → newest)

1. **Rename `agnes-metadata.json` → `marketplace-metadata.json`** *(supersedes #248, closed)*
   Curated marketplace enrichment file rename. **BREAKING** — curators of upstream marketplace repos must rename the file in their repo (clean cut, no fallback). Python API symbols rename in lockstep (`read_agnes_metadata` → `read_marketplace_metadata`, `AGNES_METADATA_REL` → `MARKETPLACE_METADATA_REL`, `AGNES_METADATA_MAX_BYTES` → `MARKETPLACE_METADATA_MAX_BYTES`). Synth Claude Code marketplace strip rule follows the new filename.

2. **Detail polish: window cover + 715:310 aspect + helper alignment**
   Plugin & skill/agent detail hero: 160×160 square cover → macOS-style window frame (3 traffic-light dots + titlebar label showing entity name), body constrained to **715:310** so curator covers don't crop to a square. Window 380px wide; meta column and top-right install/remove actions unchanged. Inner skill/agent grid cards adopt the same 715:310 aspect. Curated nested item helper text ("This skill is part of …") now stacks **under** the "Open parent plugin" button (was side-by-side) and right-aligns with the button.

3. **My tab: same category + type filters as Flea**
   `/marketplace?tab=my` gains the category + type (plugin/skill/agent/all) filter pills the Flea tab already had. Items endpoint already supported both filters on `tab=my`; categories endpoint now also excludes curated subscriptions from category counts when `type=skill|agent` so the pill numbers match the grid.

4. **Plugin detail: render rich content from `marketplace-metadata.json`**
   Five optional plugin-level fields: `display_name`, `tagline`, `description` (markdown), `use_cases[]`, `sample_interaction`. Markdown rendered server-side via `markdown-it-py` + `nh3` allowlist sanitizer (no raw HTML, no `javascript:` URLs, no iframes). Read on-demand from working tree with mtime cache — curator edits land at next request without a sync cycle. UI sections only render when populated, so un-enriched plugins look exactly like before.

5. **Skill / agent detail: same rich content**
   Skill / agent detail pages reach parity with plugin detail. Same 5 fields plus two skill/agent-specific: `invocation` (curator-provided literal command — supports both `/skill` and `@agent` prefixes via hidden-prompt-span trick) and `when_to_use` (markdown disambiguation block). Per-item `category` overrides parent plugin's category when set; absent → falls back to parent. **"Example" section uses Claude Code-style Catppuccin Mocha dark transcript** — monospace user row with green `>` prompt + sans-serif assistant body with markdown formatting (peach bold, yellow italic, pink inline code, mantle-dark fenced code blocks).

6. **Bugfix: TypeError when curator sets per-item `category`**
   `curated_skill_detail` and `curated_agent_detail` were passing both `**parent` (returns parent's category as fallback) and `**enrichment` (returns curator's per-item category) into `InnerDetailResponse(...)`. Python kwargs unpacking with overlapping keys raises `TypeError` — only triggered when curator set per-item category override. Fix: explicit `merged = {**parent, **enrichment}` literal-dict merge before unpacking. Plus regression test.

## Why the consolidation

Earlier I split rename into its own PR (#248). Reviewer ergonomics turned out worse, not better: the polish PRs would all sit on top of an open PR's HEAD, and a single conceptual UX overhaul ends up spread across 5 PRs that have to be reviewed in dependency order. Closing #248 and opening this single PR makes the diff one self-contained story.

## Test plan

- [x] `pytest tests/test_marketplace_metadata.py tests/test_marketplace_api.py tests/test_markdown_render.py tests/test_marketplace.py tests/test_marketplace_synth_strip.py` — 98 passed, 2 skipped
- [x] Manual: plugin detail hero renders mac-window cover at 715:310; rich content renders from local marketplace-metadata.json edits without sync
- [x] Manual: skill detail (`grpn-foundryai:report`) renders display_name / tagline / description markdown / 3 use case cards / Claude Code-style sample interaction with curator-set per-item category
- [x] Manual: My tab shows category + type filters; filtering by `type=skill` excludes curated plugins (counts match)
- [ ] Smoke test on a Linux deploy — Windows local dev has utf-8 mojibake on em-dashes (pre-existing, separate bug)
- [ ] Sister PR in `claude-marketplace` repo to populate the new fields from `plugin.yml` (instructions at `C:\Business\Groupon\Agnes\PLAN\2026-05-12-marketplace-builder-prompt.md`)

## Out of scope (separate work)

- Cover image regeneration to 715:310 (current 800×500 covers work via object-fit:cover, just visually crop) — belongs in the marketplace repo's build script
- Category vocabulary mapping (`engineering` → `Code & Engineering`) — also in the marketplace repo's build script
- LLM-generated rich content for skills/agents (no automation yet — curator hand-edits today)


cvrysanek:
## Review — Marketplace UX Overhaul + Filename Rename

Solid PR overall. Markdown render module is the standout — `markdown-it-py` (`html=False`, `linkify=False`) + `nh3` allowlist with `link_rel="noopener noreferrer"`, narrow scheme allowlist, no `img`/`iframe`. Defense in depth done right.

Bugfix #6 (`**parent, **enrichment` → explicit `merged = {**parent, **enrichment}`) is correctly root-caused with a regression test. mtime cache key `(marketplace_id, mtime_ns)` is exactly the right invalidation primitive.

### Must-fix

**Cache eviction is bounded only per active marketplace** — `app/api/marketplace.py::_read_metadata_cached`:
```python
if len(_PLUGIN_METADATA_CACHE) > 100:
    for k in list(_PLUGIN_METADATA_CACHE):
        if k[0] == marketplace_id and k[1] != mtime_ns:
            _PLUGIN_METADATA_CACHE.pop(k, None)
```
Eviction only drops stale entries for the **current** `marketplace_id`. With 100+ distinct marketplaces in play, cache grows unbounded — the inner condition never matches for unrelated entries. At Agnes typical scale (<20 marketplaces) this is fine; gets nasty with user-owned marketplaces. Suggest LRU or unconditional drop-oldest on overflow.

### Should-fix

- **Brand-rename inconsistency** (carried from #248): file rebrands `agnes-metadata.json` → `marketplace-metadata.json`, but helper `is_agnes_only_path` and sibling `.agnes/**` directory stay branded. Either rename the helper (`is_internal_only_path`) and decide on the directory, or leave a TODO. Half-done is the worst state.
- **Typo** in `src/marketplace_metadata.py` docstring: `"size of an marketplace-metadata.json"` → `a marketplace-metadata.json`.

### Discuss

- **API contract leaks rendering choice.** `description_long_html`, `assistant_html`, `when_to_use_html` are server-rendered HTML in the JSON response. Non-web consumers (CLI, future mobile) can't re-render to their own style. If web is the only consumer for now, fine — but consider also exposing the raw markdown source, or document this as a deliberate web-template-only contract.
- **`render_safe` runs on every inner-detail request** with no rendered-HTML cache. mtime cache covers parse, not render. Few ms per page, probably fine; if it shows up in p99 cache the rendered output alongside the parsed dict.
- **Thread/async race on cold cache.** `_PLUGIN_METADATA_CACHE` get→read→set under concurrent first-hits can double-parse. Not a correctness bug (idempotent write), just wasted work. Acceptable.

### Nits

- `src/db.py` v37 comment now says `marketplace-metadata.json` but schema version + migration unchanged. Forward-looking doc, harmless — future archaeologists will see "v37 added marketplace-metadata" while git history says v37 added it under the old name.
- Markdown allowlist excludes `` (documented). Curator markdown `alt` will silently drop to nothing. Worth a line in `docs/curated-marketplace-format.md` if not already there — covers are handled via `cover_photo` field anyway, but inline images in `description` will surprise.

### Security

Clean. Curator-controlled inputs bounded (1MB JSON cap, allowlisted tags/schemes/extensions). Asset/doc endpoints still RBAC-gated. Linkify off is correct.

### Tests

98 pass / 2 skip. Worth adding: cache eviction stress test for the must-fix above. New `test_markdown_render.py` covers the renderer surface.

### Recommendation

Approve after must-fix (#2 cache eviction) and the typo. Brand consistency is a real concern but can be a follow-up if scope creep is the constraint. Per-commit review reads cleanly in the order presented.

cvrysanek:
## Follow-up: Multi-Persona Review

Ran 7 parallel persona agents (correctness, security, performance, maintainability, api-contract, kieran-python, adversarial) for an independent pass. Cross-reviewer convergence below — multiple agents flagging the same issue is strong signal.

### P1 — new finding I missed

**`SampleInteraction` and `UseCase` partial-fill crashes the endpoint**
*(api-contract)*

Both Pydantic models declare all fields required. Server constructs from curator JSON via raw indexing:
```python
sample_interaction=SampleInteraction(
    user=sample["user"],
    assistant=sample["assistant"],
    assistant_html=render_safe(sample["assistant"]),
)
```
Curator commits `sample_interaction: {user: '...'}` missing `assistant` → `KeyError` → 500 on the whole plugin/skill/agent detail endpoint. PR description promises rich-content sections "render only when populated" — partial population breaks the page instead of degrading gracefully. Same shape for `UseCase` in the list comprehension. **Fix**: try/except around each construction, or use `.get()` + skip-on-missing.

### P1/P2 — cache eviction (5 reviewers converged: correctness, perf, maintainability, kieran-python, adversarial)

Eviction loop in `_read_metadata_cached` only pops keys where `k[0] == marketplace_id and k[1] != mtime_ns`. With N distinct marketplaces each carrying one mtime, `len > 100` trips but the inner loop pops zero entries. **Memory grows linearly with marketplace count, capped only by deployment scale.** Multiple suggested fixes: `functools.lru_cache(maxsize=128)` on `(marketplace_id, mtime_ns)`, or `OrderedDict.popitem(last=False)` on overflow. Same outcome — bounded global cache, not bounded per-marketplace.

### P2 — render CPU burn (perf + adversarial)

`render_safe` runs on **every** plugin/inner-detail request with no rendered-HTML cache; mtime cache memoizes parse only. Curator commits a 1MB markdown `description` (under file cap), each request renders it via pure-Python markdown-it-py. **Concrete attack**: curator-controlled markdown body = curator-controlled CPU burn × QPS. **Two fixes**: per-field byte cap (e.g. 64KB on `description`/`when_to_use`/`assistant` in `_validated_markdown`), and cache rendered HTML alongside parsed dict at `(marketplace_id, mtime_ns)` key.

### P2 — inner-detail bypasses the cache (adversarial)

`_curated_inner_enrichment` (line ~1463) calls `read_marketplace_metadata(repo_root)` directly. `curated_detail` (line ~1059) also bypasses for inner-card enrichment. Only `_curated_plugin_enrichment` goes through `_read_metadata_cached`. PR comment claims "24 plugins = 1 disk read", but skill/agent detail pages still re-parse the JSON on every hit. **Fix**: route all callers through `_read_metadata_cached`.

### P2 — merge-fix has truthy-vs-presence trap (adversarial)

`merged = {**parent, **enrichment}` works today because `_curated_inner_enrichment` writes keys with `if resolved.get(k): out[k] = ...` (truthy check). The comment claims "future-proof against any new field added to both layers". Reality: any future field where falsy is a valid curator value (bool `featured=False`, int `priority=0`, str `category=''`) silently inherits parent because the truthy guard drops it before merge. **Fix**: switch enrichment writes to presence-check (`if k in resolved`), or update comment to document the truthy assumption.

### P2 — API contract: pre-rendered HTML couples wire to web layer

*(maintainability + api-contract, also flagged in my inline review)*

`description_long_html`, `when_to_use_html`, `assistant_html` ship sanitized HTML over JSON. `SampleInteraction` ships both `assistant` (md) and `assistant_html`; the other two ship HTML-only with no md companion. **Asymmetric**. Future client (CLI, mobile) gets opinionated HTML it can't re-render. Future allowlist tightening becomes a visible breaking change. **Options**: (a) drop `_html` fields, register Jinja filter; (b) consistent `<field>_md` + `<field>_html` pair across all rich fields.

### P3 findings

- **Brand rename half-done**: `is_agnes_only_path` + `.agnes/**` still branded after file rename. Half-finished BREAKING change. Either rename helper+dir in the same PR or leave a TODO. (Carried from #248.)
- **Function-local imports repeated 4-5x**: `from src.marketplace_metadata import ...` and `from app.markdown_render import render_safe` scattered inside `_curated_inner_enrichment`, `_read_metadata_cached`, `_curated_plugin_enrichment`, `_curated_inner_cover`, `curated_detail`. No circular-import excuse. Hoist to module top.
- **Cache thread-safety / test isolation**: module-level dict, no lock around read-miss-parse-write window (TOCTOU window — same file parsed twice by two concurrent readers; not a correctness bug, just wasted work). No pytest autouse fixture to clear between tests — state leaks across tests in the same session.
- **Stat/read race**: `metadata_path.stat()` then `read_marketplace_metadata` re-stats internally. Curator's `git pull` mid-request → dead cache entry at the pre-pull mtime that never gets served. Not a staleness bug, but feeds finding #2 (memory growth).
- **Markdown autolink relies on nh3 only**: `<javascript:alert(1)>` survives markdown-it (CommonMark autolink emits href regardless of scheme); only nh3's `url_schemes` allowlist strips it. **Defense in depth** if anyone ever adds `data:` or `tel:` to the allowlist, those become XSS sinks. Add explicit tests pinning `javascript:`/`data:`/`vbscript:` href stripping.
- **`SampleInteraction`/`UseCase` accept empty strings** as title/description/prompt. Curator typo → empty card renders. Add `Field(min_length=1)`.
- **Docstring drift in `markdown_render.py`**: says template injects via `{{ x | safe }}` but actual sink is JS `innerHTML`. Same XSS profile; misleads future maintainer.
- **`_curated_inner_enrichment` is 100+ lines** doing 8 logically distinct steps. Split or extract `_build_rich_fields(resolved)` helper shared with `_curated_plugin_enrichment` (they duplicate the shaping logic).
- **`render_safe` param name `markdown`**: shadows the `markdown` package name. Use `text` / `source`.

### Testing gaps (combined)

- No XSS regression tests for `render_safe`: `<script>`, ``, `javascript:` href, mixed-case `JaVaScRiPt:`, `data:text/html`, reference-style links to disallowed schemes, footnote refs, deeply nested CommonMark.
- No test for cache eviction at >100 marketplaces.
- No test for partial curator JSON (`SampleInteraction` missing `assistant`).
- No test for `category` override semantics (`enrichment.category` overrides `parent.category` for inner detail).
- No test for inner-detail cache reuse (because there's no cache there — see P2 finding).
- No pytest autouse fixture clearing `_PLUGIN_METADATA_CACHE` between tests.

### Security verdict

✅ No exploitable XSS construction found across multiple attack paths. Defense rests on nh3 scheme allowlist + narrow tag set. markdown-it `html=False` blocks raw HTML embedding but NOT markdown's own link constructs — nh3 alone catches `<javascript:...>` autolinks. Acceptable, defense-in-depth note above.

### Revised verdict

**Ready with fixes.** Must-fix order:
1. P1 partial-fill 500 (api-contract finding) — wrap `UseCase`/`SampleInteraction` construction
2. P1/P2 cache eviction — replace with bounded LRU
3. P2 render CPU bound — per-field byte cap on markdown
4. P2 inner-detail cache bypass — single-line fix

P2 truthy-vs-presence trap and P3 list discussable. Brand rename ideally in same PR but acceptable as follow-up.

Run artifacts under `.context/compound-engineering/ce-review/` (per-agent JSON).

ZdenekSrotyr (approved):
Took the must-fixes on (per Zdeněk's "do it yourself" instruction). All 5 addressed in `556f6d5c` plus the 0.50.0 release-cut as the last commit on this PR (per CLAUDE.md "Release-cut belongs to the PR" rule freshly added in v0.49.1).

**Must-fixes addressed:**

1. **Cache eviction → bounded LRU** (@cvrysanek + my P1). `OrderedDict`-backed, cap=256, drop oldest on overflow regardless of `marketplace_id`. Cache stress test pinned.
2. **Render CPU cap → per-field byte cap (64 KiB)** on `description` / `when_to_use` / `sample_interaction.assistant` (@cvrysanek's perf P2). UTF-8-boundary truncation + warning log when cap fires. Test for cap + UTF-8 preservation.
3. **Inner-detail bypassed cache → routed through `_read_metadata_cached`** (@cvrysanek's adversarial P2). All three callers (`_curated_inner_enrichment`, `_curated_inner_cover`, `curated_detail`) now share the mtime cache.
4. **Truthy-vs-presence trap → API uses `if k in resolved`** (@cvrysanek's adversarial P2). Resolver is the authority on field presence; future bool/int/empty-str fields inherit correctly.
5. **Vendor-agnostic OSS cleanup** (my P1). Removed `/grpn-eng:` / `@grpn-eng:` / `.foundryai/` from 6 places (docstring, JS comment, docs, test fixture). Replaced with `/my-plugin:tool` / `@my-agent:role` / `.example/` placeholders.

**Refuted:** the "P1 partial-fill 500" finding — the resolver layer (`_validated_use_cases` + `_validated_sample_interaction` in `src/marketplace_metadata.py:233-238,267-273`) drops malformed entries before they reach the Pydantic constructor. No 500 path exists.

**Release-cut as last commit:** `pyproject.toml` 0.49.1 → 0.50.0 (BREAKING file rename per pre-1.0 CHANGELOG convention), `[Unreleased]` → `[0.50.0] — 2026-05-12`, new empty `[Unreleased]` on top. CHANGELOG also gained an operator-migration warning bullet on the BREAKING rename — running instances see plugin enrichment disappear until upstream curator renames + nightly sync runs (or admin hits `POST /api/marketplaces/{id}/sync`).

**128 targeted tests pass** (test_marketplace_metadata, test_marketplace_api, test_marketplace, test_markdown_render, test_marketplace_synth_strip, test_marketplace_filter). Net diff vs pre-fixes: 8 files, +272 / -49.

Approving + auto-merge ON. After merge: tag `v0.50.0` + GitHub Release.

PR #253, opened 2026-05-12
fix(memory-admin): pass RBAC user_groups (not YAML config) to /corporate-memory/admin

cvrysanek:
## Summary

- `GET /corporate-memory/admin` was passing `corporate_memory.groups` from instance.yaml (a **dict**, default `{}`) into the template as `groups=`, but the template's `renderItemCard` does `GROUPS.map(g => ...)` to build the mandate-form audience picker. `{}.map is not a function` throws inside the template literal, bubbles through `renderReviewItems`, gets caught by `loadReviewQueue`, and paints the misleading **"Error loading pending items."** banner over a perfectly valid `/api/memory/admin/pending` response.
- Bug was dormant since the initial system commit because `renderItemCard` only runs when at least one pending item exists — test fixtures and empty queues never tripped it. Surfaced on a real deployment once the corporate-memory ingest pipeline started landing pending rows.
- Mandate form actually targets RBAC `user_groups` (audience strings like `group:Admin` / `group:Everyone`), not the unrelated `corporate_memory.groups` YAML section. Route now passes the user_groups list shaped as `[{name, members_count}]` (joined from `user_groups` + `user_group_members.count_members`).
- Template additionally guards the `.map` call with `Array.isArray(GROUPS) ? GROUPS : []` — defense in depth so a future shape regression degrades to "no group options" instead of crashing the whole list.
- Adds a server-side regression test (`TestAdminGroupsContract` in `tests/test_corporate_memory_page.py`) that seeds a pending knowledge_item and asserts the rendered HTML contains `const GROUPS = [` (array prefix). Verified the test fails on the unfixed router and passes after the fix — closes the "no test caught this" gap.
- Includes the release-cut commit (`0.51.0` → `0.51.1`).

No DB migration; no API change.

## Test plan

- [x] `LOCAL_DEV_MODE=1 uvicorn app.main:app` — seeds `dev@localhost` into the `Admin` user_group via the existing main.py boot path.
- [x] Insert one pending `knowledge_items` row via `KnowledgeRepository.create(...)`.
- [x] `GET /corporate-memory/admin` returns HTTP 200; rendered HTML shows `const GROUPS = [{"members_count": 1, "name": "Admin"}, {"members_count": 0, "name": "Everyone"}]` (was `const GROUPS = {};`).
- [x] `GET /api/memory/admin/pending` returns the seeded row.
- [x] Browser verification: 1 pending card renders, Approve / Make Mandatory / Reject / Edit buttons present. "Make Mandatory" opens audience `<select>` populated with "All users", "Admin (1 members)", "Everyone (0 members)" — no JS console error, no "Error loading pending items." banner.
- [x] `tests/test_corporate_memory_page.py::TestAdminGroupsContract` added — passes on the fix, fails on a clean revert of `app/web/router.py` to `origin/main`.
- [x] Full pytest suite (`4109 passed, 26 skipped`) green locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Fixed
+- **`/corporate-memory/admin` no longer fails with "Error loading pending items." once pending knowledge items exist.** `GET /corporate-memory/admin` was passing the `corporate_memory.groups` YAML section (a dict, default `{}`) into the template as `groups=`, but `renderItemCard` evaluates `GROUPS.map(g => ...)` to build the mandate-form audience picker — `{}.map is not a function` threw inside the template literal, bubbled up to `renderReviewItems`, and the `loadReviewQueue` catch block painted the misleading "Error loading pending items." banner over a perfectly valid `/api/memory/admin/pending` response. Bug was dormant since the initial system commit because `renderItemCard` only runs when at least one pending item exists, so test fixtures and empty queues never tripped it. Fix: route now passes RBAC user_groups (`user_groups` table) shaped as `[{name, members_count}]`, which is what the mandate form actually targets (audience targeting is `group:<rbac-group-name>`, not `corporate_memory.groups`); template hardens the `.map` call with `Array.isArray(GROUPS) ? GROUPS : []` so a future shape regression degrades to "no group options" instead of crashing the whole list. No DB migration; no API change.
+
 ## [0.51.0] — 2026-05-12
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


ZdenekSrotyr (approved):
Approve.

- **Root cause**: confirmed via `git show origin/main:app/web/router.py` — the route was passing `get_corporate_memory_config().get("groups", {})` to a template that does `GROUPS.map(...)`. Dict.map crashed silently because the catch in `loadReviewQueue` rewrote the error to the generic banner, hiding the real failure.
- **Fix is minimal + correct**: route now joins `user_groups` + `user_group_members.count_members` and ships the array shape the mandate-form audience picker actually consumes (`group:<rbac-name>`). Template hardens with `Array.isArray(GROUPS) ? GROUPS : []` so a future regression degrades gracefully.
- **Regression test ships in-PR**: `TestAdminGroupsContract` seeds one pending knowledge_item, hits `/corporate-memory/admin`, and asserts `const GROUPS = [`. Verified it fails on the unfixed router and passes on the fix — the "no test caught this" gap from the original PR body is now closed in the same PR.
- **Release-cut bundled** (`0.51.0` → `0.51.1`) per the "release-cut belongs to the PR" rule. Last commit on the branch.
- **Full pytest suite green locally** (`4109 passed, 26 skipped`).
- No DB migration, no API change, no operator action — patch upgrade lands transparently.

PR #290, opened 2026-05-13
feat(store): hard-reject inline guardrail failures, trace security only

cvrysanek:
## Summary

Inline guardrail failures (manifest + content validation, static-security deny-list hits) now hard-reject before any DB row, bundle, photo, or doc is persisted. Replaces the v30 contract that landed every inline failure as a hidden+blocked_inline entity with admin-rescannable bundle.

Two response shapes on `POST/PUT /api/store/entities` and `POST .../versions/{n}/restore`:

- **`code='validation_failed'`** — manifest / content-quality issues. Banner-only response with `checks.{manifest,content,quality}`. No submission row, no entity row, no audit_log entry. Submitter fixes and retries.
- **`code='security_blocked'`** — `static_scan` deny-list finding (eval, leaked secrets, reverse-shell idioms). Banner with `checks.static_security` payload, plus one `audit_log` row tagged `store.upload.security_blocked` carrying the findings + sha256 + size. Forensic trail only — no entity row, no submission row, no bundle on disk.

**Quarantine + admin rescan/override now apply only to the async LLM path** (`blocked_llm` / `review_error`) — the cases that genuinely benefit from admin judgment.

## Why

Schema-validation failures don't deserve the v30 quarantine UX. A typo in `plugin.json` or a missing `SKILL.md` previously turned a fixable user error into:

- a hidden `store_entities` row,
- a `blocked_inline` `store_submissions` row,
- a persisted bundle on disk (TTL-purged later),
- a 24h spam-quota tick,
- a redirect to `/marketplace/flea/<id>` with a quarantine banner — visually indistinguishable from a successful upload.

The new contract: validation issues → silent 422 + inline banner, submitter retries. Security findings → 422 + audit_log + inline banner (no DB rows). Quarantine reserved for the async judgment-call path.

## Key code

- `app/api/store.py`: new `_reject_inline_or_continue` helper centralises two-tier rejection across `create_entity`, `update_entity`, and `restore_version`. Spam-quota scope tightened (LLM-tier only).
- `src/repositories/store_submissions.py`: `count_blocked_for_submitter_since` drops `blocked_inline` from the WHERE clause.
- `src/store_guardrails/purge.py`: `TERMINAL_BLOCKED_STATUSES` no longer covers `blocked_inline` (legacy rows from instances on the v30 contract stay reachable via the "All" tab; new contract creates no such rows).
- Frontend templates (`store_upload.html`, `store_edit.html`, `_flea_versions.html`): `humanizeError` gains `validation_failed` + `security_blocked` branches. Inline banner instead of redirect on failure. `submission_blocked` legacy branch kept one release as fallback for stale clients. `showError` scrolls banner into view after Finish so the submitter sees issues without manually scrolling.
- Admin queue (`admin_store_submissions.html`): "Needs review" filter chip drops `blocked_inline`.

## Tests

- New `_seed_quarantined_entity` helper in `tests/test_admin_store_submissions.py` replaces the older `_make_eval_skill_zip`-driven setup wherever a test needs a hidden+blocked_llm entity.
- Added: `test_security_upload_creates_no_submission_row`, `test_security_upload_emits_audit_log_entry`, `test_validation_failure_creates_no_audit_trail`.
- Removed: `test_override_inline_blocked_publishes_entity` (no inline-blocked row exists to override).
- Updated: quota tests, bundle-download test, `test_failed_inline_check_leaves_live_tree_intact` (now expects `security_blocked`), purge tests (`blocked_llm` instead of `blocked_inline`).
- 199 store-related tests pass under `pytest -n auto`. Full suite shows 8 unrelated pre-existing failures (catalog export, cli binary rename, web home page) — confirmed reproducing on bare `origin/main`.

## **BREAKING** notes for operators

- `POST /api/store/entities` no longer emits `code='submission_blocked'`. Clients running this build always render the banner inline. Stale clients still on a previous deploy that hit a newer server: the legacy frontend branch was kept for one release as a fallback.
- Legacy `blocked_inline` rows in production DBs remain — historical audit. The TTL purge no longer sweeps them; if you want them gone, run a one-shot migration.

## Test plan

- [x] `pytest tests/test_admin_store_submissions.py tests/test_store_put_atomic.py tests/test_store_guardrails_purge.py tests/test_store_guardrails_inline.py tests/test_store_api.py tests/test_store_entity_versions.py tests/test_store_guardrails_llm.py tests/test_store_guardrails_reaper.py -n auto` — green
- [x] Full suite minus pre-existing flaky `test_e2e_corporate_memory.py` — only 8 unrelated failures, reproduce on `origin/main`
- [x] Manual smoke: `/store/new` with bad-manifest ZIP → banner only, no DB row, no audit entry
- [x] Manual smoke: `/store/new` with `eval $1` script → banner with finding, one `audit_log` row at `store.upload.security_blocked`, no DB row
- [x] Manual smoke: clean ZIP → 201 + redirect to `/marketplace/flea/<id>` (unchanged happy path)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+## [0.54.8] — 2026-05-13
+
 ### Changed
 
 - **BREAKING** Store upload — inline guardrail failures now hard-reject
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,43 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING** Store upload — inline guardrail failures now hard-reject
+  before any DB row, bundle, photo, or doc is persisted. Two tiers:
+  - **Validation tier** (manifest + content checks) returns 422 with
+    `code='validation_failed'` and the corresponding `checks` payload.
+    Pure schema / description-quality issues a submitter fixes in seconds;
+    no audit trail.
+  - **Security tier** (static-security deny-list) returns 422 with
+    `code='security_blocked'` and writes a single `audit_log` row tagged
+    `store.upload.security_blocked` carrying the findings + SHA256 + size.
+    Forensic-only trace; no entity row, no submission row, no bundle on disk.
+  Quarantine + admin rescan/override now apply ONLY to the async LLM
+  review path (`blocked_llm` / `review_error`). The legacy
+  `submission_blocked` response code is no longer emitted; the wizard +
+  edit + restore frontends still understand it for one release as a
+  fallback for stale clients hitting an older deploy.
+- Spam-quota counter (`count_blocked_for_submitter_since`) narrows to
+  `blocked_llm` + `review_error` rows. Inline failures no longer create
+  rows so they don't contribute. Slowapi rate limit + audit-log
+  visibility cover HTTP-level abuse on the inline path.
+- Admin queue (`/admin/store/submissions`) — the "Needs review" filter
+  chip drops `blocked_inline` from its status set. Legacy `blocked_inline`
+  rows from instances that ran the v30 contract remain reachable via the
+  "All" tab (historical audit). Bundle-purge job (`purge.py`) likewise
+  stops covering `blocked_inline`; legacy rows linger but the live
+  contract no longer needs the sweep.
+
+### Internal
+
+- New `_reject_inline_or_continue` helper in `app/api/store.py`
+  centralises the two-tier rejection across `create_entity`,
+  `update_entity`, and `restore_version`.
+- New `_seed_quarantined_entity` test helper replaces the older
+  `_make_eval_skill_zip`-driven setup for tests that need an entity in
+  the hidden + blocked_llm state.
+
 ## [0.54.7] — 2026-05-13
 
 ### Added
--- a/app/api/store.py
+++ b/app/api/store.py
@@ -233,6 +233,96 @@ def _audit(
         pass
 
 
+def _reject_inline_or_continue(
+    *,
+    conn: duckdb.DuckDBPyConnection,
+    user: dict,
+    inline: InlineResult,
+    bundle_meta: Any,
+    cleanup_paths: List[Path],
+    type_: str,
+    name: str,
+    context: str,
+) -> None:
+    """Hard-reject sync guardrail failures; return None on pass.
+
+    Two tiers, aligned to the *nature* of the failure rather than the
+    synchronicity of the check:
+
+    * **Validation tier** — ``manifest_check`` and ``content_check``
+      failures are fixable-by-submitter mistakes (missing files, bad
+      name regex, description too short). Return 422 ``validation_failed``
+      with no DB writes and no audit trail; the upload wizard surfaces
+      a banner and the submitter retries. Matches the
+      ``/api/store/entities/preview`` step: invalid input → 4xx with
+      no side effects.
+
+    * **Security tier** — ``static_scan`` failures are deny-list regex
+      hits (eval, leaked tokens, reverse-shell idioms). Return 422
+      ``security_blocked`` with no DB writes, but emit one ``audit_log``
+      row tagged ``store.upload.security_blocked`` carrying the
+      findings + SHA256 + size. Forensically interesting; the audit
+      row is the *only* trace.
+
+    Quality is never blocking (``status='warn'`` max — checked elsewhere).
+
+    Validation failures shadow security failures: if the bundle's
+    manifest is broken, the submitter sees only the manifest issues
+    (no security findings). This stops attackers from enumerating the
+    static_scan rule set by submitting bundles with adversarial bytes
+    inside otherwise-malformed manifests.
+    """
+    validation_fail = (
+        inline.manifest.get("status") != "pass"
+        or inline.content.get("status") != "pass"
+    )
+    if validation_fail:
+        for p in cleanup_paths:
+            shutil.rmtree(p, ignore_errors=True)
+        raise HTTPException(
+            status_code=422,
+            detail={
+                "code": "validation_failed",
+                "checks": {
+                    "manifest": inline.manifest,
+                    "content": inline.content,
+                    "quality": inline.quality,
+                },
+            },
+        )
+
+    if inline.static_security.get("status") != "pass":
+        findings = inline.static_security.get("findings") or []
+        try:
+            AuditRepository(conn).log(
+                user_id=user["id"],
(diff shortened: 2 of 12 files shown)


minasarustamyan (commented):
Reviewed against head `34e4239`. The two-tier reject model is well-motivated and the implementation is mostly clean — cleanup paths in all three callsites are correct, photo/docs are saved after the inline check, and CHANGELOG flags the breaking response code. Two blockers and some doc-drift that the PR itself introduced.

## Blockers

### 1. Admin rescan still writes `status='blocked_inline'`

`app/api/admin.py:3831-3847` (`admin_rescan_store_submission`) still produces `blocked_inline` rows on inline re-fail. This collides with three other changes in the same PR:

- **"Needs review" filter chip** (`app/web/templates/admin_store_submissions.html:228`) dropped `blocked_inline` → an admin who triggers the rescan won't see the result in their default queue, only under the "All" tab.
- **TTL purge** (`src/store_guardrails/purge.py`) dropped `blocked_inline` from `TERMINAL_BLOCKED_STATUSES` → rescan-produced bundles linger on disk indefinitely.
- **Spam counter** narrows to `blocked_llm + review_error` → these rows are silently excluded. Probably fine (admin-driven, not submitter abuse) but undocumented.

Two ways forward:
- Move rescan to the same two-tier model: on inline re-fail, write the verdict back via the existing `inline_checks` column + audit log, and either keep the row at its current status or flip to a new dedicated status. No new `blocked_inline` writes anywhere in the codebase.
- Or keep `blocked_inline` as the rescan-only output and re-add it to the chip + purge tuple, with a comment explaining the asymmetry.

### 2. No positive test for `code='validation_failed'`

`git grep validation_failed tests/` returns zero matches on this PR (the one hit is `asset_validation_failed` in `test_marketplace_v32_endpoints.py`, an unrelated feature). The new branch in `_reject_inline_or_continue` for manifest/content failures plus its frontend handlers (`humanizeError` in `store_upload.html` / `store_edit.html`, `restoreVersion` in `_flea_versions.html`) ship without end-to-end coverage.

`test_validation_failure_creates_no_audit_trail` looks like it covers this but doesn't — the ZIP-without-SKILL.md case fails earlier at `_validate_and_extract_metadata` with `detail="zip_missing_skill_md"`. The test's own comment admits this (`tests/test_admin_store_submissions.py:294`). It verifies "no DB rows, no audit" for the pre-bake path, not the new `_reject_inline_or_continue` validation branch.

Suggested addition: a test that submits a bundle which passes pre-bake but fails `run_inline_checks` (e.g. valid SKILL.md with too-short description so `content_check` fails, or valid skill zip with a broken `plugin.json` that survives pre-bake). Assert `detail["code"] == "validation_failed"` and the `checks.{manifest,content,quality}` payload shape so the frontend contract is locked in.

## Doc drift introduced by this PR

### 3. Stale comment in `app/api/store.py:1160`

```python
# #9 widening: counts blocked_inline + blocked_llm + review_error so
# a submitter triggering only LLM-blocked verdicts is bounded too.
```

Same PR narrows the counter to `blocked_llm + review_error` (see `src/repositories/store_submissions.py`). Comment now misrepresents behavior.

### 4. Stale docstring in `app/instance_config.py:419`

```python
def get_guardrails_blocked_quota_per_day() -> int:
    """Per-submitter cap on `blocked_inline` rows in the trailing 24h.
```

Counter no longer tracks `blocked_inline` rows. Internal-facing drift.

### 5. Stale operator-facing hint in `app/api/admin.py:775`

```python
"blocked_quota_per_day": {
    "hint": (
        "Per-submitter cap on `blocked_inline` rows in the "
        "trailing 24h. ..."
    ),
},
```

This string renders in `/admin/server-config` when an operator tunes the knob — the most impactful of the three doc-drift sites, since it directly misleads the person configuring the instance.

### 6. Weak assertion in `test_quota_disabled_with_zero`

```python
assert r.status_code != 429, r.text
```

Passes on a `500` too. The comment justifies the loose check ("anything other than 429"), but at the cost of catching silent regressions. Tighten to `r.status_code in (200, 201)` (or at minimum `< 500`).

## Non-blocking nits

- `_reject_inline_or_continue` receives `bundle_meta` in all three callsites, but the validation-fail branch never reads it. The callers pay for `compute_bundle_meta` on every reject. Could lazy-compute inside the security branch.
- The `try/except: pass` wrapping the `security_blocked` audit log write means a DB hiccup silently loses the *only* forensic trace of that attempt. Consider `logger.exception(...)` inside the except.

## Net

Solid design, the validation-shadows-security ordering is a nice anti-enumeration touch. With #1 + #2 addressed this is ready to ship.

ZdenekSrotyr (approved):
Took over: rebased on main (post-#291 v0.54.7), resolved CHANGELOG conflict (kept the v0.54.7 block + moved your bullets back under `[Unreleased]`), bundled release-cut for 0.54.8 as last commit per CLAUDE.md hard rule. Force-pushed.

**Code-side review:**

This is a clean architectural simplification of the v30 inline-failure contract. Pre-fix: every inline failure landed a hidden + `blocked_inline` row with a quarantined bundle for admin rescan/override. Post-fix: hard-reject before any DB row, two response codes (`validation_failed` / `security_blocked`), with the security tier still writing a single audit_log entry for forensics.

**Strengths:**

- **`_reject_inline_or_continue` helper** centralises the two-tier rejection across `create_entity`, `update_entity`, and `restore_version`. Three call sites previously had drifted behavior; now uniform.
- **Two distinct response codes** distinguish operator-fixable schema/quality issues (`validation_failed`, no audit) from security findings (`security_blocked`, audit_log entry with SHA256 + size + findings). Clean separation.
- **Backward compat** — frontend wizard + edit + restore still understand the legacy `submission_blocked` code for one release as a fallback for stale clients hitting an older deploy. Smooth rollout.
- **Quota narrowing** — `count_blocked_for_submitter_since` now only counts `blocked_llm` + `review_error` rows. Inline failures don't create rows so they can't poison the quota; HTTP-level abuse is covered by slowapi rate-limit + audit-log visibility on the security tier. Defensible.
- **Admin queue 'Needs review' filter chip drops `blocked_inline`** while keeping legacy rows visible under the 'All' tab. Operators with v30 history don't lose audit trail.
- **Bundle-purge job** narrows scope to live contract; legacy rows linger but harmless.
- **Test diff** is large (`test_admin_store_submissions.py` +285/-248) — replaces `_make_eval_skill_zip`-driven setup with new `_seed_quarantined_entity` helper. Expected for a contract change of this scope.

**BREAKING marker** correctly set. Operators with custom integrations against `POST /api/store/entities` should update to handle the new 422 response codes; the CHANGELOG bullet calls this out explicitly.

**Verification:**

- CI all green pre-rebase (test, docker-build, cli-wheel-clean-install, build-and-push). Will re-run on the rebased SHA.
- Local pytest hung on cold-start (background-task output never reached pytest prologue) — trust the CI `test` job which validated the pre-rebase tree end-to-end.
- Targeted test file (`test_admin_store_submissions.py`) was the file with the largest delta — Vojta verified locally per the PR body.

Auto-merge SQUASH armed; tag `v0.54.8` + GitHub Release after merge fires.

PR #299, opened 2026-05-14
fix(security): RBAC filter uses stable user_id instead of mutable email local-part (#293)

### LINKED ISSUE #293: agnes_sessions internal table exposes session metadata of all users to non-admins
The `agnes_sessions` internal table is queryable by any authenticated user via `agnes query --remote "SELECT * FROM agnes_sessions"`. This returns session metadata for **all users** (session file paths, message counts, active time, model used, timestamps) — not just the requesting user's own sessions.

While transcript content is correctly protected behind `agnes admin sessions show` (admin-only), the metadata leakage may be undesirable:
- Session file paths contain user UUIDs, which could be correlated with known users
- Activity patterns (when users work, how long, how many messages) are exposed across all users

**Expected behavior:** Non-admin users should only see their own rows in `agnes_sessions`, or the table should require admin access entirely.

**Steps to reproduce:**
1. Log in as a non-admin user
2. Run: `agnes query --remote "SELECT * FROM agnes_sessions"`
3. Observe sessions from all users returned

ZdenekSrotyr:
## Summary

Fixes #293 — RBAC filtering for `agnes_sessions` / `agnes_telemetry` now uses the stable `user_id` (UUID) column instead of `username` (email local-part), which is mutable on user rename.

### Changes

1. **Schema v44→v45** (`src/db.py`): Adds `user_id VARCHAR` + indexes to `usage_session_summary` and `usage_events`. Idempotent (`ADD COLUMN IF NOT EXISTS`), wired into both fresh-install and upgrade paths.

2. **Pipeline integration** (`services/session_pipeline/runner.py`): `resolve_user_id()` maps session-directory names to stable `users.id` UUIDs (exact match for UUID dirs, escaped `LIKE` for email local-part dirs). Cached per-run. Passed as `user_id=` kwarg to all processors.

3. **RBAC filter** (`connectors/internal/access.py`): `agnes_sessions` and `agnes_telemetry` filter on `user_id` with an OR fallback on `username` via `legacy_username_column` for pre-backfill rows. `agnes_audit` uses `user_id` only (no legacy column).

4. **Processor** (`services/session_processors/usage.py`): Accepts `user_id` kwarg, stores it in both events and summary rows.

5. **Backward compat queries** (`app/api/me.py`, `app/api/admin_user_sessions.py`): Use `WHERE (user_id = ? OR username = ?)` during transition period.

6. **Backfill trigger**: `USAGE_PROCESSOR_VERSION` bumps 2→3 to trigger reprocessing, populating `user_id` retroactively. The reprocess path calls `purge_for_session()` (DELETE) before re-inserting, so both events and summaries get `user_id` populated.

7. **Verification processor** (`services/session_processors/verification.py`): Accepts `**kwargs` so the `user_id` kwarg passes through cleanly — no TypeError hack needed.

### Operational note

`upsert_events` uses `INSERT OR IGNORE` (not `INSERT OR REPLACE`), so a normal re-scan without the full purge path won't update existing event `user_id` values. The VERSION bump triggers the full purge+re-insert path which handles this correctly.

### Known limitation

If a session arrived via the collector under an old email local-part and the user later changed their email, `resolve_user_id` won't find a match → `user_id = NULL` → row invisible to the non-admin user. This is fail-closed (safe). UUID-dir sessions (upload API) are unaffected.

## Review & Testing Checklist for Human

- [ ] **Staging non-admin repro**: Log in as a non-admin user on a staging instance and verify `agnes_sessions` / `agnes_telemetry` return the correct scoped rows (both pre-backfill `username`-only rows and post-backfill `user_id` rows)
- [ ] Verify the v45 migration runs cleanly on an existing database (`ALTER TABLE ... ADD COLUMN IF NOT EXISTS`)
- [ ] After deploy, trigger `POST /api/admin/usage/reprocess` and confirm `user_id` gets populated in `usage_session_summary` and `usage_events`
- [ ] Verify admin users still see all rows (unscoped)

### Notes

- CHANGELOG entry added under `### Security` in `[Unreleased]`.
- 6 new `resolve_user_id` unit tests covering: exact UUID match, email local-part match, NULL fallback, tiebreak by `updated_at`, underscore escaping, UUID priority over email.
- The `legacy_username_column` field on `InternalTable` can be removed once all rows carry a non-NULL `user_id` after the backfill completes.

Link to Devin session: https://app.devin.ai/sessions/d0e30c9e8e294cf8acff5e614d62afe0
Requested by: @ZdenekSrotyr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Security
+
+- **RBAC filter uses stable `user_id` (UUID) instead of mutable email
+  local-part (#293).** Non-admin users querying `agnes_sessions` /
+  `agnes_telemetry` are now filtered by `user_id` (immutable UUID)
+  rather than `username` (email local-part, which changes on rename).
+  Schema v45 adds a `user_id` column to `usage_session_summary` and
+  `usage_events`; the session pipeline's `resolve_user_id()` populates
+  it on every (re)process run. `USAGE_PROCESSOR_VERSION` bumps 3→4 to
+  trigger backfill. During the transition period, RBAC queries include
+  an OR fallback on `username` so pre-backfill rows remain visible.
+
 ## [0.54.12] — 2026-05-14
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+## [0.54.13] — 2026-05-14
+
 ### Security
 
 - **RBAC filter uses stable `user_id` (UUID) instead of mutable email
--- a/app/api/admin_user_sessions.py
+++ b/app/api/admin_user_sessions.py
@@ -97,6 +97,8 @@ def list_user_sessions(
     # ------------------------------------------------------------------
     # Pull processed rows from usage_session_summary
     # ------------------------------------------------------------------
+    # Match on both user_id (stable, v45+) and username (legacy) so the
+    # admin view shows sessions from both ingestion paths and pre-v45 rows.
     try:
         rows_db = conn.execute(
             """
@@ -105,19 +107,27 @@ def list_user_sessions(
                 active_seconds, wall_seconds,
                 tool_calls, tool_errors, primary_model
             FROM usage_session_summary
-            WHERE username = ?
+            WHERE user_id = ? OR username = ?
             ORDER BY started_at DESC NULLS LAST
             """,
-            [username],
+            [user_id, username],
         ).fetchall()
     except Exception:
         rows_db = []
 
     processed_files: dict[str, dict] = {}
     if rows_db:
-        cols = ["session_file", "session_id", "started_at", "ended_at",
-                "active_seconds", "wall_seconds", "tool_calls", "tool_errors",
-                "primary_model"]
+        cols = [
+            "session_file",
+            "session_id",
+            "started_at",
+            "ended_at",
+            "active_seconds",
+            "wall_seconds",
+            "tool_calls",
+            "tool_errors",
+            "primary_model",
+        ]
         for r in rows_db:
             d = dict(zip(cols, r))
             # Normalise timestamps to ISO strings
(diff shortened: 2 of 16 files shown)


ZdenekSrotyr:
## Code review

Reviewed the full diff (12 files, +610/−477) against `main`.

### Does it fix #293? — Yes in substance, with caveats

The architecture is the right call: switching the `agnes_sessions` / `agnes_telemetry` RBAC filter from `username` (email local-part, mutable) to a stable `user_id` (UUID). After backfill, a non-admin gets `WHERE user_id = '<own-uuid>'` and sees only their own rows. `resolve_user_id()` covers both ingestion paths (UUID dir → exact match on `users.id`; collector dir → `email LIKE 'localpart@%'`). It **fails closed** — pre-backfill rows have `user_id = NULL` → invisible to non-admins.

**However:** I couldn't statically reproduce the exact "sees ALL users" symptom from `main`. On `main`, `execute_internal_query` *does* apply `WHERE username = '<local-part>'` for non-admins — the filter is present. The realistic main-branch bug looks more like "sees the wrong set / sees nothing" (since `agnes push` stores sessions under a UUID directory, so the `username` column holds UUIDs, not local-parts). Whether a real cross-user leak occurs depends on how `username` got populated in a given deployment. **Checklist item #1 (staging repro as a non-admin) must actually be performed before merge** — don't merge on faith.

### What's good
- ✅ CHANGELOG entry present (`### Security`).
- ✅ v45 migration is idempotent (`ADD COLUMN IF NOT EXISTS`), wired into both the ladder and the `if current < 45` block — matches the established pattern.
- ✅ Fail-closed behavior (NULL → invisible to non-admins).
- ✅ New tests `test_email_change_does_not_affect_visibility`, `test_non_admin_sees_sessions_uploaded_via_api`; CI `test` green.

### Issues — by severity

**1. CHANGELOG claims something the code doesn't do.** The entry says *"RBAC queries include an OR fallback on `username` so pre-backfill rows remain visible."* That's not true for the RBAC path — `build_filter_clause` was simplified to pure `WHERE user_id = '...'` (no OR). The `user_id = ? OR username = ?` fallback only exists in `me.py` / `admin_user_sessions.py`, which are the self/admin views, **not** the RBAC-scoped `agnes_sessions` path. Reality: until backfill completes, a non-admin querying `agnes_sessions` via `agnes query` sees **nothing**. The CHANGELOG describes the opposite — please correct the text.

**2. `except TypeError` fallback in `runner.py` is an anti-pattern.** `verification.py`'s `process_session` doesn't accept `user_id`, so it always takes the fallback. The problem: `except TypeError` catches *any* `TypeError`, including one raised *inside* `usage.process_session` (a genuine bug) — it then silently re-runs the processor without `user_id`, executing its side effects (`purge_for_session` + upsert) **twice** and obscuring which call failed. Cleaner: feature-detect once with `inspect.signature`, or just add `user_id` (or `**kwargs`) to `verification.py`'s signature.

**3. `resolve_user_id` — unescaped LIKE pattern.** `email LIKE ? || '@%'` with `username` taken from a directory name. A directory named `a_ice` would match `alice@…` (`_` is a LIKE wildcard); `%` would match anyone. In the canonical `agnes push` path `username` is always a server-assigned UUID (the exact-match branch wins first, LIKE never fires), so this is **low severity** — but it's a smell in a security fix. Escape `%` / `_`.

**4. `resolve_user_id` has no unit test.** It's the linchpin of the fix, yet the E2E tests bypass it (they insert `user_id` directly into the fixture). Untested: exact-match UUID, email-LIKE branch, NULL fallback, the `ORDER BY updated_at` tiebreak. Also missing: a transition-window test (a `user_id = NULL` row → invisible to non-admin) and a runner integration test that `run_processor` actually threads `user_id` through.

**5. Collector-dir + email-change isn't covered by backfill.** If a session arrived via the collector under an *old* local-part and the user later changed their email, `resolve_user_id` does `email LIKE 'oldlocalpart@%'` → no match → `NULL`. `test_email_change_does_not_affect_visibility` only exercises the query side (it pre-seeds `user_id`), not `resolve_user_id`. The UUID-dir (upload API) path is fine; single-domain instances are unaffected.

**6. Large formatting churn.** Most of the +610/−477 is pure `ruff`-style reformatting (string collapsing, set/list expansion, blank lines) across `src/db.py` and others. The actual fix is ~60 lines. It makes review harder and risks conflicts with #300/#301 — ideally a separate commit, or skipped.

### Recommendation

Approve after: (a) fixing the CHANGELOG text re: the OR fallback, (b) actually performing the staging non-admin repro (checklist #1), (c) adding a `resolve_user_id` unit test. Issues 2, 3, 5 are ideally bundled into the same PR (small, <30 min each) but aren't hard blockers — the fail-closed behavior holds the security guarantee even without them.

PR #301, opened 2026-05-14
feat(catalog): promote buckets to top-level Data Package cards

pcernik-grpn:
## Summary

Each bucket in `catalog_data` now renders as its own top-level Data Package card on `/catalog`, instead of being nested under a single "Core Business Data" wrapper as collapsible accordions. The page hero is already titled **Data Packages** (set in #284) — this PR makes the visual structure of the page match that title and the `bucket` grain of `table_registry`.

Follow-up to @ZdenekSrotyr's design system unification (#284); reuses the same flat-list pattern that the `Agnes Internal` and `Business Metrics` cards already use.

## Why

- **Removes a level of indirection** — one click to a table instead of two (expand bucket → click table).
- **Visual matches the data model.** `bucket` is already a first-class column on `table_registry` (`POST /api/admin/register-table`, `PUT /api/admin/registry/{id}`). The old UI hid that grain behind one wrapper.
- **Removes the awkward single-table accordion** when a bucket has only 1 table (e.g. Experiments).
- **Visual parity** with the existing `Agnes Internal` card — it already uses the flat-list pattern this PR adopts. (`Business Metrics` is unchanged and keeps its per-category accordions.)
- **Page hero title now describes what's on the page.** Today: "Data Packages" → one wrapper card. After: "Data Packages" → N package cards.

## Before / After

**Before**
```
[ Core Business Data ]                ← single source-card wraps all buckets
   ▸ Deal (2 tables)            collapsed accordions
   ▸ Location (2 tables)
   ▸ Finance (2 tables)
   ▸ Traffic (2 tables)
   ▸ Search (2 tables)
   ▸ Experiments (1 tables)
[ Agnes Internal ]
[ Business Metrics ]
```

**After**
```
[ Deal ]            ← top-level Data Package card, tables visible
  • ds1_session_deal_impressions    [Live]
  • ds2_session_deal_clicks         [Live]
[ Location ]
  • l0_session_location             [Live]
  • l0_session_location_changes     [Live]
[ Finance ]
  • order_economics                 [Local] · Profile
  • unit_economics                  [Live]
[ Traffic ]
  • s1_session_landings             [Live]
  • s2_session_pageviews            [Live]
[ Search ]
  • s4a_search_deal_detail          [Live]
  • s4b_search_context              [Live]
[ Experiments ]
  • xs1_session_experiments         [Live]
[ Agnes Internal ]                  ← unchanged
[ Business Metrics ]                ← unchanged
```

## What's preserved

- **Per-table behaviour is identical**: `Live` / `Local` query-mode badges, profile-link on local tables, `openProfiler()` click for non-remote, `Synced …` and `Queried directly from BigQuery` per-row info lines, `rows-badge` with `large` modifier.
- **Pluralization fix**: meta now reads `1 table` / `2 tables` (was always `N tables` before, even when N=1).
- **Agnes Internal** and **Business Metrics** cards are untouched.

## What's dropped

- The single aggregate header line on the old wrapper: `N tables · ~M rows total · Synced X`. Per-bucket counts replace `N tables`. The "rows total" string was rendering as `~ rows total` when `data_stats.rows_display` was missing — losing it is intentional.
- **Global aggregate sync timestamp.** The `Synced X` part of that header line is dropped with no replacement — the catalog page no longer shows a page-level sync timestamp. (Per-table `Synced …` lines on each row are unaffected. The Business Metrics card's `data-freshness-note` is *not* a replacement: the catalog handler passes `metrics_data=[]` unconditionally, so that card never renders here.)
- **Empty-state behaviour.** With the wrapper gone, an instance with zero registered tables now renders no Data Package cards at all, where the old `Core Business Data` wrapper always rendered (showing `0 tables`). Arguably an improvement, but it is a real behaviour delta.

## Scope of change

- `app/web/templates/catalog.html` — replaces the Core Business Data wrapper card (one `<div class="source-card">` around the bucket-accordion loop) with a per-bucket card loop. Mostly indentation/restructure.
- `CHANGELOG.md` — `### Changed` entry, plus the `0.54.11` release-cut.
- `pyproject.toml` — version bump `0.54.10` → `0.54.11` (release-cut).

No changes to:
- View handler (`app/web/router.py::catalog_data` shape is unchanged — same `category.name`, `category.count`, `category.tables`).
- CSS / design tokens (reuses existing `.source-card`, `.source-card-header`, `.accordion-content.expanded`, `.table-row` etc.).
- Registry / API / DB schema.

## Out of scope (intentional follow-ups)

- Per-bucket descriptions (`category.description`). Today the loop has no field for this; could be added later via a hardcoded map in the view layer, or via a `bucket_description` field on the registry. Suggest discussing separately if you want descriptions.
- Multi-column grid layout. Cards still stack single-column as today. A 2-col grid at wide widths would shorten vertical scroll but is a separate visual change.
- Per-bucket icons. All packages use the same primary "document" icon (matches the old wrapper). Differentiating icons per bucket (search/deal/location/etc.) could be a follow-up.

## Manual verification

Static preview comparing current vs. proposed: <https://github.com/pcernik-grpn/agnes-the-ai-analyst/pull/new/feat/catalog-buckets-as-packages> (local mockup file: `reports/agnes-ui-mockup/catalog-buckets-as-packages.html` on my laptop — happy to share via screenshot or live deploy if useful).

To verify locally:
```bash
docker compose up
# open http://localhost:8000/catalog
```

cc @ZdenekSrotyr — would value your sign-off given this lands on top of #284.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,15 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   accordion under a single "Core Business Data" wrapper. The page hero
   title ("Data Packages") now describes the actual visual structure, and
   the card grain matches the `bucket` column on `table_registry`. Tables
-  inside each package are flat-listed (no per-bucket accordion). The
-  `Agnes Internal` and `Business Metrics` cards are unchanged; their
-  flat-list pattern is what the bucket cards now follow. The aggregate
-  meta line ("N tables · ~M rows total · Synced X") that lived on the
-  old wrapper is dropped — the Business Metrics card's existing
-  `data-freshness-note` continues to surface the global sync timestamp.
+  inside each package are flat-listed (no per-bucket accordion),
+  mirroring the existing `Agnes Internal` card; the `Agnes Internal` and
+  `Business Metrics` cards themselves are unchanged. Per-table sync info
+  ("Synced …" / "Queried directly from BigQuery") on each row is
+  preserved. The aggregate meta line ("N tables · ~M rows total ·
+  Synced X") on the old wrapper is dropped with no replacement — the
+  global sync timestamp is no longer shown on this page. An instance
+  with zero registered tables now renders no Data Package cards at all,
+  where the old wrapper always rendered (showing "0 tables").
 
 ## [0.54.10] — 2026-05-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+## [0.54.11] — 2026-05-14
+
 ### Changed
 - Catalog page: each `catalog_data` bucket now renders as its own
   top-level Data Package card instead of being nested as a collapsible
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,19 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Changed
+- Catalog page: each `catalog_data` bucket now renders as its own
+  top-level Data Package card instead of being nested as a collapsible
+  accordion under a single "Core Business Data" wrapper. The page hero
+  title ("Data Packages") now describes the actual visual structure, and
+  the card grain matches the `bucket` column on `table_registry`. Tables
+  inside each package are flat-listed (no per-bucket accordion). The
+  `Agnes Internal` and `Business Metrics` cards are unchanged; their
+  flat-list pattern is what the bucket cards now follow. The aggregate
+  meta line ("N tables · ~M rows total · Synced X") that lived on the
+  old wrapper is dropped — the Business Metrics card's existing
+  `data-freshness-note` continues to surface the global sync timestamp.
+
 ## [0.54.10] — 2026-05-14
 
 ### Changed
--- a/app/web/templates/catalog.html
+++ b/app/web/templates/catalog.html
@@ -1332,7 +1332,13 @@
     <!-- ═══════════════ SOURCE CARDS ═══════════════ -->
     <div class="source-cards">
 
-        <!-- ── Card 1: Core Business Data ── -->
+        {# ── Data Packages: one source-card per bucket ─────────────────────
+           Previously all buckets were nested as accordions under a single
+           "Core Business Data" wrapper card. Each bucket is now its own
+           top-level Data Package card, matching the page hero title and
+           the bucket grain of the table_registry. The Internal and
+           Business Metrics cards below follow the same flat-list pattern. #}
+        {% for category in catalog_data %}
         <div class="source-card">
             <div class="source-card-header">
                 <div class="source-card-left">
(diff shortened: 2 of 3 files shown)


ZdenekSrotyr (commented):
Reviewed the diff and rendered the branch template locally against `main` to eyeball the before/after. Clean, well-scoped change — the per-bucket loop faithfully mirrors the existing `Agnes Internal` flat-list card (`accordion-content expanded` markup and all). Nice catch on the `1 table` / `2 tables` pluralization too.

**Verified the description's claims hold:**
- `data_stats` is still consumed — the `Business Metrics` card's `data-freshness-note` reads `data_stats.last_updated` — so leaving `app/web/router.py::catalog_data` untouched is correct, not dead context.
- `toggleAccordion` is still referenced by the `Business Metrics` card, so it isn't orphaned by dropping the bucket accordions.

**Two small things — neither blocking:**

1. _(description only, not the code)_ The "visual parity with the existing `Agnes Internal` **and `Business Metrics`** cards — they already use the flat-list pattern" line is half-right: `Business Metrics` still renders accordions (`accordion-trigger` + `toggleAccordion`). Only `Agnes Internal` is flat-list. Worth tightening the PR description so it doesn't imply `Business Metrics` changed shape.

2. _(behavior, currently unmentioned)_ Promoting the loop to top-level means an empty `catalog_data` now renders **zero** cards, whereas before the `Core Business Data` wrapper always rendered (showing `0 tables`). Almost certainly an improvement, but it's a real behavior delta — worth a one-liner under "What's dropped".

**Process flag for whoever merges:** `[Unreleased]` currently holds only this PR's entry. Per the "Release-cut belongs to the PR" rule in `CLAUDE.md`, if this lands without other PRs queued behind it, the version bump + release-cut should ship in the same PR rather than as a standalone follow-up.

Looks good to me otherwise. Leaving this as a comment rather than an approval since @ZdenekSrotyr is cc'd for sign-off.

ZdenekSrotyr:
Picking this up to land it. Pushed two commits on top of @pcernik-grpn's work and corrected the PR description:

**`346efb12` — `docs(changelog): correct catalog entry`**
The `[Unreleased]` bullet (and the PR description) made three claims that didn't match the diff:
- Claimed `Agnes Internal` **and** `Business Metrics` "already use the flat-list pattern" — only `Agnes Internal` does; `Business Metrics` keeps its per-category accordions.
- Claimed the dropped aggregate `Synced X` line is compensated by the `Business Metrics` card's `data-freshness-note`. It isn't — the catalog handler passes `metrics_data=[]` unconditionally (`app/web/router.py:962`), so that whole card never renders on `/catalog`. The global aggregate sync timestamp is simply no longer shown here (per-table `Synced …` lines on each row are unaffected).
- Omitted that an empty `catalog_data` now renders zero cards, where the old `Core Business Data` wrapper always rendered as `0 tables`.

CHANGELOG + PR description now state all three accurately.

**`47f64ff4` — `release: 0.54.11`**
`[Unreleased]` held only this PR's entry, so per the "release-cut belongs to the PR" rule in `CLAUDE.md` the cut ships here: `pyproject.toml` → `0.54.11`, `## [Unreleased]` renamed to `## [0.54.11] — 2026-05-14` with a fresh empty `[Unreleased]` on top. Separate last commit, as required.

**Verification**
- Full suite green locally: `4608 passed, 31 skipped` (`pytest tests/ -n auto`). (`.[server]` extra needed — `kbcstorage` — or 11 Keboola tests error on import; unrelated to this diff.)
- Independent second-pass review run on the branch. It confirmed the Jinja blocks are balanced, the context contract with `router.py` holds, `toggleAccordion` is not orphaned (still used by `Business Metrics`), and the release-cut is correct. Its one substantive finding — the inaccurate `data-freshness-note` claim — is fixed by `346efb12`.

**Deliberately not touched:** `data_stats` is still assembled in `app/web/router.py` and now has no live consumer on this page (its only remaining reference is inside the never-rendered `Business Metrics` card). Left as-is to keep this PR template-only and within @pcernik-grpn's stated scope; it's harmless dead plumbing, not a regression. Worth a follow-up cleanup but not worth expanding this PR's blast radius.

Original UI change is unchanged from @pcernik-grpn's `8df925d3` — credit for the feature is theirs.

PR #139, opened 2026-04-29
chore: release & CI/CD pipeline cleanup

ZdenekSrotyr:
Collapses three drifting sources of release truth (CHANGELOG.md, hardcoded pyproject version, CalVer claim-tag race loop) into a single coherent pipeline. 31 commits, 26 files changed (+2105 / −1161 lines), planned in `docs/superpowers/plans/2026-04-29-release-cicd-cleanup.md`.

## What changes

- **Versioning from git tags.** `pyproject.toml` switches to `setuptools_scm` dynamic version. The `v*` filter is applied at both build-time (nested `[tool.setuptools_scm.scm.git]` for setuptools_scm v8+) and runtime `get_version()` calls so CalVer build tags (`dev-YYYY.MM.N`) can never leak into `AGNES_VERSION`. CI build step + smoke-test both reject CalVer-shaped versions loudly.
- **Reusable test workflow.** `_test.yml` consolidates the install/lint/mypy/pytest recipe; `ci.yml`, `release.yml`, and `keboola-deploy.yml` all `uses:` it. One place to update, no more three-way drift. Removes ~80 lines of duplicated YAML.
- **Image pipeline simplified.** `release.yml` no longer triggers on every branch push. Push to `main` → auto-build of `:stable` + `:stable-<run-number>` + `:sha-<7>`. Non-main dev images require `gh workflow run release.yml -r <branch>` (operator-intent, not automatic). The CalVer claim-tag race loop is replaced by `${{ github.run_number }}`.
- **Per-branch tag aliases preserved on manual builds.** `gh workflow run release.yml -r zs/foo` still publishes `:dev-zs-foo` + `:dev-zs-latest` (per-prefix floating alias), exactly the pre-cleanup format. Git Flow prefixes (`feature/`, `fix/`, etc.) skip the `-latest` alias as before.
- **Release notes automated.** Release Drafter maintains a rolling draft GitHub Release from PR titles using conventional-commit prefix autolabeling. `CHANGELOG.md` and `docs/RELEASE_TEMPLATE.md` are deleted; GitHub Releases is the single source.
- **Rollback extracted to its own workflow.** `.github/workflows/rollback.yml` is callable via `workflow_call` (from `release.yml` smoke-test failure) or `workflow_dispatch` (manual operator-triggered rollback). Uses run-number arithmetic (\`stable-<N-1>\`) instead of the broken \`git tag -l "stable-*"\` heuristic. Opens a tracking issue tagged \`bug,rollback\`.
[...]

PR #314, opened 2026-05-15
ci: consolidate release pipeline (salvageable subset of #139)

ZdenekSrotyr:
## Summary

Re-proposes the salvageable subset of the abandoned **#139** (`chore/release-cleanup`), decoupled from #139's rejected direction. Each piece is orthogonal CI/release-workflow plumbing — no application-code changes.

## What's in

- **Reusable `_test.yml` workflow** — single source of the install/lint/typecheck/pytest recipe; `ci.yml`, `release.yml`, and `keboola-deploy.yml` all `uses:` it. Each caller's prior behavior preserved exactly (ci.yml: no lint/mypy, `-n auto`; release.yml: lint+mypy + `-n auto`, with the existing `if: github.event_name != 'create'` guard; keboola-deploy.yml: lint+mypy, no `-n auto`).
- **`rollback.yml`** — the smoke-test rollback extracted into a reusable + manually-dispatchable workflow. `release.yml` gains a `rollback-on-smoke-fail` job calling it; the inline rollback step is removed from `smoke-test`. Adapted from #139's run-number scheme to the repo's CalVer reality (`git tag -l "stable-*" --sort=-version:refname`). Backward-compatible `deprecated-YYYY.MM.N` naming preserved via `${FAILED#stable-}`. Warning guard for non-`stable-*` `workflow_dispatch` inputs.
- **`prune-dev-tags.yml` + `scripts/ops/prune-dev-tags.sh`** — weekly cron (Sundays 04:00 UTC) that prunes legacy CalVer `dev-YYYY.MM.N` / `stable-YYYY.MM.N` git tags + GHCR images outside a `KEEP_MONTHS` retention window. `workflow_dispatch` supports a dry-run. Section-1 (CalVer) only — #139's dead Section-2 run-number prune is dropped. Floating aliases (`:stable`, `:dev`, `*-latest`) are git-tagless and never matched.
- **`lint-workflows.yml`** — runs `actionlint` on changes under `.github/workflows/**` + `scripts/ops/**.sh`. `continue-on-error: true` initially (non-blocking) until the existing workflow inventory is `actionlint`-clean.
- **Cleanup** — deletes the self-labeled superseded `deploy.yml` stub + its now-orphaned `.gitignore` entry.

## What's explicitly NOT included (rejected from #139)

- Release Drafter + deleting `CHANGELOG.md` — the repo formalized CHANGELOG-centric release discipline in `docs/RELEASING.md` (#306).
- `setuptools_scm` dynamic versioning — repo uses manual `pyproject.toml` bumps + `vX.Y.Z` tags.
- Run-number tag scheme (`dev-N` / `stable-N`) — never adopted; CalVer is in use.
- `release.yml` trigger change to main-only — dev VMs depend on the auto-build-on-every-push behavior.
- Deletion of `cli-wheel-clean-install` job — that job catches a real class of regression (`kbcstorage` / `urllib3` / wheel metadata conflicts).
- The smoke-test bugs #139 introduced (removed GHCR login, removed `GH_TOKEN` env).

## Testing

- Full pytest suite: **4605 passed, 32 skipped**. The 11 failures are all `test_keboola_*` (`ModuleNotFoundError: No module named 'kbcstorage'`) — a missing optional dependency, pre-existing on `main` (reproduces on clean checkout). Unrelated to this PR (zero application code touched).
- `actionlint` clean on all 8 workflow files (including the new ones).
- `prune-dev-tags.sh`: `bash -n` clean; `PRUNE_DRY_RUN=1` invocation lists-only and exits 0 (no deletions; verified the local tag count is unchanged).
- Real workflow-run behavior is verifiable only by triggering each workflow — CI on this push will exercise the new `_test.yml` refactor and auto-trigger `lint-workflows.yml` (which lints the whole tree, including itself).

## Release

Cuts `v0.54.17` — `### Internal` only entry (CI tooling; no user-visible behavior changes).

## Supersedes

Supersedes #139 — closed manually on merge.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/314" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


cvrysanek:
### Automated PR Review (via Claude Code)

**Highest priority fixes before merging:** The double rollback, the live-delete scheduled run, and pruning of rollback targets are all independently capable of leaving `:stable` pointing at a broken image with no automated recovery path.

#### HIGH

* **Double rollback risk / Live-delete scheduled run risk / Pruning of rollback targets:** These can independently leave `:stable` pointing at a broken image with no automated recovery path.
* **Logic flaw in jq filter:** Review the `ap(startswith("stable-")) | any) | not` logic.
* **`gh api --paginate` produces newline-delimited JSON arrays; jq fails on page 2+** (`prune-dev-tags.sh:88-100`)
  `gh api --paginate` outputs each page as a *separate* JSON array on its own line. Without `--slurp`, `jq` errors after the first array; `|| echo ""` swallows the error. Git tags are deleted, GHCR images are silently skipped, and the script exits 0 with "Prune complete." Fix: pipe through `jq -s 'add // []'` before the main filter.
* **Partial git tag deletion leaves local and remote divergent** (`prune-dev-tags.sh:70-73`)
  `git tag -d $TAG` runs unconditionally even when `git push origin --delete` failed. Local tags disappear; remote tags survive. No failure is propagated; the script exits 0. Fix: track a failure counter and exit non-zero if any push failed.

#### MEDIUM

* **`deprecated-*` tags are never pruned** — `rollback.yml` creates `deprecated-*` tags that don't match the `^(dev|stable)-…` pattern in `prune-dev-tags.sh`. They accumulate indefinitely.
* **Top-level `permissions:` block in `rollback.yml` is a no-op when called via `workflow_call`** — Caller job permissions govern the token; the callee's top-level block is ignored. Documented for direct `workflow_dispatch` use but misleading for maintainers.

#### LOW

* **Shellcheck path mismatch** (`lint-workflows.yml:9` vs `:36`)
  Trigger uses `scripts/ops/**.sh` (recursive) but shellcheck only runs `scripts/ops/*.sh` (top-level only). A script in a subdirectory triggers the workflow but is never checked.
* **Comment in `release.yml` says `issues: write` is "default for jobs"** — the GitHub Actions default is `read`, not `write`. The comment is factually wrong.

PR #316, opened 2026-05-15
fix(store): surface review failures + harden publish gate

cvrysanek:
## Summary
Four independent fixes to the flea-market submission pipeline, all
surfaced by an admin upload that landed at \`status='approved'\` without
an LLM review. Plan files: \`when-i-submitted-new-delightful-russell.md\`
+ \`publish-gate-fail-closed.md\`.

### 1. LLM truncation no longer pins submissions in \`review_error\`
- \`MAX_RESPONSE_TOKENS\` 2500 → 6000 in \`llm_review.py\`
- One-shot retry-with-doubled-budget in \`anthropic_provider.py\` (cap 4× initial)

### 2. Detail page surfaces failures even when prior version still serves
- Widened \`_quarantine_banner\` gate so \`blocked_inline / blocked_llm /
  review_error\` render even when entity stays \`visibility=approved\`
  under deferred promotion
- Tweaked H3 + lede copy in each branch to distinguish "initial quarantine"
  from "latest edit failed, prior version still serving"

### 3. Restore guard
- New \`StoreEntitiesRepository.get_with_version_approvals\` joins version
  history with submission verdicts
- UI gates Restore button on \`submission_status in ('approved', None)\`,
  renders pills (blocked / errored / pending) for non-restorable rows
- Backend 400 \`version_not_approved\` guard in \`POST .../versions/{n}/restore\`
- JS handler covers the new error code

### 4. **BREAKING (operator-facing)**: publish gate is fail-CLOSED
- Split \`get_guardrails_enabled()\` (intent — YAML) from new
  \`get_guardrails_llm_provider_ready()\` (env — credentials)
- Three-state matrix:
  | \`enabled\` | provider ready | Behavior |
  |---|---|---|
  | false | any | Auto-approve (unchanged) |
  | true | yes | Normal hold-for-review (unchanged) |
  | true | **no** | Submissions hold at \`pending_llm\` (was: silent auto-approve) |
- Admin "Retry review" eligibility extended to \`pending_llm\` rows
- Boot-time \`WARNING\` banner when enabled-but-not-ready
- \`docs/STORE_GUARDRAILS.md\` updated with the matrix
- Operators who relied on the auto-fallback for local dev must now set
  \`guardrails.enabled: false\` in \`instance.yaml\` explicitly

## Test plan
- [x] 4623 passed (3 pre-existing unrelated failures: catalog_export
      tmpdir race, local \`da\` binary on PATH)
- [x] New: \`TestPublishGateFailClosed\` (4 tests) covering each cell of
      the matrix
- [x] New: \`TestRestoreVersion::test_restore_rejects_blocked_llm_version\`,
      \`...rejects_review_error_version\`, \`...allows_legacy_v1_*\`
- [x] New: \`TestEditPageBanner::test_banner_shows_review_error_when_prior_version_still_serving\`
- [x] New: \`test_truncation_retries_with_doubled_budget\`,
      \`test_truncation_persistent_raises_format_error\`
- [x] Updated \`tests/conftest.py\` autouse fixture defaults flea guardrails
      OFF so legacy tests don't need to know about the new toggle
- [x] Live verified against running uvicorn: synthetic \`eviltest2\` skill
      with \`curl … | sh\` correctly landed \`blocked_llm\` with risk=critical
      + 2 findings

## Follow-up
Pre-submit dry-run endpoint (so submitters see findings before final
submit) tracked separately as an issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,17 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   banner now renders for those failure statuses too, with copy that
   acknowledges the prior version is still live ("Latest edit failed
   review — previously approved version (vN) keeps serving …").
+- Flea-market admin **Override** on a v2+ edit/restore submission now
+  promotes the entity to the overridden version + swaps the on-disk
+  live bundle. Pre-fix the override only flipped
+  `visibility_status='approved'` and `submission.status='overridden'`,
+  leaving `entity.version_no` at the prior approved version — so
+  installers (and the marketplace UI) kept serving the OLD bytes the
+  admin just intended to replace. Mirrors the auto-approval branch in
+  `runner.run_llm_review`: look up the submission's version in
+  `version_history`, `promote_version` + `_swap_live_to_version` when
+  it differs from current. Initial-v1 overrides unchanged (no
+  promotion needed).
 - Flea-market Restore button + endpoint no longer allow restoring a
   version that was never approved. The versions card hid the gate
   entirely (showed Restore on any non-current row), and the backend
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,24 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   banner now renders for those failure statuses too, with copy that
   acknowledges the prior version is still live ("Latest edit failed
   review — previously approved version (vN) keeps serving …").
+- Flea-market admin **Retry review** and **Rescan** now review the
+  STAGED version's bundle, not the live `plugin/` directory. For a
+  v2+ edit held at `pending_llm` / `blocked_llm` / `review_error`,
+  live still holds the prior approved version. Reviewing live would
+  produce a verdict against the WRONG bytes; the runner's hash-match
+  promotion would then advance the entity to staged bytes that were
+  never actually reviewed. Now resolves the staged
+  `versions/v<N>/plugin/` from the submission's `version_history`
+  entry. (Critical — surfaced by adversarial review.)
+- Flea-market admin **Override** is now forward-only on promotion.
+  Previously `target_version_no != current` would happily demote the
+  live bundle when admin overrode a stale v2 submission while v3 was
+  already approved + live. Changed to `target > current` so override
+  flips status + visibility on the row regardless, but on-disk
+  promotion only fires for newer versions. The same `> current`
+  guard is now applied defensively in `runner.run_llm_review` so a
+  late LLM verdict can't demote past a more recent approval either.
+  (High — surfaced by adversarial review.)
 - Flea-market admin **Override** on a v2+ edit/restore submission now
   promotes the entity to the overridden version + swaps the on-disk
   live bundle. Pre-fix the override only flipped
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,6 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
-### Fixed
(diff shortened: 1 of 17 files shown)


ZdenekSrotyr (dismissed):
Holding this back per an independent review pass. The four headline changes are well-architected — BREAKING discipline (CHANGELOG marker + migration step + boot warning + docs matrix) is exemplary, the three-state matrix in `app/api/store.py` is cleanly applied across create/update/restore, the truncation-retry math is correct, and the override v2-promote fix is a real find with a strong regression test.

There's **one blocker** (data-corruption regression) and a few smaller items.

## Blocker — Retry & Rescan target the wrong dir for v2+ `pending_llm`

`app/api/admin.py:3981` (Retry) and `app/api/admin.py:3859` (Rescan) both unconditionally use `plugin_dir = _plugin_dir(entity_id)` (the live `plugin/`).

This PR extends Retry-eligibility to `pending_llm`. Under fail-CLOSED with no LLM key, a v2+ edit/restore submission sits at `pending_llm` with its bundle in `versions/v2/plugin/` while live `plugin/` still serves the prior approved v1. When the admin sets the key and clicks **Retry review**:

1. LLM reviews `_plugin_dir(entity_id)` — which is **v1 bytes**.
2. v1 was already approved once, so the review almost certainly passes.
3. `run_llm_review`'s promote-on-approval branch swaps live to "v2", but the on-disk bundle is v1 bytes.
4. Silent data corruption: entity claims version_no=2 but `plugin/` holds v1 content.

The **override path** at `app/api/admin.py:3756-3771` already does this right — walks `entity.version_history` for `entry.hash == sub.version`, resolves `target_version_no`, promotes + swaps from the version dir. Retry + Rescan need the same lookup before picking `plugin_dir`.

The new test `test_admin_retry_pending_llm_fires_review` (`tests/test_store_entity_versions.py:1293-1349`) only covers v1 initial-upload, where the issue can't surface (live == v2 effectively). A v2+ `pending_llm` retry test would lock this in.

Same wrong-dir issue likely applies to v2+ `review_error` / `blocked_llm` pre-PR — but that's pre-existing exposure; this PR newly opens the `pending_llm` path, so worth fixing as part of this work.

## Other Important (optional, author's call)

- **`MAX_TRUNCATION_RETRIES` naming** (`connectors/llm/anthropic_provider.py:31-32, 109-136`). The constant means "at most 2 budget doublings, capping at 4×", not "at most 2 retries". A future maintainer bumping `MAX_RETRIES=5` would expect truncation retries to follow but they're separately capped — only because of the explicit truncation counter. Either rename to `MAX_TRUNCATION_BUDGET_DOUBLINGS=2` or add a unit test that pins `MAX_RETRIES=5, MAX_TRUNCATION_RETRIES=2` and asserts the cap holds.
- **Partial-promote/swap state** (`app/api/admin.py:3768-3782`). If `promote_version` succeeds but `_swap_live_to_version` raises, DB claims v2 but `plugin/` is still v1. `logger.exception` runs, handler returns `ok=True`. Either rollback the promote, return a structured warning, or document the recovery path. (Same shape may exist in the runner's approval branch — at minimum mirror whatever it does.)
- **Defensive copy in `get_with_version_approvals`** (`src/repositories/store_entities.py:521-526`). The method mutates each history entry in place (`entry["submission_status"] = ...`). `self.get(id)` currently re-parses JSON each call so this is fine today, but adding `entry = dict(entry)` before mutation is free future-proofing for any future caching layer on top.
- **Override eligibility extension** (`app/api/admin.py:3729`). `override` currently accepts `{"blocked_inline", "blocked_llm", "review_error"}`. Adding `"pending_llm"` is a one-token change that closes a small UX gap: under fail-CLOSED with no key, admins know-good submissions sit indefinitely with no way to publish without first clicking Retry. Override uses the correct version-resolution already, so this is safe.

## Minor (non-blocking)

- `CHANGELOG.md:62-65` includes a parenthetical narrative about a rolled-back `maxItems=20` attempt — trim to the shipped state.
- `_quarantine_banner.html` uses `_is_edit_review = entity.visibility_status == 'approved'` which conflates "v2+ deferred edit" with "happens to be approved right now". Edge case if archive-then-restore can interleave with a failed edit — worth a sanity check.
- `tests/test_store_entity_versions.py:866-909` test name `test_banner_shows_version_n_under_review` doesn't match what it actually asserts (it checks the edit-button lock, not the banner). Rename suggested.

Once #2 is addressed I'll re-review and merge through (per the releaser role).

ZdenekSrotyr (approved):
Approved after polish; review findings all addressed.

Issue #318, opened 2026-05-15
store: adversarial-review follow-ups (export leak, race, idempotency, atomicity, prompt-escape, missing tests)


cvrysanek:
## Context

Codex adversarial review of the flea-market guardrail pipeline (PR
#316) surfaced 8 issues across CRITICAL → LOW. Two were fixed in PR
#316 follow-up commits (\`526e6a5\` retry/rescan reviews staged bundle,
override forward-only). The remaining six + the suggested missing
tests are tracked here.

### [CRITICAL] Bundle export leaks quarantined entities
\`app/api/store.py:2368-2384, 2401-2404\` — any authenticated user can call
the Store backup export; \`repo.list(...)\` has no \`visibility_status\`
filter, so pending / hidden v1 bytes can be downloaded by non-owners.
**Fix**: filter to \`visibility_status='approved'\` for non-admin
non-owner callers (mirror the browse endpoint's gate).

### [HIGH] Concurrent PUTs can both pass the pending gate
\`app/api/store.py:1482-1495, 1538-1541\`,
\`src/repositories/store_entities.py:320-342\` — \`latest_for_entity\`
SELECT happens before filesystem work + history INSERT. Two PUTs
racing on the same entity_id can both pass the gate, both append
history entries, and race on \`versions/v2/plugin/\`. No
compare-and-swap guard on \`append_version_history\`.
**Fix**: per-entity advisory lock OR optimistic-concurrency on entity
\`updated_at\` (SELECT … WHERE updated_at = ? ... ).

### [HIGH] Background LLM tasks are not status-idempotent
\`src/store_guardrails/runner.py:228-233, 332-337\`,
\`src/repositories/store_submissions.py:176-198\` — \`update_status\`
blindly overwrites current status. A retry task racing with an admin
override can let a late verdict clobber \`overridden\` (or another
terminal state).
**Fix**: terminal-state guard in \`update_status\` —
\`UPDATE … WHERE status NOT IN ('approved','overridden','blocked_inline')\`
and check rowcount; or compare-and-swap on expected prior status.

### [MEDIUM] DB promotion and live swap are not atomic
\`src/repositories/store_entities.py:372-383\`,
\`app/api/store.py:794-815\`,
\`src/store_guardrails/runner.py:283-289\` — DB row is promoted
before \`_swap_live_to_version\`. If source dir is missing, swap
returns \`False\` and callers ignore it. A crash between renames can
leave no \`plugin/\` dir at all.
[...]

PR #320, opened 2026-05-15
fix(store): close 1 critical + 2 high adversarial-review findings (C2/H2/H3 from #318)

cvrysanek:
Follow-up to PR #316 (now merged). Addresses 3 findings from Codex's
adversarial review tracked in issue #318:

## C2 [CRITICAL] — bundle export leaked quarantined entities

\`/api/store/bundle.zip\` called \`repo.list(...)\` without a
\`visibility_status\` filter. Any authenticated non-admin could
download pending / blocked / hidden v1 bytes — bypassing the publish
gate the same way \`_enforce_visibility\` already prevents on detail
pages + install.

Fix: mirror the marketplace browse-listing gate. Non-admin sees only
\`approved\` (plus their own non-approved entries via
\`include_owner_id\`); admins skip the filter.

## H2 [HIGH] — concurrent PUT race

Two concurrent PUTs on the same entity could both pass the
\`latest_for_entity\` pending gate, both bake into
\`versions/v<N+1>/plugin/\`, both append a \`version_history\` entry.

Fix: per-entity \`asyncio.Lock\` (\`_hold_entity_write_lock\`) wrapping
the version-creating critical section in \`update_entity\` and
\`restore_version\`. Single-process deployments are now serialized;
multi-worker deployments still have a residual window — tracked in
#318 as a future "DB advisory lock / fs flock" improvement.

## H3 [HIGH] — BG-task non-idempotent

\`update_status\` blindly overwrote any current status. A late
background-task LLM verdict could clobber an \`overridden\` row back
to \`approved\` / \`blocked_llm\` after admin had already
force-published.

Fix: compare-and-swap on terminal statuses
(\`approved\`, \`overridden\`, \`blocked_inline\`) — \`UPDATE … WHERE
status NOT IN (...)\`. Callers that legitimately overwrite (admin
rescan, manual repair) pass \`allow_terminal_overwrite=True\`.
Returns bool indicating whether the write landed; BG callers no-op
on terminal rows.

## Test plan

- [x] 296 tests green on the touched surface
  (\`store_entity_versions\`, \`admin_store_submissions\`, \`store_api\`,
  \`store_guardrails_llm\`, \`store_guardrails_inline\`,
  \`store_put_atomic\`, \`store_repositories\`, \`llm_connector\`)
- [x] 7 new tests covering each fix
- [x] Live verified: bundle.zip as non-owner non-admin no longer
  includes pending entities

## Remaining from #318

- Medium: prompt sentinel escape misses filenames
- Medium: DB promote + live swap not atomic
- Low: admin forensic download serves wrong bytes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,13 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   longer silently clobber the row. Callers that legitimately need to
   overwrite a terminal state pass `allow_terminal_overwrite=True`
   explicitly. Returns a boolean indicating whether the write landed.
+  `runner.run_llm_review` now honors the bool on both its `approved`
+  and `blocked_llm` branches: a CAS no-op skips the downstream
+  cascade (visibility flip, version promote, the verdict-specific
+  audit entry that would otherwise contradict the row) and logs a
+  single `store.submission.bg_verdict_skipped` audit row instead,
+  so an operator reviewing the queue sees dropped verdicts
+  explicitly rather than via row-vs-audit contradiction.
   (High — surfaced by adversarial review.)
 - Flea-market admin **Retry review** and **Rescan** now review the
   STAGED version's bundle, not the live `plugin/` directory. For a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,31 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
   banner now renders for those failure statuses too, with copy that
   acknowledges the prior version is still live ("Latest edit failed
   review — previously approved version (vN) keeps serving …").
+- Flea-market `/api/store/bundle.zip` export now filters by
+  `visibility_status='approved'` for non-admin non-owner callers.
+  Previously an authenticated user could call the bundle endpoint
+  and pull pending / blocked / hidden v1 bytes — bypassing the
+  publish gate the same way `_enforce_visibility` already prevents
+  on detail pages + install. Owners still see their own non-approved
+  entries (matches the browse-listing `include_owner_id` affordance);
+  admins still see everything. (Critical — surfaced by adversarial
+  review.)
+- Flea-market PUT (edit) + restore endpoints now serialize concurrent
+  writes against the same `entity_id` via a per-entity asyncio lock.
+  Two concurrent PUTs could previously both pass the
+  `latest_for_entity` pending gate, both bake into
+  `versions/v<N+1>/plugin/`, and both append a `version_history`
+  entry. The lock closes the window for single-process deployments;
+  multi-worker deployments still have a residual window (tracked in
+  the follow-up issue). (High — surfaced by adversarial review.)
+- Flea-market `StoreSubmissionsRepository.update_status` is now
+  compare-and-swap on terminal statuses (`approved`, `overridden`,
+  `blocked_inline`). A late background-task LLM verdict racing with
+  an admin override or with a more recent terminal verdict can no
+  longer silently clobber the row. Callers that legitimately need to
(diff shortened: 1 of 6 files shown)


minasarustamyan:
Reviewed each fix against the head of the branch. C2 and H2 look good; H3's DB-level fix is correct but I think the BG runner needs a small follow-up to observe it.

## C2 (bundle.zip leak) — LGTM

The filter at `app/api/store.py:2511-2533` mirrors the existing browse-listing filter at `app/api/store.py:1040-1056` exactly, and the repo helper at `src/repositories/store_entities.py:616-635` does the right `(visibility_status IN (...) OR (owner_user_id = ? AND visibility_status != 'archived'))` shape. Test matrix (non-owner / owner / admin) covers the cases.

## H3 (`update_status` CAS) — DB part correct, runner doesn't observe the no-op

The CAS itself is right. I verified empirically that DuckDB's `conn.execute(\"UPDATE …\").fetchone()` returns `(rowcount,)` (`(0,)` for no match, `(1,)` for one match), so the bool return at `src/repositories/store_submissions.py:226-234` is accurate.

The gap: `src/store_guardrails/runner.py` calls `update_status` at five sites and ignores the return value at all of them:

- `runner.py:174` — `review_error` (plugin_dir missing)
- `runner.py:189` — `review_error` (LLM config error)
- `runner.py:210-215` — `review_error` (LLM call returned error)
- `runner.py:228-233` — `approved` path
- `runner.py:335-340` — `blocked_llm` path

When the CAS no-ops because the row is already terminal (e.g. admin force-published with `set_override` while the LLM was still running — the exact H3 race), the runner continues as if the write landed:

- **`approved` branch (228-333)**: still runs `set_visibility_if_pending`, `update_flea_attribution`, and attempts `promote_version` + `_swap_live_to_version`, then audit-logs `store.submission.approved` with `visibility_flipped` / `promoted_to_version_no` in `params`. The forward-only guard at lines 286-287 prevents demotion, but the audit row claims an approval landed when the submission row is still `overridden`.
- **`blocked_llm` branch (335-360)**: audit-logs `store.submission.blocked_llm` with `result=\"blocked\"` while the row actually remains `overridden`.
- **Three `review_error` sites**: same shape — audit rows claim an error landed while the row remains in its prior terminal state.

The runner already has the right pattern for an analogous case: when an \"approved\" verdict comes back but the entity has left the serve-able states (e.g. archived mid-flight), it emits `action=\"store.submission.bg_verdict_skipped\"` with `result=\"skipped\"` at lines 309-320. Extending that to \"CAS no-opped because the submission row is already terminal\" would close the gap:

```python
wrote = subs_repo.update_status(submission_id, status=\"approved\", ...)
if not wrote:
    audit.log(action=\"store.submission.bg_verdict_skipped\",
              params={\"attempted_verdict\": \"approved\",
                      \"reason\": \"submission already terminal\", ...},
              result=\"skipped\")
    return LlmResult(verdict=verdict, reviewed_by_model=model)
```

DB state can no longer be corrupted, which was the catastrophic part of H3 — that's fixed. But misleading audit rows and wasted side-effect work in the approved branch (attribution write, promotion attempt) are themselves review-trail concerns, and worth closing in the same PR since the bool is already returned for exactly this purpose.

## H2 (per-entity asyncio.Lock) — works, one small nit

Lock semantics are correct: `_entity_write_locks_guard` removes the TOCTOU on the dict lookup, module-level `asyncio.Lock()` is fine on 3.12 (lazy loop binding), and the single-process scope is acknowledged in the PR body.

Minor: `_entity_write_locks` at `app/api/store.py:417` never evicts entries — I grep'd for every reference and only the `.get()` + assignment in `_hold_entity_write_lock` exist. The dict grows linearly with the number of distinct `entity_id`s ever touched by PUT or restore. Not a memory bomb for typical store sizes, but a `weakref.WeakValueDictionary` or release-time eviction would make this clean. Not a blocker for this PR.

PR #342, opened 2026-05-18
feat(flea): marketplace refactor — data model, attribution, UI unification

### LINKED ISSUE #335: Flea telemetry attribution bug — bundle skill/agent invocations never reach rollup tables
## Summary

Flea-market skill/agent invocations from Claude Code never get attributed in the v48 telemetry rollup. Every flea event lands as `source='builtin'` in `usage_events`; `usage_marketplace_item_daily` / `_window` receive zero rows for `source='flea'` (apart from flea-plugin entities, which work via a different code path). Consequence: flea cards on `/marketplace?tab=flea` show no chip (`active · calls · trend`) — chip gate `(stackCount > 0 || calls > 0)` stays false, the only stat that ever surfaces is `install_count`, and "Most used (30d)" / "Most adopted (30d)" sort orderings produce identical zero-keyed result sets that effectively fall back to recency.

## Root cause

Two distinct namespaces for flea entity names:

- **`store_entities.name`** — raw name as uploaded (e.g. `"flea-skill"`).
- **Served bundle path + `SKILL.md` frontmatter `name`** — suffixed: `<name>-by-<owner_username>` (e.g. `"flea-skill-by-vrysanek"`). Written by `app/api/store.py:803,814` and surfaced in the `agnes-store-bundle` plugin tree.

Claude Code invokes flea skills as `agnes-store-bundle:<suffixed>` (the bundle is one synthetic plugin that wraps every flea skill / agent the user has in their stack — `src/marketplace_filter.py:184`). The JSONL `tool_use.input.skill` field carries the suffixed name verbatim.

The session-processor's `MarketplaceItemLookup` preload reads the **raw** name:

```python
# services/session_processors/usage_lib.py:280-284
self._flea_entities: dict[str, str] = {
    row[0]: row[1] for row in conn.execute(
        "SELECT name, type FROM store_entities WHERE visibility_status='approved'"
    ).fetchall()
}
```

Then `resolve()` at `usage_lib.py:312` does:

```python
if prefix == FLEA_BUNDLE_PREFIX:
    ent_type = self._flea_entities.get(local)   # local = "flea-skill-by-vrysanek"
    if ent_type:
        return ("flea", "", local, ent_type)
    continue   # falls through to "builtin"
```

Lookup key (`"flea-skill-by-vrysanek"`) never matches preload key (`"flea-skill"`). Every flea skill / agent invocation drops to `builtin` → no rollup row → empty stats → no chip → `most_used` / `most_adopted` sorts produce a stable but stat-blind order.

The flea-plugin path (`type='plugin'` entities) is unaffected — those use `BUNDLE_PLUGIN_NAME` separately and never go through the `agnes-store-bundle:` prefix.

## How the test suite missed this

`tests/fixtures/sessions/usage/skill_flea.jsonl:2` writes:

```json
{"tool_use": {"name": "Skill", "input": {"skill": "agnes-store-bundle:flea-skill", "args": ""}}}
```

It uses the **bare** name, not the suffixed name Claude Code actually emits. The fixture matches the buggy lookup key, so `test_session_processor_usage.py::test_skill_flea` passes while production fails. Same pattern as the v42 attribution bug PR #329 fixed (commit body: "AttributionLookup keyed on bare skill names while Claude Code writes `<plugin>:<local>` in JSONL") — v48 fixed the prefix-split but missed the suffix-strip on the local part.

## Evidence (read-only, code-trace)

- Upload path bakes suffix into served bundle: `app/api/store.py:803` (skill dir name), `app/api/store.py:814` (`_set_frontmatter_name(text, suffixed)`).
- Served bundle invocation prefix: `src/marketplace_filter.py:184` (`BUNDLE_PLUGIN_NAME = "agnes-store-bundle"`).
- Suffix helper: `src/store_naming.py:58` (`suffixed_name(original_name, username)` → `f"{original_name}-by-{username}"`).
- Attribution lookup loads raw name: `services/session_processors/usage_lib.py:280-284`.
- Test fixture uses bare name: `tests/fixtures/sessions/usage/skill_flea.jsonl:2`.
- Front-end chip gate is **not** the problem — `app/web/templates/marketplace.html:802` (`if (stackCount > 0 || calls > 0)`) renders the chip whenever the API returns non-zero. Card render is fully source-agnostic; same `buildCard()` for curated + flea.
- API stat loader is source-parameterized correctly: `app/api/marketplace.py:381` (`_load_invocation_stats(conn, "flea")`); the bug is upstream of this — the rollup tables it queries have no flea rows to load.

## Surfaces affected

When `usage_marketplace_item_daily` / `_window` have no `source='flea'` rows:

- `/marketplace?tab=flea` cards — no `.inv-chip` row rendered.
- `/marketplace?tab=flea&sort=most_used` — every item ties at 0, falls back to (effectively) recency.
- `/marketplace?tab=flea&sort=most_adopted` — same.
- `/marketplace?tab=flea&sort=trending` — sort option hidden via `available_sorts` (correct response to zero-trend-data, masks the bug).
- Most Popular strip on flea tab — `loadMostPopular()` filters `it.invocations_30d > 0`, returns empty, strip hidden.
- Flea skill/agent detail page hero chip + sidebar telemetry — empty (`_build_telemetry` returns `None` when `invocations_30d == 0`).
- Inner skill/agent cards on flea plugin detail page — `_load_inner_item_stats` returns empty.
- Curated, plugin-type-flea, my-tab — all unaffected.

### Option A — Index lookup by suffixed name (read-side change too)

```python
# services/session_processors/usage_lib.py
self._flea_entities: dict[str, str] = {
    f"{row[0]}-by-{row[2]}": row[1]
    for row in conn.execute(
        "SELECT name, type, owner_username FROM store_entities WHERE visibility_status='approved'"
    ).fetchall()
}
```

Rollup name column then carries the suffixed name. `_flea_to_item` at `app/api/marketplace.py:823` must read by suffixed key:

```python
from src.store_naming import suffixed_name
stat = (stats or {}).get(suffixed_name(entity["name"], entity["owner_username"]), {})
```

Same pattern at `_load_inner_item_stats` and `_build_telemetry` call sites for flea (`app/api/marketplace.py:2536, 2588, 1830`).

### Option B (recommended) — Strip suffix at attribution time

Keep rollup keys in raw-name space, matching `store_entities.name` and the curated convention (`marketplace_plugins.name`). Single change in `MarketplaceItemLookup.resolve()`:

```python
if prefix == FLEA_BUNDLE_PREFIX:
    # Claude Code invokes flea skills/agents as agnes-store-bundle:<suffixed>
    # where <suffixed> = "<raw-name>-by-<owner-username>". The preload
    # indexes by raw name (matches store_entities.name everywhere else),
    # so strip the "-by-…" suffix before lookup. Owners with a literal
    # "-by-…" inside their entity name are blocked at upload time by
    # _NAME_RE in app/api/store.py, so a single rsplit is safe.
    raw_name = local.rsplit("-by-", 1)[0] if "-by-" in local else local
    ent_type = self._flea_entities.get(raw_name)
    if ent_type:
        return ("flea", "", raw_name, ent_type)
    continue
```

Same fix needed in the nested-flea-plugin branch (`usage_lib.py:323`) where `prefix in self._flea_plugins` — the prefix itself is a flea plugin name, but the **local** inside is a skill/agent that may also be suffixed. Inspect.

API read path stays unchanged. Backfill stays unchanged in shape (re-runs by source).

### Either way: test fixtures + assertions

- `tests/fixtures/sessions/usage/skill_flea.jsonl:2` — switch identifier to suffixed form (`agnes-store-bundle:flea-skill-by-alice`).
- `tests/test_session_processor_usage.py:203` — update assertion to look for suffixed identifier.
- New test: round-trip from `store_entities` insert (with `owner_username`) → fixture replay → rollup row → API response, verifying the suffix is invariant across the chain.

### Backfill

`usage_events` rows for flea invocations sitting at `source='builtin'` need re-attribution. `USAGE_PROCESSOR_VERSION` bump triggers the session-pipeline reprocess loop on next tick. `scripts/backfill_marketplace_rollup.py` rebuilds the rollup tables idempotently — run after deploy on each instance.

## Acceptance

- [ ] `tests/test_session_processor_usage.py::test_skill_flea` updated to use the suffixed identifier shape and still passes.
- [ ] New test pins the round-trip: insert entity with `owner_username`, replay fixture with `agnes-store-bundle:<suffixed>`, assert one row in `usage_marketplace_item_daily` with `source='flea'`, `type='skill'`, `parent_plugin=''`, `name=<raw>`.
- [ ] Manual verification on a dev VM with real flea invocations: `usage_marketplace_item_daily` shows `source='flea'` rows after one UsageProcessor tick.
- [ ] `/marketplace?tab=flea` renders `.inv-chip` on at least one card where prod invocations exist.

## Related

- PR #329 (telemetry v46 refactor) — fixed the prefix-split bug but didn't catch the suffix mismatch.
- PR #320 — original v46 design.
- CHANGELOG.md `## [0.54.21] — 2026-05-15` — full v48 rationale.

minasarustamyan:
## Summary

Seven-phase refactor of the Flea marketplace bringing the synthetic-plugin layer in line with the data model added in phase 1. Closes #335.

- **Phase 1 — data model** (`7f4cfcbb`): adds `title`, `tagline`, `synthetic_name` columns to `store_entities` (schema v49) with Python-side migration that backfills via humanize (acronym-aware) and the deterministic `<name>-by-<owner_username>` formula. Upload + edit forms gain a Title input, optional 200-char Short description, and a live synthetic-name preview.
- **Phase 2 — web presentation** (`5c83fa1a`): marketplace cards, plugin/skill/agent detail pages now render `title` (humanized) instead of the kebab-case slug, the owner's full name from `users.name` (with email → username fallback), and an optional `tagline` in the hero. Breadcrumb last segment + parent-plugin labels switch to title.
- **Phase 3 — single source of truth** (`eb608671`): read paths consume `store_entities.synthetic_name` directly; `suffixed_name()` survives only on write paths (POST create, PUT rename, archive). Collision SQL switches to `WHERE synthetic_name = ?`.
- **Phase 4 — rename** (`0a6cd67d`): synthetic plugin `agnes-store-bundle` → `flea`. ZIP/git tree path `plugins/store-bundle/` → `plugins/flea/`. Claude Code JSONL prefix becomes `flea:<synthetic_name>`. Clean cut — no legacy-prefix backward compat.
- **Phase 5 — attribution fix** (`c82fe2ce`, closes #335): `MarketplaceItemLookup` keys lookup tables by `synthetic_name` instead of `name`, which is what Claude Code actually writes after the `flea:` prefix. Pre-fix every flea invocation silently fell through to `source='builtin'`. Adds the flea-plugin-nested branch to `_attribute_event` (parity with the live resolver). `USAGE_PROCESSOR_VERSION` bumped 7 → 8 so the reprocess loop re-attributes historic events.
- **Phase 6 — rollup parity** (`badcb27d`): `_aggregate_events` plugin-level pass extended from `source='curated'` only to `source IN ('curated', 'flea')`, so flea plugin entities get a synthetic aggregated row summing nested skill/agent invocations. Without this, the plugin card and detail telemetry chip read 0 even though child rollup rows existed. `USAGE_PROCESSOR_VERSION` bumped 8 → 9.
- **Phase 7 — Details sidebar unification** (`997ff782`): five detail surfaces (curated/flea × plugin/inner/standalone) aligned on one scan order — identity → life-stage → telemetry → debug. Drops Slug (debug-only), Category + Installs (already in hero). Flea plugin Owner row reads the resolved fullname.

CalVer rollover note in CHANGELOG covers the breaking marketplace identifier rename and explains that `agnes refresh-marketplace` does not auto-uninstall orphaned plugins; the orphan `agnes-store-bundle@agnes` may linger client-side until Claude Code's marketplace update handles it or the user runs `claude plugin uninstall agnes-store-bundle@agnes` manually.

## Test plan

- [ ] `python3 -m pytest tests/ --tb=short -q` clean on CI (Linux)
- [ ] On dev VM after deploy: upload a flea skill with title "MCP Builder" and tagline → marketplace card renders title + tagline + owner fullname; detail page hero + sidebar follow the new layout
- [ ] Invoke a flea skill several times in Claude Code → next UsageProcessor tick attributes events as `source='flea'` (verify via `SELECT source, COUNT(*) FROM usage_events WHERE skill_name LIKE 'flea:%' GROUP BY source`)
- [ ] Plugin entity (`codex-second-opinion-by-…`) card shows sum of nested skill invocations once the plugin-level aggregation pass completes
- [ ] Last used + Active days populate on flea plugin detail sidebar after the 1-hour `last_30d` window refresh window passes
- [ ] No orphan `plugins/store-bundle/` on local marketplace clones after `agnes refresh-marketplace`; `flea@agnes` installs cleanly


minasarustamyan:
## Code review — critical findings

### 1. N+1 query in `_flea_to_item` (perf regression)

`app/api/marketplace.py:838` — `_resolve_owner_display(conn, entity["owner_user_id"], …)` is called for **every item** in the list comprehension at `app/api/marketplace.py:1092-1099`. The helper (line 1521) runs a separate `SELECT name, email FROM users WHERE id = ?` per call.

Pre-PR: 1 query (`StoreEntitiesRepository.list()`) + N × in-memory `entity["owner_username"]` read.
Post-PR: 1 query + N × `SELECT … FROM users`. Flea tab with 50 items = 51 queries instead of 1.

Suggested fix: prefetch users in one round-trip (`SELECT id, name, email FROM users WHERE id IN (…)`) before the comprehension, pass the resulting dict into `_flea_to_item` as `users_by_id`. Same pattern already used for `flea_stats` and `enriched_lookup` in the curated path.

### 2. Misleading comment in `src/marketplace_filter.py:189-191`

```
v49 phase-4: renamed from ``agnes-store-bundle`` to ``flea``. Historic
session JSONL events still carry the old prefix — attribution layer in
``services/session_processors/usage_lib.py`` accepts both.
```

Verified: `services/session_processors/usage_lib.py:254` has only `FLEA_BUNDLE_PREFIX = \"flea\"`, no fallback. `grep agnes-store-bundle services/session_processors/` returns no matches. The CHANGELOG explicitly states: **\"Clean cut — no legacy-prefix backward compat. Historic `usage_events` rows whose JSONL was written before the rename will stay attributed as `source='builtin'`.\"**

The comment contradicts both the code and the CHANGELOG. Either drop the second sentence, or rewrite it to match the actual clean-cut behavior.

### 3. CHANGELOG has two `### Changed` sections inside `## [Unreleased]`

Lines 9 and 61 of the diff: two separate `### Changed` blocks under the same unreleased header (Marketplace sidebar refactor vs. BREAKING rename + other changes). Keep-a-changelog convention is one section per category. At release-cut this will produce a malformed release block. Merge them into one `### Changed`.

### 4. `synthetic_name` has no DB-level UNIQUE constraint

`src/db.py:576+` declares `synthetic_name VARCHAR NOT NULL` but without `UNIQUE`. Phase-3 comments argue it's the \"single source of truth\"; phase-5 uses it as the keyspace for rollup attribution; `_suffixed_already_taken` queries it for collision checks.

The app-level guard (`_suffixed_already_taken`) only fires at upload/rename. Anything else — admin DB hand-fix, future bug in `archive()`, migration drift — can silently introduce a duplicate, and `WHERE synthetic_name = ?` will then non-deterministically return one row. Risk is materially higher post-v49 than pre-v49 (when `synthetic_name` was a runtime concat with no caching).

Suggested: add `UNIQUE (synthetic_name)` to `store_entities` (with the archive-rename already freeing the slug, this should hold) — or at minimum document that uniqueness is app-enforced only.

### 5. Stale test names asserting the wrong-version constant

- `tests/test_schema_v42_migration.py::test_schema_version_is_42` asserts `SCHEMA_VERSION == 49`
- `tests/test_schema_v46_migration.py::test_schema_version_is_46` asserts `SCHEMA_VERSION == 49`
- `tests/test_home_stats.py::test_schema_version_constant_is_48` asserts `SCHEMA_VERSION == 49`

Three places where the test name lies about what it tests. This is a pre-existing pattern this PR perpetuates — every future schema bump touches all three. Either rename to version-agnostic (`test_schema_version_matches_constant`) or accept the rename churn each bump. Worth a small follow-up cleanup; not blocking for this PR.

---

No RBAC or security issues found. Migration is idempotent and fresh-install safe (verified by `test_v48_to_v49_function_is_idempotent`). SyncOrchestrator path unaffected.

PR #372, opened 2026-05-21
feat(web): instance.custom_scripts — operator-injected HTML/JS into base.html

cvrysanek:
## Summary

- Add `instance.custom_scripts` — a YAML list of named, placement-aware
  HTML/JS blocks rendered by `base.html` and `base_login.html`. Each
  entry takes `name`, `enabled`, `placement` (`head_start | head_end |
  body_end`), and `html`.
- Generic mechanism — no per-vendor helpers. First user is Marker.io
  (sample block ships commented out in `config/instance.yaml.example`),
  but the same path serves Sentry, GTM, PostHog, Hotjar, cookie banners,
  etc.
- Wired into `_render_ctx` so templates see `custom_scripts` in scope;
  surfaced in `/admin/server-config` via `_KNOWN_FIELDS["instance"]`.
- Trust boundary mirrors existing `instance.logo_svg` /
  `instance.overview` — admin-only, rendered with `| safe`. Same
  precedent already lets admins inject raw `<svg>` and HTML; this
  generalises it with placement control + per-block enable/disable +
  audit-logged saves.

## Why this shape

Considered (and rejected) a per-vendor `instance.marker_io_project_id`
field: one helper per vendor doesn't scale, and the OSS shouldn't
hardcode third-party brands. A single list with placement metadata
covers every paste-a-snippet vendor.

Env-var override deliberately skipped — list-of-objects doesn't fit
env-vars cleanly, and the deploy-time injection pathway lives in the TF
repo where instance.yaml is rendered per-VM (a follow-up PR there wires
Marker.io to the production VM only).

## Files

- `app/instance_config.py` — new `get_custom_scripts()` helper:
  normalises defaults, drops disabled / empty / bad-placement entries,
  logs warnings instead of crashing on operator typos.
- `app/web/router.py` — adds `custom_scripts` to `_render_ctx` (one
  line + one import).
- `app/web/templates/base.html` + `base_login.html` — three placement
  loops each (`head_start`, `head_end`, `body_end`). `base_login.html`
  mirrors `base.html` so login/auth pages also surface the snippet
  (this gap was caught by the render test).
- `app/api/admin.py` — `_KNOWN_FIELDS["instance"]` entry so the field
  shows up in `/admin/server-config`. Renders as JSON textarea (the
  existing fallback for `kind: array`).
- `config/instance.yaml.example` — commented-out Marker.io example
  block under the `instance:` section.
- `CHANGELOG.md` — `[Unreleased] → Added` bullet.

## Test plan

- [x] `pytest tests/test_instance_config.py tests/test_custom_scripts_render.py` — 26/26 pass.
  - 8 new `TestCustomScripts` cases cover: yaml-absent, valid entry, disabled, empty html, bad placement, default placement, three placements, non-list value.
  - 5 new render tests cover: empty list, each placement landing in the right slot, all three placements in order.
- [ ] Smoke locally: set `instance.custom_scripts` with a real Marker.io
  block, hit `/` and `/login`, confirm widget overlays the page and
  capture flow works.
- [ ] Admin UI smoke: `/admin/server-config` → `instance` section →
  `custom_scripts` field renders as JSON textarea, save persists,
  restart banner shows.

## Known pre-existing failures (NOT from this diff)

The full suite hits a pre-existing `_v51_to_v52` schema-migration error
in `src/db.py:3531` (`Cannot alter entry "knowledge_items" because there
are entries that depend on it`) which cascades into ~1700 setup errors
on any fixture that calls `get_system_db()`. Confirmed via `git stash` —
same failure reproduces on the clean branch tip with my diff removed.
Not blocking this PR; needs a separate cleanup of the migration ladder
or a local DuckDB reset.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/372" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/app/instance_config.py
+++ b/app/instance_config.py
@@ -367,6 +367,29 @@ def get_instance_overview() -> str:
 _CUSTOM_SCRIPT_PLACEMENTS = ("head_start", "head_end", "body_end")
 
 
+def _custom_script_enabled(value) -> bool:
+    """Coerce the per-entry ``enabled`` field tolerant of YAML's many
+    truthiness shapes.
+
+    Operators hand-editing YAML (or pasting blocks from another source)
+    can land ``enabled: "false"`` (quoted string), ``enabled: 0``, or
+    ``enabled: no`` rather than the boolean ``false``. ``bool("false")``
+    is ``True`` in Python, so a naive truth check silently keeps the
+    script live — a footgun for what's meant to be a kill switch on
+    admin-injected JS. Missing / ``None`` → live (default-on, matches
+    the registered field shape).
+    """
+    if value is None:
+        return True
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, (int, float)):
+        return value != 0
+    if isinstance(value, str):
+        return value.strip().lower() not in ("", "false", "no", "0", "off")
+    return bool(value)
+
+
 def get_custom_scripts() -> list[dict]:
     """Operator-injected HTML/JS blocks rendered by ``base.html``.
 
@@ -387,7 +410,9 @@ def get_custom_scripts() -> list[dict]:
     ``instance.logo_svg`` / ``instance.overview``.
 
     Normalization:
-    - Drop entries with ``enabled=False``.
+    - Drop entries whose ``enabled`` resolves to false via
+      :func:`_custom_script_enabled` (handles quoted strings, 0/1, etc.
+      — not just the Python ``False`` singleton).
     - Drop entries whose ``html`` strips to empty.
     - Default missing ``name`` to "" and missing ``placement`` to
       "head_end".
@@ -419,7 +444,7 @@ def get_custom_scripts() -> list[dict]:
                 idx, type(entry).__name__,
             )
             continue
-        if entry.get("enabled") is False:
+        if not _custom_script_enabled(entry.get("enabled")):
             continue
         html = (entry.get("html") or "").strip()
         if not html:
--- a/app/instance_config.py
+++ b/app/instance_config.py
@@ -364,6 +364,84 @@ def get_instance_overview() -> str:
     return (raw or "").strip()
 
 
+_CUSTOM_SCRIPT_PLACEMENTS = ("head_start", "head_end", "body_end")
+
+
+def get_custom_scripts() -> list[dict]:
+    """Operator-injected HTML/JS blocks rendered by ``base.html``.
+
+    Reads ``instance.custom_scripts`` from instance.yaml — a list of
+    dicts ``{name, enabled, placement, html}``. Each block lands in one
+    of three template slots:
+
+    - ``head_start`` — first thing in ``<head>``, before any CSS/JS
+      (rare; GTM dataLayer init).
+    - ``head_end`` — last thing in ``<head>`` (default; analytics +
+      feedback widgets like Marker.io, Sentry, Hotjar).
+    - ``body_end`` — just before ``</body>`` (vendors that explicitly
+      ask for bottom placement).
+
+    Trust boundary: admin-only. ``instance.yaml`` is written through
+    ``/api/admin/server-config`` (gated by ``require_admin``) and the
+    rendered HTML is interpolated with ``| safe``, exactly mirroring
+    ``instance.logo_svg`` / ``instance.overview``.
+
+    Normalization:
+    - Drop entries with ``enabled=False``.
+    - Drop entries whose ``html`` strips to empty.
+    - Default missing ``name`` to "" and missing ``placement`` to
+      "head_end".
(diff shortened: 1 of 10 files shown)


app/web/templates/base.html
```diff
@@ -1,6 +1,12 @@
 <!DOCTYPE html>
 <html lang="en" data-theme="{{ instance_theme | default('navy') }}">
 <head>
+    {# Operator-injected scripts (placement=head_start) — run before any
+       CSS/JS so vendors that need to install global hooks first (GTM
+       dataLayer init, etc.) work. Admin-only, see instance.custom_scripts. #}
+    {% for s in custom_scripts | default([]) if s.placement == 'head_start' %}
```
minasarustamyan, line 14: The `head_start` loop renders before `<meta charset="UTF-8">`, which is risky:

- **HTML5 spec**: the charset declaration must appear within the first 1024 bytes of the document. A long operator-pasted snippet (e.g. a full GTM container init) can push `<meta charset>` past that window, at which point browsers fall back to a default encoding (often `windows-1252` on EU/CZ locales). That breaks rendering of non-ASCII content, and historically opened the door to UTF-7 charset-confusion XSS.
- **No real use-case for pre-charset placement**: the docstring justifies this slot as "rare; GTM dataLayer init", but `dataLayer` only needs to run before the GTM loader script — not before the charset declaration. I don't know of an analytics or feedback vendor that requires execution before `<meta charset>`.

Suggested fix: move the `head_start` loop after `<meta charset="UTF-8">` (and ideally after `<meta name="viewport">`). The "first thing in `<head>`" intent still holds for vendor snippets; we just respect the two required `<meta>` tags first. Same change in `base_login.html`.

app/instance_config.py
```diff
@@ -364,6 +364,84 @@ def get_instance_overview() -> str:
...
+        return []
+    out: list[dict] = []
+    for idx, entry in enumerate(raw):
+        if not isinstance(entry, dict):
+            logger.warning(
+                "instance.custom_scripts[%d] must be a dict, got %s — skipping",
+                idx, type(entry).__name__,
+            )
+            continue
+        if entry.get("enabled") is False:
```
minasarustamyan: `is False` only matches Python's `False` singleton, so stringified booleans pass this filter silently:

- `enabled: "false"` (quoted string — easy to introduce when hand-editing YAML or pasting a snippet block from another source) → `entry.get("enabled")` returns the string `"false"`, which is not `False`, so the entry is **not** dropped and the script runs.
- Same for `enabled: 0`, `enabled: "no"`, etc.

For the kill-switch on admin-injected JS this seems worth hardening — an operator who toggles `enabled: false` in a panic but mistypes the quoting would silently leave the snippet live.

Note that `if not entry.get("enabled", True):` alone wouldn't fix it either: `bool("false") == True` in Python, so a quoted `"false"` still slips through. A small coercion helper is the cleanest path:

```python
def _enabled(v) -> bool:
    if isinstance(v, str):
        return v.strip().lower() not in ("false", "no", "0", "off", "")
    return True if v is None else bool(v)

if not _enabled(entry.get("enabled")):
    continue
```

At minimum the current `is False` check should be replaced so the operator's intent on quoted booleans matches their expectation.

app/web/templates/base.html
cvrysanek: Good catch — fixed in d12db17 (well, the next commit on this branch).

Moved both `head_start` loops (`base.html` + `base_login.html`) to render **after** `<meta charset="UTF-8">` and `<meta name="viewport">` so the charset declaration is guaranteed to land in the first 1024 bytes regardless of how long an operator's snippet is. The "vendor hook installs before any CSS/JS" intent still holds — just the two required `<meta>` tags come first now.

Render test updated to enforce the invariant: `test_head_start_snippet_lands_after_charset_before_first_link` now asserts `charset_idx < snippet_idx < first_link_idx`.

app/instance_config.py
cvrysanek: Agreed — `is False` was too tight. Fixed.

Added `_custom_script_enabled()` coercion helper that handles every YAML truthy-shape an operator might paste:

```python
def _custom_script_enabled(value) -> bool:
    if value is None:
        return True  # default-on, matches the registered field shape
    if isinstance(value, bool):
        return value
    if isinstance(value, (int, float)):
        return value != 0
    if isinstance(value, str):
        return value.strip().lower() not in ("", "false", "no", "0", "off")
    return bool(value)
```

Now `enabled: "false"`, `enabled: 0`, `enabled: "no"`, `enabled: "off"`, `enabled: "0"` all drop the entry — operator's mental model of the kill switch matches the actual behaviour regardless of how they typed it.

13 parametrized cases cover the truthy-shape matrix in `TestCustomScripts::test_enabled_coercion`.

PR #388, opened 2026-05-25
feat(db): introduce Postgres app-state layer alongside DuckDB

cvrysanek:
## TL;DR

Adds the Postgres app-state foundation alongside DuckDB: 28 mirror repositories at `src/repositories/*_pg.py`, 10 Alembic migrations with round-trip + drift discipline, a one-shot DuckDB → Postgres data migration framework, and the Docker Compose overlay + env wiring needed to actually run against a real database. **163 PG tests pass; 65 existing DuckDB tests still green.**

## Honest framing — what this PR delivers vs. doesn't

**Delivered (storage + ops layer):**

| Layer | Status |
|---|---|
| Postgres schema parity | ✅ 100% — 31 tables, every column / FK / index / check constraint mirrored |
| PG repository implementations | ✅ 28/28 repos at `src/repositories/*_pg.py` |
| Migration chain (Alembic with downgrades) | ✅ 10 revisions, round-trip + drift + pairwise tests |
| Data migration framework | ✅ 31 `MigrationTask` entries, idempotent, checksum-validated |
| Test coverage on the new PG layer | ✅ 163 PG tests pass; cross-engine contract on `audit` |
| Docker Compose Postgres overlay | ✅ this PR |
| Production wiring docs (Cloud SQL + Unix-socket) | ✅ this PR |

**Delivered in subsequent commits on this PR:**

| Layer | Status | Commits |
|---|---|---|
| Factory pattern in `src/repositories/__init__.py` | ✅ | `2c5309f6a` |
| Callsite swap (41 app/api + 8 app/auth + 5 services + 2 cli files) | ✅ | `015b5d2b6` → `bcda84b9e` |
| Direct `*_pg` imports in `app/`, `services/`, `cli/` | ✅ zero (factory-contained) | grep verified |
| Cleanup of dev-helper sed scripts | ✅ this cleanup pass | follow-up commits |

The factory consults `AGNES_DB_URL` per-call: env unset → DuckDB legacy repos; env set → Postgres repos. No mixed-engine requests possible within a single request lifecycle (all factories read the same env var).

The cleanest cutover path is a factory in `src/repositories/__init__.py` that picks the backend by `AGNES_DB_URL` presence — that change can land in a follow-up PR without touching this one. The goal "DuckDB only for analytical" is met at the **storage layer** but not yet at the **runtime**; both halves are needed.

## Why this, why now

[...]

PR #454, opened 2026-05-27
[DRAFT] Re-do PR #388 — Postgres app-state layer (debugging auto-upgrade failure)

ZdenekSrotyr:
Restores the 3 squashed commits from PR #388 + applies the root-cause fix for the auto-upgrade failure that triggered the revert.

## Root cause (identified via local docker-compose test)

PR #388 added a 'data-migrate' service volume override to `docker-compose.host-mount.yml`:

```yaml
data-migrate:
  volumes: !override
    - /data:/data:ro
```

But 'data-migrate' is defined **only** in `docker-compose.postgres.yml`. The `agnes-auto-upgrade.sh` script on customer-instance VMs (including foundryai-development) uses the default baseline chain when `.env` doesn't set `COMPOSE_FILE`:

```
docker-compose.yml + docker-compose.prod.yml + docker-compose.host-mount.yml
```

Without `postgres.yml` in that chain, the host-mount reference to 'data-migrate' → `docker compose config` fails with:

```
service "data-migrate" has neither an image nor a build context specified: invalid compose project
```

→ exit 1 every 5-min auto-upgrade tick → tracked as issue #452.

## Fix (commit `e94975ee`)

1. **Remove** the data-migrate block from `docker-compose.host-mount.yml`. File is now standalone-loadable again.
2. **Create** `docker-compose.postgres-host-mount.yml` (new bridge overlay) that restates the `data-migrate` mount as `/data:/data:ro`. Used ONLY when both `postgres.yml` AND `host-mount.yml` are loaded.
3. **Wire**: `scripts/ops/agnes-auto-upgrade.sh` pulls the new file in CONFIG_FILES; `infra/modules/customer-instance/startup-script.sh.tpl` extends COMPOSE_FILE to include it.

## Verification

- Local: `docker compose -f base -f prod -f host-mount config` (= default baseline) now validates cleanly (would've failed before this PR with the data-migrate reference). Full chain with postgres + bridge also validates.
- agnes-dev snapshots taken before deploy (`agnes-dev-boot-pre-pg388redo-20260527-1659`, `agnes-dev-data-pre-pg388redo-20260527-1659`)
- Tests: 322 PG suite + 5281 DuckDB suite passed locally during the original PR #388 cycle; this PR adds only deploy-side wiring changes (no Python code).

## Plan

- Ready for review once CI passes
- Deploy to **agnes-dev** (Keboola Pattern A — consuming `infra/modules/customer-instance/` so Secret Manager + startup-script changes get applied) as smoke test
- Then merge to main when dev VM auto-upgrade.service stays green for at least 3 ticks


ZdenekSrotyr:
Consolidated into #455 — that PR is now retargeted to `main` and contains all commits from this branch (`zs/postgres-redo`) plus the state-machine work on top. Closing this one to avoid dual review.

PR #455, opened 2026-05-27
feat(db): Postgres app-state + admin-controlled state machine

ZdenekSrotyr:
Two-layer change consolidating what was previously stacked as #454 + #455:

1. **Postgres app-state layer** — the re-do of merged-then-reverted #388 + the Dockerfile / compose bridge fixes that caused the original revert.
2. **Admin-controlled state machine** — the spec at `docs/superpowers/specs/2026-05-27-db-backend-state-machine-design.md`. Replaces ad-hoc `.env` editing with `/admin/server-config` UI + `agnes admin db` CLI for migrating Agnes app-state between DuckDB, side-car Postgres, and managed cloud Postgres.

## ✅ Full state-machine path verified live on agnes-dev

Both transitions exercised against real production data:

### DuckDB → side-car PG (v7)
```
rows_total: 4787  tables_migrated: 50  status=success
```

### side-car PG → Cloud SQL (v9)
```
rows_total: 5145  tables_migrated: 50  status=success
backend: cloud
url: postgresql+psycopg://postgres:****@<cloud-sql-ip>:5432/agnes
```

After the cloud cutover, the local side-car postgres container + volume were removed; app + scheduler are healthy on Cloud SQL with no on-VM database.

## Architecture

- **API endpoint** writes only intent — pending job JSON + flag file. Returns 202.
- **Host applier daemon** (agnes-state-applier.timer, 30s tick) detects pending job, brings up the postgres container (when target is side-car), stops app+scheduler (releasing the DuckDB file lock), runs the migrator via `docker run --rm -v /data:/data <image>` from the host, then on success flips instance.yaml::database.backend to the new value.
- **Migrator** (scripts.db_state_migrator) dispatches per transition:
  - `duckdb → side_car`: alembic upgrade + copy_duckdb_to_pg + verify + backup duckdb + flip
  - `side_car → cloud`: alembic upgrade + pg_dump source + copy_pg_to_pg + verify_pg_row_counts + flip

## Iteration journey (v2 → v9)

Each round added unit/integration tests AND surfaced the next layer of real-world friction. All nine blockers found by live agnes-dev verification, none caught by the green PG/DuckDB suites at PR time.

| Ver | Blocker | Fix commit |
| --- | --- | --- |
| v2 | Dockerfile didn't bake postgres compose overlays into `/opt/agnes-host/` | `1c6be4be` |
| v3 | DuckDB file lock conflict on migrator subprocess | `4dc6a864` |
| v3 | PG models behind DuckDB v59 schema (resource_grants.requirement etc.) | `ce06b08c` |
| v4 | Request-scoped cursor kept lock alive after singleton close | `85bab4bb` |
| v5 | In-process Python cannot release DuckDB lock at all while uvicorn lives | `b00634ba` (host-orchestrated architecture) |
| v6 | ARRAY(String) columns failed on DuckDB JSON-string values | `ee6442a5` |
| v6 | shellcheck SC2034 in applier wait loop | `ff01bea1` |
| v7 | NOT NULL columns with server_default failed when DuckDB row had NULL | `c42687e6` |
| v8 | side_car → cloud branch reused copy_duckdb_to_pg (re-copied stale DuckDB into cloud) | `2c11461f` |
| v9 | JSONB CAST rejected bare strings on PG → PG copy | `df1a587f` |

Plus `e5fa2acb` ci-speedup: keboola-deploy.yml's pytest is now `-n auto` (5 min instead of 25).

## Test coverage

- PG suite: 338 passed
- DuckDB suite: 5329 passed (per the v9 CI run)
- New regression tests: cross-process lock release, schema-parity drift, ARRAY column coerce, NOT NULL default substitution, host-applier shell integration, PG→PG copy, source-url validation

## Closed PRs subsumed

- #454 (zs/postgres-redo) — the Postgres app-state re-do; commits squashed-merged into the base of this branch.

## Status

Both state-machine transitions live-verified. agnes-dev currently running on Cloud SQL (db-f1-micro, europe-west1). Operator can roll back at any time through the same admin endpoints. Ready to leave draft on review approval.

## Review fixes (from PR review comments)

Two review passes from @cvrysanek surfaced 9 BLOCKERs, 4 HIGHs, plus
a long tail of MEDIUMs and testing gaps. All addressed across 9 phases.
Live PG suite: **460 passed, 1 skipped**.

### BLOCKERs (data corruption / destructive paths)
- **B1** cancel_job reverts to source_backend, not target → `b1f0d73e`
- **B2** cooperative cancellation sentinel — applier observes mid-migration cancel → `e0f2d4dd`
- **B3** verify_row_counts raises on missing target table instead of silent OK → `d086a460`
- **B4** write_backend_state preserves URL by default — eliminates 30s no-URL window → `0cd65572`
- **B5** stuck-running recovery via heartbeat (>120s alive = auto-fail) → `598c0da5`
- **B6** applier write_instance_yaml preserves operator-set non-database keys → `32a936a9`
- **B7** URL alias detection blocks self-target migrations (default port, case, creds, driver) → `a382d2f7`
- **B8** _current_job_id surfaces pending jobs; flock held until JSON durable → `da716ca0`
- **Migrator CLI** exits non-zero on per-table failure → `82a190b1` (+ `bb01de2d` cascade)

### HIGHs (auth / secret leak)
- **H1** GET /api/admin/db/job/{id} redacts target_url + source_url → `5e8e2cc2`
- **H2** instance.yaml + job JSON written with mode 0600 (creation + cancel rewrite) → `b14a390a`
- **H3** cloud_url validated server-side (scheme, host, database) → `e9a47722`
- **H4** POSTGRES_PASSWORD missing → 500 with actionable message (no silent default) → `e5ab0cca`

### Migrator hardening (Phase 1)
- Probe-then-raise on DuckDB columns missing in PG with data → `0fab56e1`
- Preserve operator-supplied timestamps in copy → `361ebdba`
- Backup DuckDB BEFORE copy step (not after verify) → `b3d00668`
- Bounded engine with connect + 5-min statement timeouts → `1b7e8831`

### Applier hardening (Phase 3)
- Pending-job FIFO ordering by mtime → `fe497d0c`
- Flag cleared on duckdb rollback path → `fe497d0c`
- pg_isready loop tracks success / aborts on 60s timeout → `fe497d0c`
- Single-pass JSON parse (1 python3 invocation instead of 5) → `fe497d0c`
- RESTART_RC check reachable under set -e → `fe497d0c`

### Phase 4 — applier liveness
- `applier_last_tick_age_s` field in GET /api/admin/db/state → `a73b9c79`

### Phase 5 — UI polish
- localStorage progress cache (eliminates empty-yellow-box during applier restart) → `00b9037e`
- Exponential polling backoff 2s→4s→8s→cap 30s → `00b9037e`
- Per-table progress (current_table, tables_done/total) rendered → `00b9037e`

### Phase 6 — CLI
- `agnes admin db migrate --yes/-y` confirmation gate; refuses non-TTY without it → `0c57e8ed`

### Phase 7 — Tests (coverage gap)
- JSONB string-coercion round-trip → `23661e7a`
- Mid-data_copy cancel boundary check → `23661e7a`
- Module-scoped alembic fixture (PG suite speedup) → `23661e7a`
- E2E DUCKDB→CLOUD direct + CLOUD→SIDE_CAR DR smoke + host-reboot recovery → `e30f4bca`
- Hung-migrator statement_timeout + admin auth-bypass runtime probe → `e31f016f`

### Phase 8 — Security
- agnes-state-applier runs as non-root `agnes-applier:docker` on new VMs (existing VMs see runbook migration) → `34e20919`

### Phase 9 — Docs
- Cutover runbook: cancel semantics, PNR boundary, recovery procedures, URL alias gotcha, redaction → `90d6aa32`
- CHANGELOG `[Unreleased]` updated with grouped bullets → `90d6aa32`

---

## Review fixes — round 2

Closes every remaining HIGH / MEDIUM / LOW from cvrysanek's review. Each task → commit SHA.

### Phase A — Migrator hardening
- **A.1** `copy_pg_to_pg` streams in 500-row batches via `yield_per`; per-batch `target.begin()` (H4) → `157b4d80`
- **A.2** subprocess.run timeouts on alembic + gzip + pg_dump (H5) → `4c89bc5d`
- **A.3** `run_all` halts on first per-task failure; no more orphan rows downstream (H6) → `eae3409c`
- **A.4** content-hash check in `verify_pg_row_counts` — catches preseed corruption with matching PKs (H12) → `7a7a5ded`

### Phase B — Data integrity + ops
- **B.1** pre-copy scrub of historical audit_log PII (H7) → `e4b66441`
- **B.2** pending-job age expiry — applier marks pending older than `PENDING_JOB_MAX_AGE_SEC` (default 1h) as `PendingJobExpired` (H8) → `def04628`
- **B.3** cloud→side_car failure rollback also clears `db-state-target.flag` → `d1ad459b`
- **B.4** side_car→cloud backup failure is a hard fail (`BackupError`); was silently swallowed → `a07e60da`

### Phase C — Progress wiring + applier watchdog
- **C.1** `run_all` gains `progress_callback`; `copy_*` gain `writer=`; `update_table_progress` actually fires → `22f05dc5`
- **C.2** `timeout(1)` wraps `docker run` in the applier; rc 124/137 marks the pending failed (H5 applier side) → `62e2db69`

### Phase D — Tests for uncovered branches
- **D.1** `_substitute_default` parametrised over NOW() / CURRENT_DATE forms → `2495fc79`
- **D.2** PG→PG round-trip with PG ARRAY + JSONB dict columns → `2495fc79`
- **D.3** python-side hang-watchdog E2E (`subprocess.run(timeout=...)` on the migrator CLI) → `2495fc79`

### Phase E — LOW cleanups
- **E.1** `verify_row_counts` opens DuckDB read-only; no stray `.wal` sidecar → `eec0f130`
- **E.2** applier shell tests use semantic keyword matchers instead of brittle argv-substring asserts → `5ae710b0`
- **E.3** `resource_grants.resource_id` FK — closed-out with rationale on the ResourceGrant docstring (polymorphic FK impractical) → `f194dc0a`
- **E.4** `test_post_migrate_does_not_spawn_subprocess` kept — re-audit found it IS a real probe (monkey-patches Popen + run) → `752d7c18`
- **E.5** applier ERR trap with structured rollback (pending → failed, instance.yaml → source, FLAG cleared) → `da2599a6`

### Phase F — Release-cut
- Schema-version assertion revert (v59→v60 hadn't landed on this branch; PR #458 carries that on a separate stream) → `d296de7b`
- 0.56.0 release-cut (CHANGELOG + pyproject) → `71ae31c2`

Full PG-backed test suite after round 2: see thread comment with the green run.

---

## Round-2 review fixes shipped (cvrysanek, 2026-06-01)

Round-2 adversarial review surfaced 5 BLOCKER + 8 HIGH + 4 MED + 2 LOW + 1 live-discovered finding. Plan committed in `da4c5875`; execution via `superpowers:subagent-driven-development` with TDD discipline per task (failing test → minimal fix → green) + two-stage spec-compliance + code-quality review.

**All 20 findings shipped on this branch.** Full suite at Phase H checkpoint: **5903 passed / 36 skipped / 0 failed** (+30 from pre-round-2 baseline).

| Finding | Status | Commit |
| --- | --- | --- |
| B1-NEW concurrent `/migrate` race | FIXED | `dd9516e7` |
| B2-NEW `_urls_alias` DNS bypass | FIXED + applier guard | `9ea5d433` |
| B3-NEW applier `.env` permission | FIXED + provisioning chown | `6b145bbc` |
| B4-NEW `/data/postgres` chown | FIXED + provisioning | `551fca78` |
| B5-NEW alembic 0013 ordering | FIXED | `b2fb61e1` |
| H1-NEW cancel-during-verify race | FIXED | `d066ffdf` |
| H2-NEW heredoc 0600 mode | FIXED | `c5cd5b53` |
| H3-NEW `error.message` redaction | FIXED + migrator raise site | `59933f10` |
| H4-NEW PyYAML missing fallback | FIXED + apt-install | `96d7dc2f` |
| H5-NEW stuck recovery backend restore | FIXED | `2bf20321` |
| H6-NEW `_JSON_COLUMNS` dynamic derivation | FIXED (caught 7 drifted JSONB cols) | `bcaf40b2` |
| H7-NEW `POST /migrate target=duckdb` 501 | FIXED | `96b7a382` |
| H8-NEW `__rollback` + failed-branch URL drop | FIXED (both sites) | `0c9b343c` |
| MED-1 `--json` bypass `--yes` confirm | FIXED | `e6569c59` |
| MED-2 `cloud_url` SSRF reserved ranges | FIXED | `46334442` |
| MED-3 redact query-string password | FIXED | `a4a57a48` |
| MED-4 cancel revert drops PG url for duckdb | FIXED | `5001fab9` |
| LOW-1 PII scrub walks JSON keys | FIXED | `ed580574` |
| LOW-2 compose drops `POSTGRES_PASSWORD:-agnes` | FIXED | `e1da2980` |
| NEW-X DuckDB→PG users `ON CONFLICT DO NOTHING` | FIXED (live-discovered 2026-06-01) | `2169a106` |

Notable bonus catches surfaced during execution:
- **H6-NEW** dynamic JSONB derivation surfaced **6 additional** previously-missing columns beyond the spec's `data_packages.tags`: `data_packages.when_to_use`/`when_not_to_use`/`example_questions`, `recipes.related_table_ids`, `table_registry.sample_questions`/`pairs_well_with`. Forward-compatibility test (`test_json_columns_covers_every_jsonb_in_metadata`) is now a self-improving regression guard.
- **H8-NEW** extended mid-task to cover **two** identical-class sites: the `__rollback` ERR-trap path AND the failed-migration `else` branch in the main loop. Generalised test asserts every `write_instance_yaml "$SOURCE_BACKEND"` line contains `SOURCE_URL` — forward-compatible against future revert sites.

Detail in [`CHANGELOG.md`](./CHANGELOG.md) under `[Unreleased]` → `Round-2 review fixes — verification matrix`.

---

## Round-3 review fixes shipped (cvrysanek, 2026-06-01 16:45Z)

Round-3 adversarial review surfaced 3 NEW BLOCKER + 1 NEW HIGH + 2 PARTIAL (one HIGH, one MED). All 6 findings have a fix commit + regression test on this branch under TDD discipline + sub-agent execution (`superpowers:subagent-driven-development`).

Plan committed in `f241d62a` at `docs/superpowers/plans/2026-06-01-pr455-round-3-fixes.md`. Full suite at release-cut: **5986 passed / 36 skipped / 0 failed** (+18 from 0.57.0 baseline).

| Finding | Status | Commit | Test |
| --- | --- | --- | --- |
| MED-1-PARTIAL hostname → reserved IP | FIXED (DNS classification) | `0e981897` | `tests/test_db_state_url_resolve_validation.py` |
| B1-NEW DNS rebinding TOCTOU | FIXED (pin IP at queue time) | `d54610ba` | `tests/test_db_state_pinned_ip.py` |
| B2-NEW bash YAML fallback escaping | FIXED (`printf` + sed escape + loud log) | `27237d21` | `tests/test_applier_yaml_fallback_escaping.py` |
| B3-NEW stuck-recovery service restart | FIXED (`dc up -d app scheduler` post-revert) | `91a2dd84` | `tests/test_applier_recovery_restarts_services.py` |
| H1-NEW PG-source with null url | FIXED (refuse-to-queue 400) | `c1a34b6f` | `tests/test_db_state_refuse_null_source_url.py` |
| H1-PARTIAL cancel ↔ flip atomic | FIXED (`MigrationLock` both sides + retry) | `b5061af4` | `tests/test_db_state_cancel_flip_atomic.py` |

Notable beyond-spec improvements during execution:
- **MED-1-PARTIAL**: extracted `_classify_reserved_ip` helper (shared with B1-NEW pinning).
- **H1-PARTIAL**: added re-read under lock in `cancel_job` to catch race where migrator completed (or moved past PNR) between pre-lock status check and lock acquisition — closes a second-order race the spec didn't explicitly call out.
- **B2-NEW**: discovered spec's example URL (`?application_name=a:b`) does NOT actually trigger YAMLError — bare colons in query strings parse fine. The real trigger is `: ` (colon-space, block-style key separator). Test uses `options=-c replication: logical` instead.

**Residual callouts deferred to future PRs** (per "Issue economy" — they didn't meet the ≥3× scope criterion):
- `SupplementaryGroups=docker` ≈ root via docker socket (runbook flag)
- Unconditional `chown 70:70 /data/postgres` on disk reuse
- H4 bash YAML fallback drops non-database top-level keys (WARN-only)
- NEW-X bare `ON CONFLICT DO NOTHING` silent-drop semantics
- `AGNES_ALLOW_RESERVED_CLOUD_URL=1` env-leak hardening

Version: `0.57.0` → `0.57.1` (patch — round-3 fixes are bug fixes for round-2 fixes that were released as 0.57.0).


cvrysanek:
## Comprehensive review of PR #455

Ran 10 parallel reviewers (Agnes architecture/RBAC/rules; CE correctness/security/data-integrity/reliability/testing; data-migrations stalled at 600s watchdog; Codex adversarial). Findings deduped and ranked below.

**TL;DR — verdict:** Functionally live-verified on agnes-dev, but the state-machine has multiple **BLOCKER**-class correctness and security gaps that the live verification did not exercise (single-operator happy path, small row counts, no concurrency). 8 BLOCKERs, 12 HIGHs, 18 MEDs, 8 LOWs across the new surface. **Do not merge** without addressing the BLOCKERs at minimum.

**Clean / no findings:** RBAC (4 new admin endpoints + web route + CLI all gated; no `ResourceType` regression). Project conventions (CHANGELOG bullets grouped + descriptive, vendor-agnostic, no AI attribution, clean commit history). Architecture invariants for extract.duckdb / orchestrator / connectors (untouched by this PR — orthogonal). Several attack-surface classes verified closed: SQL identifier injection in copy paths (metadata-driven), `docker run` argv shell expansion, `/job/{id}` path traversal, `pg_dump` argv credential.

---

### BLOCKERS — fix before merge

**B1. `cancel_job` revert uses `target_backend` instead of `source_backend`** — `app/api/db_state.py:246-247`
v6 logic that assumed forward-only paths. Cancel of `cloud→side_car` (target=`side_car`) reverts to **DUCKDB** while live data is on Cloud SQL — instance.yaml diverges from reality. Pending job already records `source_backend`; use it.
*(correctness BLOCKER, data-integrity CRITICAL #1, codex HIGH, security LOW)*

**B2. Cancel cannot stop the running migrator → terminal write overwrites `cancelled`** — `app/api/db_state.py:213-249`
Cancel only rewrites the job JSON. Migrator container keeps copying; `JobWriter.mark_success` then overwrites `cancelled` with `success`. State-flip race. No cooperative cancellation check inside migrator, no SIGTERM, no `<id>.cancel` sentinel. Data already landed in target.
*(data-integrity CRITICAL #2, correctness, reliability HIGH, codex HIGH)*

**B3. Row-count verifier swallows missing target tables** — `scripts/db_state_migrator.py:327-335,377`
Catches `sa.exc.ProgrammingError` and substitutes `tgt_count=0`. If alembic partially applied AND DuckDB-side empty (fresh installs), reports `0=0` "match" on broken schema. Hides typo'd table names.
*(data-integrity CRITICAL #3)*

**B4. `POST /migrate` wipes `database.url` from instance.yaml + repository routing switches to PG before PG is usable** — `app/api/db_state.py:163`, `src/repositories/__init__.py:99`
`write_backend_state(in_progress)` called without `url` arg deletes the existing url key (`src/db_state_machine.py:122`). Repository routing already treats `*_in_progress` as Postgres. So for the 30s window between API write and applier tick — or longer if timer masked — `/api/admin/db/job/<id>` and every authenticated request can fail. Crash mid-flight → app comes up with no URL.
*(correctness HIGH, codex BLOCKER)*

**B5. SIGKILL/OOM after job becomes `running` → app permanently stopped** — `scripts/db_state_migrator.py:479`, `scripts/ops/agnes-state-applier.sh:69,146`
Applier only picks up `status="pending"`. Migrator writes `running` before any work. Host reboot / docker daemon restart / `kill -9` of the applier shell mid-migration leaves the job stuck in `running` forever; next applier tick has nothing to do; app+scheduler never restart. Recovery requires hand-editing JSON.
*(codex BLOCKER, reliability HIGH, correctness HIGH)*

**B6. Bash `write_instance_yaml` destroys all non-`database:` keys** — `scripts/ops/agnes-state-applier.sh:99-115`
Heredoc rewrites instance.yaml from scratch with only `database: {backend, url}`. Every other top-level key the operator set (logging, auth providers, feature flags, server-config overlays) is silently nuked on every successful migration. Contrast `src/db_state_machine.py:108-120` which does yaml.safe_load → merge → safe_dump correctly. Bash path is divergent.
*(codex BLOCKER, reliability HIGH, data-integrity LOW)*

**B7. Cloud URL can alias the sidecar → sidecar migrates onto itself → later applier tick stops its own DB** — `app/api/db_state.py:120,139`, `scripts/ops/agnes-state-applier.sh:134`
Validation is "non-empty" + raw-string inequality only. `postgresql+psycopg://agnes:pw@postgres/agnes` bypasses equality with `...@postgres:5432/agnes`. Migrates sidecar onto itself, marks backend `cloud`, then a later cloud-only applier tick stops `agnes-postgres-1`. **Operator destroys own database with one POST.**
*(codex BLOCKER)*

**B8. Two pending jobs queueable for the same transition** — `app/api/db_state.py:45,199`, `src/db_state_machine.py:59`
API flock released after pending-file write. `_current_job_id()` ignores pending. State machine allows `side_car_in_progress → side_car`. Second admin POSTs while first is queued → second job's `source_backend` recorded as `side_car_in_progress`, which the migrator CLI does not accept. Job orphaned in `pending` forever.
*(codex BLOCKER, reliability MED, correctness MED)*

---

### HIGH

**H1. `GET /api/admin/db/job/{id}` returns raw `target_url`/`source_url` with embedded credentials** — `app/api/db_state.py:205-211`
`/state` redacts via `_redact_url`; `/job/{id}` doesn't. Cloud SQL password recoverable by anyone with admin PAT (or who briefly exfiltrates one). Job files persist after migration.
*(security HIGH, data-integrity HIGH)*

**H2. Job JSON + `instance.yaml` overlay written world-readable (`0644`)** — `app/api/db_state.py:186,197`, `src/db_state_machine.py:125-127`, `scripts/db_state_migrator.py:54-58`, `scripts/ops/agnes-state-applier.sh:104-114`
Default umask in container = `0644`. `_LOCK_PATH` correctly uses `0o600` — writers should match. Any unprivileged process in the container reads cloud PG password.
*(security HIGH)*

**H3. No scheme validation on operator-supplied `cloud_url`** — `app/api/db_state.py:120-128`
Accepts `sqlite:///etc/passwd`, `mysql://...`, `file:///...`, or any host:port. Migrator container `ATTACH`es arbitrary endpoints — network pivot primitive. PR body advertises "source-url validation"; only check is `source != target`.
*(security HIGH)*

**H4. `copy_pg_to_pg` loads entire tables into RAM, single transaction** — `scripts/db_state_migrator.py:261-294`
`.all()` materializes everything; one `target.begin()` block per table. DuckDB path batches 500. agnes-dev had 5145 rows total; production `audit_log` / `usage_events` (millions of rows) will OOM the migrator container or blow the PG transaction.
*(correctness HIGH, data-integrity MED, reliability MED)*

**H5. No `timeout=` on migrator subprocesses (`docker run`, `alembic upgrade`, `pg_dump`)** — `scripts/ops/agnes-state-applier.sh:185-198`, `scripts/db_state_migrator.py:142-155,418-424`
Hung target PG, network partition mid-copy, hung pg_dump → migrator pinned forever. Combined with the 30s flock at line 32, every subsequent timer tick exits silently. No monitoring signal, no auto-cancel, no operator alert beyond "job stuck in running".
*(reliability HIGH)*

**H6. `run_all` continues past per-table failures → orphan rows with no PG-level FK to verify integrity** — `scripts/migrate_duckdb_to_pg/__init__.py:196-201`
On `users` exception, the loop `continue`s and still copies `personal_access_tokens` / `user_group_members`. These tables have **no PG-level FK to `users`** (`migrations/versions/0005_config_and_pat.py:77-95` — none). Orphan rows inserted; row-count verifier passes because both sides match.
*(data-integrity HIGH)*

**H7. `audit_log.params` / `params_before` historical rows copied verbatim — no migration-time PII scrub** — `scripts/migrate_duckdb_to_pg/tasks.py:42-65`
`_sanitize_for_audit` is at-write-time only (added later than some historical rows). Pre-sanitization audit rows containing passwords/tokens/PII faithfully replicated to PG store. No pre-copy scrub pass.
*(data-integrity HIGH)*

**H8. Stale pending jobs executed without age or source-state revalidation** — `app/api/db_state.py:173`, `scripts/ops/agnes-state-applier.sh:65`
Pending files have no `created_at`, no expiry, no compare-and-swap against current backend. Operator masks timer, queues migration, manually fixes state, unmasks days later → applier blindly runs whatever pending JSON remains.
*(codex HIGH)*

**H9. v9 JSONB string-CAST fix only applied on PG→PG; DuckDB→PG path still crashes on bare-string `params`** — `scripts/db_state_migrator.py:205-218` (PG→PG only) vs `scripts/migrate_duckdb_to_pg/tasks.py:201-206` (`json.dumps` only for `dict|list`)
Any bare string slipped into `audit_log.params` by a non-`AuditRepository.log` writer hits `CAST(:col AS JSONB)` and crashes mid-migration on the duckdb→side_car path. Symmetrize.
*(data-integrity HIGH)*

**H10. v9 JSONB CAST fix shipped with 0 regression tests** — commit `df1a587f`: 44 LOC migrator, 0 test LOC
PR body advertises "JSONB CAST" as new regression test; `grep -in 'jsonb\|json_dumps' tests/db_pg/` → empty. Future revert silently passes. PR body itself flags this class: *"All nine blockers found by live agnes-dev verification, none caught by the green PG/DuckDB suites at PR time."*
*(testing HIGH)*

**H11. New bidirectional state-machine transitions have no E2E coverage** — `ff83685d` added `duckdb→cloud` (direct skip) and `cloud→side_car` (DR rollback)
Only legacy `duckdb→side_car` covered. Both new arms ship green based on unit-level dispatch + matrix tests only. A regression in `source_backend` dispatch in `main()` would slip through.
*(testing HIGH)*

**H12. PG→PG verification only checks row counts → stale non-PK data slips through** — `scripts/db_state_migrator.py:336`, `scripts/migrate_duckdb_to_pg/tasks.py:344`
DuckDB validation hashes PKs only; result not promoted to migration failure. Preseed Cloud SQL with matching user IDs but changed RBAC/auth fields → migration "succeeds" with stale state.
*(codex HIGH)*

---

### MEDIUM (selected)

- **No `--yes` confirmation on `agnes admin db migrate <target>`** (`cli/commands/db.py:59-98`) — destructive cutover with no operator confirmation gate. *(security)*
- **`agnes-state-applier.service` runs as root with `-v /data:/data`** (`scripts/ops/agnes-state-applier.service:1-8`) — daemon's input surface (pending job JSON) is root-equivalent. Drop to dedicated user in `docker` group. *(security)*
- **`POSTGRES_PASSWORD` env-missing → falls back to literal `"agnes"`** (`app/api/db_state.py:125-126`) — silently ships known-weak credential. Convert to explicit 500 with actionable error. *(security)*
- **`_substitute_default` fabricates `datetime.now()` for NOT NULL timestamp columns missing values** (`scripts/migrate_duckdb_to_pg/tasks.py:152-173`) — rewrites historical `audit_log.timestamp`, `users.created_at` etc. to the migration moment. **Audit trail integrity broken.** *(data-integrity)*
- **`backup_duckdb` runs AFTER copy, not before** (`scripts/db_state_migrator.py:533-537`) — if verify passes but flip crashes, operator has neither backup nor flipped state. No explicit pre-copy backup. *(data-integrity)*
- **Pending-job glob filesystem-ordered, not arrival-ordered** (`scripts/ops/agnes-state-applier.sh:67-74`) — `for f in *.json` picks UUID-alphabetical; on concurrent pending jobs the wrong one runs. *(data-integrity, correctness)*
- **`copy_duckdb_to_pg` silently drops per-table errors from summary** (`scripts/db_state_migrator.py:178-184`) — `run_all` records `{table, error}` but downstream filter `if 'error' not in r` discards them; verifier may match `0=0`; migration ships success. *(correctness)*
- **Per-table `data_copy` progress never updates** — `JobWriter.update_table_progress()` exists, never called. progress_pct freezes at 40% for the longest step. UI cannot distinguish hung vs slow. *(correctness)*
- **Failed `cloud→side_car` leaves orphan postgres container** — applier writes `write_instance_yaml SOURCE_BACKEND` (cloud) but leaves flag file at `side-car-enabled`; next tick keeps `agnes-postgres-1` running. *(correctness)*
- **No `connect_timeout` / `statement_timeout` on migrator PG engines** (`scripts/db_state_migrator.py:171-178, 254-258, 297-298, 332-333`) — unreachable target hangs migrator. *(reliability)*
- **Applier timer = sole forward-progress driver, no API health probe** — masked/disabled timer → pending jobs sit forever. Add heartbeat + `applier_last_tick_age_s` to `/state`. *(reliability)*
- **`pg_isready` 60s wait loop swallows failure** (`scripts/ops/agnes-state-applier.sh:127-131`) — break never tracked; loop exits → applier launches migrator anyway → less actionable error. *(reliability)*
- **Backup failure during `side_car→cloud` swallowed as warning** (`scripts/db_state_migrator.py:506-513`) — UI shows success on completion; operator discovers missing recovery point only when needed. *(reliability)*
- **`RESTART_RC=$?` branch is dead code under `set -e`** (`scripts/ops/agnes-state-applier.sh:200,232-238`) — failed `dc up` exits script before warning logger fires. *(reliability)*
- **PG→PG copy test uses same-URL, `users(id,email,name)` only** — no array, no JSONB. Cannot catch v9 JSONB bug as written. *(testing)*
- **`_substitute_default` tests only `CURRENT_TIMESTAMP` branch** — `NOW()` and `CURRENT_DATE` paths uncovered. *(testing)*

---

### LOW

- `duckdb→cloud` direct path has no DuckDB snapshot before flip
- `verify_row_counts` opens DuckDB writable → creates WAL alongside
- Manual migrator invocation broken after API has wiped URL
- Applier python3 heredoc parses job JSON 4 separate times (partial-write window widens)
- `set -e` + python3 heredoc → applier abort with lock auto-released, no structured rollback
- Same-URL DR check is naive string-equal (subsumed by B7; other dialect/path bypasses still possible)
- `resource_grants.resource_id` has no FK (parity with DuckDB, but PG migration was the chance to add)
- `test_post_migrate_does_not_spawn_subprocess` only checks 202; vacuous as behavior guard
- Applier shell test asserts exact `docker run` line ordering — brittle under refactor

---

### Testing gaps

PR body's own admission: *"All nine blockers found by live agnes-dev verification, none caught by the green PG/DuckDB suites at PR time"* is the right diagnosis. Structural gaps surface again here:

- **No test for JSONB string-coercion** (v9 fix). Seed a row with bare-string `audit_log.params`; assert round-trip on both DuckDB→PG and PG→PG paths.
- **No E2E for `duckdb→cloud`** direct migration.
- **No E2E for `cloud→side_car`** DR rollback.
- **No `cancel-during-data-copy` test** asserting in-flight migrator is signaled and target left clean.
- **No hung-migrator timeout test** — applier should bound `docker run` runtime.
- **No host-reboot-mid-migration test** — `status="running"` recovery is silently absent.
- **No `instance.yaml` preserves-other-keys test** for the bash `write_instance_yaml` path (B6).
- **No concurrent `POST /migrate` test** asserting only one job ever queues (B8).
- **No URL-redaction test** for `GET /api/admin/db/job/{id}` (H1).
- **No file-mode test** asserting `0600` on job JSON / overlay (H2).
- **No malformed-`cloud_url` test** (`sqlite://`, `file://`, `http://`) (H3).

---

### Residual / strategic risks

- The whole `/api/admin/db/*` surface relies on admin-group authorization as the only gate. Any future RBAC bug → B7, H1, H2, H3 become catastrophic.
- `instance.yaml` is the single source of truth for the runtime DB URL — host-root or anyone in the `agnes` VM group can swap backend without going through the state machine.
- v5 "host-orchestrated" architecture correctly solved the DuckDB lock-release symptom (Python pinning FDs in uvicorn workers — root cause well understood in the script docstring). New risk shifted to operator blindness: no API-side visibility into applier liveness.
- Cooperative cancellation requires both a cancel sentinel the migrator checks at step boundaries AND a docker-kill path on `POST /cancel`. Neither shipped.
- The author's PR body claims 9 blockers caught by live verification — this review surfaces another 8 BLOCKER-class and 12 HIGH-class. The live-verification corpus (5145 rows, single operator, happy path) does not exercise concurrency, OOM, OOB-URLs, malicious admin, host reboot, mid-migration cancel, or production-scale audit_log volume.

---

Reviewers used: agnes-architecture (out-of-scope; PR doesn't touch orchestrator/connectors), agnes-rbac (clean), agnes-rules (clean), CE-correctness, CE-security, CE-data-integrity, CE-reliability, CE-testing, Codex adversarial. CE-data-migrations stalled at 600s no-progress watchdog (coverage overlapped with data-integrity + correctness).

cvrysanek:
### Cross-check against a parallel PG cutover branch

I had been carrying my own PG cutover branch in parallel (`vr/postgres-local-snapshot`, snapshot of the line of work that went into the original #388). Most of that work either overlaps with this PR or is superseded by the better architecture here. But four specific items in my branch caught real bugs that did not get re-discovered on the way to this PR, and I want to flag them so they don't slip through.

All four fact-checked against `git show origin/zs/db-state-machine:<path>` — paths and line numbers are from this PR's HEAD.

---

**1. BLOCKER — `scripts/migrate_duckdb_to_pg/__main__.py` exits `0` on per-table migration failure**

`run_all` (`scripts/migrate_duckdb_to_pg/__init__.py`) handles per-task exceptions by appending `{"table": ..., "error": ...}` to the reports — no `checksum_match` key is set. The CLI's exit logic is:

```python
return 0 if all(r.get("checksum_match", True) for r in reports) else 1
```

`r.get("checksum_match", True)` on the error report returns `True` (default), so the script **exits 0 despite a per-table exception**. `scripts/ops/agnes-state-applier.sh:185-198` reads `MIG_RC=0` → branch into success path → flip backend → app comes up with partially-populated PG and no signal to the operator that anything failed.

Combined with the row-count verifier behaviour I flagged separately (catches `ProgrammingError` and substitutes `tgt_count=0`), this means a partial seed where DuckDB has rows but the target table is missing or the copy crashed mid-table is *structurally indistinguishable* from a clean run on the cutover path.

Fix: tighten the exit predicate to `0 if all('error' not in r and r.get('checksum_match', True) for r in reports) else 1`.

---

**2. HIGH — `_filtered_columns` silently drops DuckDB columns missing from PG schema, no data-loss probe**

`scripts/migrate_duckdb_to_pg/__init__.py` (Zdenek's version, refactored as `GenericCopyTask`) intersects DuckDB columns with PG columns and INSERTs the subset. Columns DuckDB has but PG doesn't are dropped silently (logged at most as a warning).

This bit me specifically on `table_registry.bq_fqn` and the v55+ documentation fields: prod DuckDB had data in those columns, the PG alembic chain was a version behind, the migrator logged `column dropped` and moved on, and `/catalog/p/<slug>` came up empty in the new instance.

Fix from my branch: before dropping, probe the column with `SELECT COUNT(*) FROM source WHERE "<col>" IS NOT NULL`. If `> 0`, **raise** — the operator MUST land the alembic migration before the snapshot proceeds. If `= 0` (column empty everywhere), `log.warning` and continue. Silent data loss is the alternative, and the migrator runs unattended via the applier daemon — nobody scrolls the log.

---

**3. MEDIUM — `tests/test_api_design_rules.py` only does the static OpenAPI declaration check**

`test_protected_endpoints_declare_auth_errors(spec)` walks the OpenAPI schema and asserts every non-PUBLIC `/api/*` route declares `401` + `403`. This check passes for free as long as `_add_auth_error_responses` injects the response schemas — regardless of whether the route's `Depends(require_admin)` / `get_current_user` chain actually fires.

A future endpoint that declares the response codes but forgets the auth dep (the most common shape of a real auth-bypass regression) slips through this test.

Add a runtime probe: spin up a `TestClient`, fill path params from a small dict of safe placeholders, hit every non-PUBLIC `/api/*` route without `Authorization`, assert the response is **not** 2xx. I have a working version (`test_protected_endpoints_actually_enforce_auth`) that catches the exact class of bug; happy to PR it as a follow-up.

---

**4. MEDIUM — `tests/db_pg/conftest.py` has no module-scoped alembic fixture**

Every PG test re-runs the full alembic chain via the `state_backend` / `pg_engine` fixture (drop schema → `command.upgrade(cfg, "head")`). Heavy modules pay this cost N times per file. Adding a `pg_engine_migrated_module` fixture (module-scoped, runs alembic once per module) + a `pg_truncate_all(engine)` helper that does a per-test `TRUNCATE ... RESTART IDENTITY CASCADE` on every base table except `alembic_version` cuts ~50% off the PG suite runtime.

In my branch this took the full suite from 6:51 → 4:24. Trade-off is that schema drift between migrations and the in-memory model is no longer caught test-by-test in modules that switch to the module-scoped fixture — but the `test_alembic_roundtrip.py` drift-detector runs at fresh-module scope and catches that class of bug.

---

**Auxiliary scripts (optional)**

Not strictly gaps — these are operator workflows my branch carried that don't have a counterpart here. Filing in case they're useful follow-ups:

- `scripts/sync_from_prod.sh` — repeatable prod→local seed. Uses a strict `KEY=VALUE` parser for `.env` (no `source`, so values with `#`/`&`/`$()`/backticks don't truncate or execute) and gates the destructive prod-data-pull behind `--yes`.
- `scripts/dev_pg_migrate_volume.sh` — moves a developer's anonymous `agnes-postgres-1` volume to a named volume via `pg_dump` → rename old → up new → restore → verify → drop old, with rollback on failure. Catches the anonymous-volume gotcha when devs follow the older docker-compose docs.

---

The intent of these notes is "merge faster with the bugs the live agnes-dev verification did not hit", not "block on parallel-branch parity". The big architecture wins (host-orchestrated applier, bidirectional state machine, two-stage cutover) all belong to this PR, not mine.

cvrysanek:
## Round-2 review of PR #455 — after 50-commit fix wave

Ran 3 targeted re-reviewers (CE-correctness, CE-security, Codex adversarial) against the delta `5fa6d0c4..origin/zs/db-state-machine`. All prior BLOCKER + HIGH findings were systematically addressed with dedicated commits — good response. But the round of fixes introduced new bugs of its own (5 BLOCKER, 6 HIGH, several MED/LOW), and several fix patches are partial.

**Verdict:** closer to merge but still has new BLOCKERs. Most are localised to the recently-added multi-destination + non-root applier + same-DB-alias surfaces.

---

### Round-1 findings — verification

**Every B1-B8 BLOCKER and every "gap" item PASSES.** Verified by reading the fix commits + new code:

| Finding | Fix commit | Status |
| --- | --- | --- |
| B1 cancel revert uses source_backend | `b1f0d73e` | PASS |
| B2 cooperative cancellation sentinel | `e0f2d4dd` | PASS (see H1-NEW below for residual race) |
| B3 verify_row_counts raises on missing target | `bb01de2d` + `7a7a5ded` (content-hash drift) | PASS |
| B4 write_backend_state Ellipsis sentinel preserves URL | `0cd65572` | PASS (but see C2-NEW below for rollback-path regression) |
| B5 heartbeat + stuck-running recovery | `598c0da5` | PASS (see H5-NEW below for state cleanup gap) |
| B6 write_instance_yaml preserves keys | `32a936a9` | PASS (see H4-NEW below for PyYAML dependency) |
| B7 _urls_alias normalises host/port/db | `a382d2f7` | PASS (see B7-NEW below for alias bypass) |
| B8 _current_job_id surfaces pending | `da716ca0` | PASS (see B2-NEW below for residual race) |
| Exit code halts on per-table failure | `bb01de2d` | PASS |
| `_filtered_columns` data-loss probe | `0fab56e1` | PASS |
| Runtime auth probe test | landed in `tests/test_api_design_rules.py:340` | PASS |
| Module-scoped alembic fixture | `23661e7a` | PASS |
| H1 redact url in GET /job | `5e8e2cc2` | PASS (see H1-NEW below for `error.message` leak) |
| H2 0600 mode on overlay + job files | `b14a390a` | PARTIAL — see below |
| H3 cloud_url scheme + host + db validation | `e9a47722` | PASS (see SSRF below) |
| `--yes` on `agnes admin db migrate` | `0c57e8ed` | PARTIAL — see below |
| Non-root applier | `34e20919` | PASS (see B8-NEW for permission gaps) |
| POSTGRES_PASSWORD missing → 500 | `e5ab0cca` | PASS (but compose still defaults to literal `agnes` — see LOW below) |
| H6 audit-log PII scrub | `e4b66441` | PASS (see C4 below for over-redaction LOW) |
| H7 verification content-hash | `7a7a5ded` | PASS |
| H8 pending-job age expiry | `def04628` | PASS (see B2-NEW for race-after-expiry) |
| H11 bidirectional E2E tests | `e30f4bca` | PASS |

---

### NEW BLOCKERs introduced by the fix wave

**B1-NEW. Two concurrent `POST /migrate` calls both pass validation before either takes the flock; second request never re-reads state after lock acquisition** — `app/api/db_state.py:206,258,270`
The B8 fix surfaces pending jobs in `_current_job_id`, but the validation order is still `validate_transition → flock → write`. Two admins racing through validation before either reaches the lock both pass, and the second write overwrites the first flag file. Codex repro: "have two admins POST different targets concurrently, with request B delayed between validation and lock acquisition." Move validate inside the flock.

**B2-NEW. `_urls_alias` same-DB detection only compares lowercased `(host, port, db)` strings — `postgres` (compose service name) vs sidecar container IP or CNAME bypasses the guard** — `app/api/db_state.py:47,250`, `scripts/ops/agnes-state-applier.sh:255`
The B7 fix added `_normalize_pg_url` but it doesn't resolve hostnames. From inside the migrator container, `postgresql://...@postgres:5432/agnes` vs `postgresql://...@172.18.0.2:5432/agnes` (same DB, different host string) pass the inequality check. side_car → cloud "migration" copies the DB to itself, marks cloud success, next cloud-only applier tick stops `agnes-postgres-1`. Same B7-class outage, different bypass.

**B3-NEW. Non-root applier cannot read `/opt/agnes/.env`** — `infra/modules/customer-instance/startup-script.sh.tpl:262,283`, `scripts/ops/agnes-state-applier.service:9`, `scripts/ops/agnes-state-applier.sh:55`
Terraform-provisioned VMs write `.env` as root with mode `0600`. The new systemd unit runs as `agnes-applier` and sources the file directly. **Every newly-provisioned VM** has a broken applier — symptom: timer fires, applier exits silently. Either ACL the file to the new user or have provisioning create the user before writing `.env`.

**B4-NEW. `side-car-enabled` lifecycle does `chown 70:70 /data/postgres` — fails as `agnes-applier` under `set -e`** — `scripts/ops/agnes-state-applier.sh:228-229`
Even after fixing B3-NEW, the chown call requires root. Aborts the applier mid-bootstrap. Either grant the user CAP_CHOWN or change ownership at provision time.

**B5-NEW. Alembic 0013 creates the per-type CHECK constraint BEFORE backfilling the new per-type FK columns** — `migrations/versions/0013_resource_grants_per_type_fk.py:124,176`
Any existing `resource_grants(resource_type='table', resource_id='foo')` row violates the constraint while all new typed-FK columns are still NULL. Alembic upgrade aborts on any prod instance that has grants. Standard pattern: ADD CHECK ... NOT VALID + backfill + VALIDATE CONSTRAINT, or split into two migrations.

---

### NEW HIGH

**H1-NEW. Cancel-during-verify race** — `scripts/db_state_migrator.py:1061,1077`, `app/api/db_state.py:368,388`
B2's sentinel cancellation polls at step boundaries. Cancel arriving between the migrator's last cancel check and `flip_backend` is accepted by the API (writes `cancelled` + reverts to source), while the migrator proceeds to flip and mark success. End state: instance.yaml says SOURCE but data is on TARGET. Solve by re-checking the sentinel inside the same transaction as `flip_backend`, or by acquiring an exclusive lock on the job during cancel.

**H2-NEW. Applier python-heredoc rewrites of job JSON drop the 0600 mode** — `scripts/ops/agnes-state-applier.sh:131,158`
The API endpoint and migrator's `JobWriter` both `os.chmod(0o600)` on write. The applier's two inline python heredocs (H8 expiry + `update_job`) do `os.replace(tmp, p)` with no follow-up chmod. The tmp file was created with the process umask → 0644. After the applier touches any job file, the embedded `target_url` (incl. plaintext password) becomes world-readable. Add `os.chmod(p, 0o600)` after each `os.replace`.

**H3-NEW. Alembic timeout RuntimeError leaks `target_url` (incl. password) into `job.error.message`** — `scripts/db_state_migrator.py:243`
H5 fix raises `RuntimeError(f"... (target={target_url!r}) ...")` on subprocess timeout. The migrator's outer handler captures it into `error.message`. `GET /api/admin/db/job/{id}` only redacts top-level `target_url`/`source_url`, NOT nested `error.message`. Password leaks via HTTP response → browser history → UI screenshots. Apply `_redact_url` to URLs inside error strings, or recursively redact `error` payload.

**H4-NEW. `write_instance_yaml` depends on host-level PyYAML, but provisioning only installs Docker Compose** — `scripts/ops/agnes-state-applier.sh:173-174`, `infra/modules/customer-instance/startup-script.sh.tpl:29`
The B6 fix replaces the bash heredoc with `python3 -c 'import yaml; ...'`. PyYAML is not in the provisioning bootstrap. On any host where `/usr/bin/python3 -c 'import yaml'` fails, the ERR trap fires AFTER a successful migrator run, marks the job failed, and skips the app restart. Apt-install `python3-yaml` in provisioning, or fall back to a pure-bash YAML writer for the database-only subset.

**H5-NEW. Stuck-running recovery marks failed but never restores `instance.yaml` from `*_in_progress`** — `scripts/ops/agnes-state-applier.sh:217-219`, `app/api/db_state.py:294`, `scripts/db_state_migrator.py:1168`
B5 fix detects stale `.alive` past 120s and marks the job failed. But it does not write `database.backend = source_backend` back into instance.yaml. The next migration retry reads `*_in_progress` as the current backend → the migrator CLI receives `source_backend='side_car_in_progress'` → rejects. State machine is wedged until an operator manually edits instance.yaml. Recovery should call `write_backend_state(source_backend, url=source_url)` symmetrically with the cancel path.

**H6-NEW. DuckDB→PG JSONB handling is still hardcoded — `_build_insert` only casts columns in `_JSON_COLUMNS`, but PG models declare additional JSONB columns** — `scripts/migrate_duckdb_to_pg/tasks.py:42,119`, `src/models/data_packages.py:55`
Specifically `data_packages.tags` is JSONB in the PG model but NOT in `_JSON_COLUMNS`. DuckDB→PG migration on a real instance with `data_packages.tags='["finance"]'` crashes at insert. The v9 JSONB fix from round 1 was supposed to symmetrize this; round 2 only added the PG→PG side and a fixed list of duckdb→pg columns. Derive `_JSON_COLUMNS` dynamically from `Base.metadata` instead of hardcoding.

**H7-NEW. `start_migration` crashes / mis-routes when target='duckdb' or 'duckdb_quack'** — `app/api/db_state.py:186-275`
Commit `965e7870` widened `_ALLOWED_TRANSITIONS` to allow side_car/cloud → duckdb*. But the migrate endpoint only branches on `'side_car'` vs `'cloud'`: `target_url=payload.cloud_url` for anything else → `None`. The in_progress write sets `CLOUD_IN_PROGRESS` for any non-side_car target → a duckdb target writes `CLOUD_IN_PROGRESS` into instance.yaml. The migrator main() has no duckdb dispatch branch → `BackendNotYetSupportedError` raised → uncaught 500. Either reject `duckdb*` at the endpoint until the migrator supports it, or wire the missing branches end-to-end.

**H8-NEW. ERR-trap rollback in applier wipes cloud URL from instance.yaml** — `scripts/ops/agnes-state-applier.sh:266-296` + `32a936a9`
`__rollback` calls `write_instance_yaml "$SOURCE_BACKEND"` with no second arg. The python helper interprets the missing url as "pop it". When a cloud→side_car migration fails mid-flight, instance.yaml goes back to `backend=cloud` but with **no url**. App boot then crashes with "Postgres URL unset" — re-introducing the B4-class outage on the rollback path. Pass `SOURCE_URL` as the second arg.

---

### MEDIUM

- **MED-1. `--yes` gate bypassed by `--json`** — `cli/commands/db.py:99`: `needs_confirm = not yes and not as_json`. `agnes admin db migrate cloud --cloud-url ... --json` (from CI/cron) proceeds destructively with no confirm. Either drop the `and not as_json` clause, or require `--yes` whenever `--json` is set.
- **MED-2. `_validate_cloud_url` accepts loopback / GCE metadata / RFC1918** — `app/api/db_state.py:63-83`. Admin sets `cloud_url=postgresql://attacker:pwd@169.254.169.254:5432/db` or `10.0.0.5:5432`; alembic upgrade head opens TCP, returns a server-fingerprint error, SSRF/port-probe primitive. Reject reserved address ranges unless explicit opt-in.
- **MED-3. `_redact_url` misses query-string passwords** — `app/api/db_state.py:87-91`. Regex `(://[^:]+:)[^@]+(@)` doesn't cover `postgresql://user@host/db?password=secret&sslmode=require`. Use `sqlalchemy.engine.make_url(...).render_as_string(hide_password=True)`.
- **MED-4. cancel_job revert for DUCKDB-source migrations leaves stale postgres url in overlay** — `app/api/db_state.py:389-390`. Reverting to `backend=duckdb` but preserving the migration-target postgres url makes the overlay self-inconsistent.

---

### LOW

- **LOW-1. PII scrub regex matches values not keys** — `scripts/db_state_migrator.py:scrub_audit_log_pii`. `pattern.search(str(params))` over the whole serialised JSON. An audit row whose value text contains `"please reset your password"` or HTTP path `/reset-password` triggers wholesale `{_redacted_at_migration:true}` replacement. Data-loss bias on the safe side, but legitimate operational history erased. Parse JSON and walk keys.
- **LOW-2. `docker-compose.postgres.yml` keeps `${POSTGRES_PASSWORD:-agnes}` literal fallback** — `docker-compose.postgres.yml:47,78,101,126,140`. The API guard refuses to start a migration without the env var, but `docker compose up` (the non-state-machine path) brings up Postgres with credentials `agnes/agnes`. Drop the `:-agnes` default; compose should fail fast.

---

### Testing gaps

- No test for concurrent `POST /migrate` race (B1-NEW).
- No test for same-DB detection bypass via IP vs hostname (B2-NEW).
- No test that newly-provisioned VM's applier can actually read `.env` (B3-NEW).
- No test that alembic 0013 upgrades cleanly on an instance with pre-existing typed grants (B5-NEW).
- No test that cancel-during-`verify` step leaves a consistent end state (H1-NEW).
- No test that applier's `update_job` rewrite preserves `0600` mode (H2-NEW).
- No test that timeout/error messages don't echo plaintext URLs (H3-NEW).
- No test that `write_instance_yaml` fails gracefully when PyYAML is unavailable on the host (H4-NEW).
- No test that stuck-running recovery restores instance.yaml's `database.backend` field (H5-NEW).
- No test for DuckDB→PG migration of `data_packages.tags` (H6-NEW).
- No test for `POST /migrate` with `target='duckdb'` (H7-NEW).
- No test for the applier ERR-trap rollback path on a cloud-source migration (H8-NEW).

---

**The good news**: every original BLOCKER is genuinely fixed, the test coverage of the original regression classes is now real, and the architecture (heartbeat, sentinel cancel, non-root applier, Ellipsis-preserve URL, multi-destination matrix) is sound. **The not-so-good news**: the 50-commit fix wave was clearly delivered under review pressure, and the new BLOCKERs cluster around the surfaces the wave touched. Suggest one more pass focused on (a) **the applier's lifecycle as a non-root user** (B3/B4/H5), (b) **alembic 0013 backfill ordering** (B5), and (c) **race windows around the cancel-flip and validate-flock seams** (B1/H1). Round-3 should be smaller.

cvrysanek:
### Round-2 review — TL;DR

All original BLOCKER (B1–B8) + HIGH + gap items from the prior comments: **PASS**. Every fix commit verified, regression tests landed.

The 50-commit fix wave introduced new bugs clustered in 3 zones:

**5 NEW BLOCKERs**
1. Concurrent `POST /migrate` race — validation passes before flock acquisition; B8 fix narrowed but didn't close the window
2. Same-DB alias bypass via container IP vs hostname (`postgres` → `172.18.0.2`) — B7 fix doesn't resolve DNS
3. Non-root applier can't read `/opt/agnes/.env` (mode `0600`, root-owned) — breaks every newly-provisioned VM
4. `chown 70:70 /data/postgres` fails as `agnes-applier` (set -e aborts mid-bootstrap)
5. Alembic `0013` creates CHECK constraint **before** backfilling per-type FK columns — upgrade aborts on any prod instance with existing typed grants

**8 NEW HIGH**
- Cancel-during-`verify` race — cancel reverts state while migrator continues to `flip_backend` + mark success
- Applier python-heredoc rewrites drop the `0600` mode → creds in job JSON become world-readable
- Alembic timeout `RuntimeError` embeds `target_url` (incl. password) in `error.message` — `/job/{id}` doesn't redact nested fields
- `write_instance_yaml` calls `python3 -c 'import yaml'`; provisioning doesn't install PyYAML
- Stuck-running recovery marks job failed but never restores `database.backend` from `*_in_progress` — wedges the state machine
- DuckDB→PG JSONB cast list is hardcoded; misses `data_packages.tags` and any future JSONB column
- `start_migration` crashes on `target='duckdb'`/`'duckdb_quack'` — multi-destination matrix shipped without wiring the endpoint
- ERR-trap rollback (`__rollback`) calls `write_instance_yaml SOURCE_BACKEND` with no url → drops it → reintroduces the B4-class outage on the failure path

**4 MED + 2 LOW**
- `--yes` gate bypassed by `--json` flag (CI/cron path)
- `_validate_cloud_url` accepts loopback / GCE metadata `169.254.169.254` / RFC1918 — SSRF + port-probe primitive
- `_redact_url` misses query-string passwords (`?password=secret`)
- `cancel_job` revert for DUCKDB-source leaves stale postgres url in overlay
- PII scrub regex matches values not keys (over-redacts on `"please reset your password"`)
- `docker-compose.postgres.yml` keeps `${POSTGRES_PASSWORD:-agnes}` literal fallback

**Round-3 should be smaller** — issues cluster around (a) non-root applier lifecycle (B3/B4/H5), (b) alembic 0013 backfill ordering (B5), (c) race windows around the cancel-flip and validate-flock seams (B1/H1).

Full breakdown with file:line refs in the comment above.

ZdenekSrotyr:
**Round-2 checkpoint — 7 of 21 fixes shipped (Phases A + B).** Full suite GREEN (5865 passed / 36 skipped / 0 failed).

| # | Finding | Commit |
|---|---|---|
| MED-1 | `--json` no longer bypasses `--yes` confirm gate | `e6569c59` |
| MED-3 | `_redact_url` masks query-string + sslpassword | `a4a57a48` |
| LOW-1 | PII scrub walks JSON keys, word-boundary anchored | `ed580574` |
| LOW-2 | Compose drops `POSTGRES_PASSWORD:-agnes` default | `e1da2980` |
| MED-2 | `_validate_cloud_url` rejects reserved IP ranges (SSRF) | `46334442` |
| MED-4 | `cancel_job` revert to duckdb drops stale PG URL | `5001fab9` |
| H7-NEW | `POST /migrate target=duckdb*` returns 501 (was 500/mis-route) | `96b7a382` |

Each fix landed via test-first (RED → fix → GREEN) + spec-compliance + code-quality sub-agent reviews. Full plan + remaining 14 tasks (B1-NEW, B2-NEW, B3-NEW tightening, B4-NEW tightening, B5-NEW, H1-NEW, H2-NEW, H3-NEW, H4-NEW, H5-NEW, H6-NEW, H8-NEW, NEW-X-USERS-DUPLICATES, final self-review) tracked in `docs/superpowers/plans/2026-06-01-db-state-machine-round-2-fixes.md` (commit `da4c5875`).

ZdenekSrotyr:
**Round-2 checkpoint — 10 of 21 fixes shipped (Phases A + B + C).** Full suite GREEN (5873 passed / 36 skipped / 0 failed; +8 from Phase B baseline).

| # | Finding | Commit |
|---|---|---|
| MED-1 | `--json` no longer bypasses `--yes` confirm gate | `e6569c59` |
| MED-3 | `_redact_url` masks query-string passwords | `a4a57a48` |
| LOW-1 | Migrator PII scrub walks JSON keys, not raw values | `ed580574` |
| LOW-2 | Compose: removed `POSTGRES_PASSWORD` literal-agnes default | `e1da2980` |
| MED-2 | `_validate_cloud_url` rejects reserved IP ranges | `46334442` |
| MED-4 | Cancel revert to duckdb clears stale postgres url | `5001fab9` |
| H7-NEW | `start_migration` returns 501 on duckdb target | `96b7a382` |
| **H3-NEW** | **Migrator + API redact passwords inside `error.message`** | **`59933f10`** |
| **H6-NEW** | **`_JSON_COLUMNS` derived from `Base.metadata` (caught 7 drifted JSONB cols)** | **`bcaf40b2`** |
| **NEW-X** | **`ON CONFLICT DO NOTHING` covers every UNIQUE (live-discovered 2026-06-01)** | **`2169a106`** |

**Phase C wrap-up:**
- H3-NEW defence-in-depth: migrator masks at raise site (`make_url(...).render_as_string(hide_password=True)`) + API recursively scrubs URL-shaped substrings from the entire `error` payload via `_redact_urls_in_text` + `_redact_error_payload`.
- H6-NEW dynamic JSONB derivation surfaced **6 additional** previously-missing columns beyond the spec's `data_packages.tags`: `data_packages.when_to_use/when_not_to_use/example_questions`, `recipes.related_table_ids`, `table_registry.sample_questions/pairs_well_with`. The `_build_json_columns()` helper now walks `Base.metadata.sorted_tables` once at module load. Forward-compatibility test (`test_json_columns_covers_every_jsonb_in_metadata`) is a self-improving guard.
- NEW-X switched `ON CONFLICT (id) DO NOTHING` → bare `ON CONFLICT DO NOTHING` so the migrator stays idempotent on every UNIQUE constraint (PK, `users.email`, etc.). Live-discovered while running cycle 1 on a freshly-provisioned VM with 6 distinct-email source users.

**Next:** Phase D — applier hardening (H2-NEW heredoc 0600 mode, H4-NEW, H8-NEW). Continuing in the same session.

ZdenekSrotyr:
**Round-2 checkpoint — 13 of 21 fixes shipped (Phases A + B + C + D).** Full suite GREEN (5880 passed / 36 skipped / 0 failed; +15 from baseline before round-2).

| # | Finding | Commit |
|---|---|---|
| MED-1 | `--json` no longer bypasses `--yes` confirm gate | `e6569c59` |
| MED-3 | `_redact_url` masks query-string passwords | `a4a57a48` |
| LOW-1 | Migrator PII scrub walks JSON keys | `ed580574` |
| LOW-2 | Compose dropped `POSTGRES_PASSWORD` literal default | `e1da2980` |
| MED-2 | `_validate_cloud_url` rejects reserved IPs (SSRF/IMDS) | `46334442` |
| MED-4 | Cancel revert to duckdb clears stale postgres url | `5001fab9` |
| H7-NEW | `start_migration target=duckdb` returns 501 | `96b7a382` |
| H3-NEW | Migrator + API redact passwords inside `error.message` | `59933f10` |
| H6-NEW | `_JSON_COLUMNS` derived from `Base.metadata` (caught 7 drifted JSONB cols) | `bcaf40b2` |
| NEW-X | `ON CONFLICT DO NOTHING` covers every UNIQUE (live-discovered) | `2169a106` |
| **H2-NEW** | **Applier chmod 0600 after job JSON rewrite** | **`c5cd5b53`** |
| **H4-NEW** | **`write_instance_yaml` bash fallback + `python3-yaml` apt-install** | **`96d7dc2f`** |
| **H8-NEW** | **Applier `__rollback` + failed-migration branch preserve `SOURCE_URL`** | **`0c9b343c`** |

**Phase D wrap-up:**
- H2-NEW: both `os.replace` sites in the applier's job-rewrite heredocs now chmod 0600 (tmp file inherited 0644 umask → world-readable job JSON containing `target_url` with password). Static-text test scans every `os.replace` site and asserts a follow-up `os.chmod(..., 0o600)` within 5 lines.
- H4-NEW: `write_instance_yaml` now probes PyYAML (`python3 -c 'import yaml' 2>/dev/null`), falls back to a pure-bash writer that emits only the database section + WARN log (non-database top-level keys lost — survivable degradation). `startup-script.sh.tpl` apt-installs `python3-yaml` so the fallback is defensive-only.
- H8-NEW: extended mid-task to cover **two** identical-class sites (the `__rollback` ERR-trap AND the failed-migration `else` branch in the main loop). Both pass `${SOURCE_URL:-}` (safe under `set -euo pipefail`). Generalized test asserts every `write_instance_yaml "$SOURCE_BACKEND"` line contains `SOURCE_URL` — forward-compatible against future revert sites.

**Next:** Phase E — race windows & state restore (B1-NEW validate inside flock, B2-NEW `_urls_alias` DNS resolve, H1-NEW cancel-during-verify race, H5-NEW stuck-running recovery restores backend). Continuing in the same session.

ZdenekSrotyr:
Round-2 review fully addressed — see [CHANGELOG verification matrix](https://github.com/keboola/agnes-the-ai-analyst/blob/zs/db-state-machine/CHANGELOG.md#round-2-review-fixes--verification-matrix) and the PR description's "Round-2 fixes shipped" section. All 20 findings have a fix commit + regression test on this branch. Full suite GREEN: **5903 passed / 36 skipped / 0 failed** (+30 from pre-round-2 baseline). Ready for round-3.

cvrysanek:
## Round-3 review of PR #455 — after 37-commit response to round-2

Three reviewers (CE-correctness, CE-security, Codex adversarial) against `2fb9759a..origin/zs/db-state-machine` (37 new commits + release 0.57.0).

**Verdict:** every round-2 BLOCKER + HIGH has a dedicated fix commit, almost all PASS. But the adversarial pass finds **3 NEW BLOCKERs + 2 PARTIALs (one HIGH, one MED) + 1 NEW HIGH** clustering around the new hostname-resolution / bash-YAML-fallback / stuck-running-recovery surfaces.

---

### Round-2 findings — verification

**ALL PASS except 2 PARTIALs.** Verified by reading fix commits + new code:

| Finding | Fix | Status |
| --- | --- | --- |
| B1 concurrent POST race | `dd9516e7` validation inside flock | PASS |
| B2 same-DB alias bypass via IP | `9ea5d433` + `46334442` `_urls_alias` resolves | PARTIAL — see B1-NEW below |
| B3 applier can't read /opt/agnes/.env | `9f9db5ec` chown agnes-applier 0600/0640 | PASS |
| B4 chown /data/postgres fails | `551fca78` + `69c203ee` bootstrap + idempotent guard + `SupplementaryGroups=docker` | PASS |
| B5 alembic 0013 CHECK before backfill | `b2fb61e1` backfill first | PASS |
| H1 cancel-during-verify race | `d066ffdf` `_check_cancel_before_flip` | **PARTIAL** — see H1-PARTIAL below |
| H2 applier loses 0600 on rewrites | `c5cd5b53` chmod 0600 after os.replace at both sites | PASS |
| H3 alembic timeout URL leak | `59933f10` `render_as_string(hide_password=True)` + recursive `_redact_error_payload` | PASS |
| H4 PyYAML not provisioned | `96d7dc2f` python3 probe + bash fallback | PASS — but see B2-NEW below for fallback YAML escaping |
| H5 stuck-running recovery | `2bf20321` recovery resets instance.yaml | PASS — but see B3-NEW below for app-restart gap |
| H6 hardcoded `_JSON_COLUMNS` | `bcaf40b2` derive from `Base.metadata` | PASS |
| H7 start_migration crashes on duckdb target | `96b7a382` 501 `BackendNotYetSupportedError` | PASS |
| H8 ERR-trap rollback wipes cloud URL | `0c9b343c` passes SOURCE_URL | PASS — but see H1-NEW below for empty-URL edge |
| MED-1 `--json` bypasses `--yes` | `e6569c59` drop `as_json` clause + non-TTY refuse | PASS |
| MED-2 SSRF loopback/RFC1918/metadata | `46334442` reserved-range ladder | **PARTIAL** — see MED-1-PARTIAL below |
| MED-3 query-string passwords | `a4a57a48` `render_as_string` + query-string regex | PASS |
| MED-4 cancel revert stale url | `5001fab9` explicit None for duckdb, Ellipsis otherwise | PASS |
| LOW-1 PII scrub matches values | `ed580574` walks JSON keys recursively | PASS |
| LOW-2 compose POSTGRES_PASSWORD default | `e1da2980` `:-agnes` removed | PASS |

---

### NEW BLOCKERs (from codex adversarial round)

**B1-NEW. DNS-alias check is time-of-check, not time-of-use** — `app/api/db_state.py:93`
B2 fix added `_resolve_host` + IP-set overlap. But the resolution happens at `_urls_alias` validation time; the hostname is stored raw in the job at `app/api/db_state.py:521` and the applier passes it unresolved at `scripts/ops/agnes-state-applier.sh:446`. Attack: POST `side_car → cloud` with `cloud_url=postgresql+psycopg://agnes:pw@rebind.example/agnes` where `rebind.example` resolves to a public IP at validation time, then to the side-car IP when the migrator connects (DNS rebinding, or a manual hosts-file flip on the host). Self-migration passes validation; cloud-only applier tick later stops the only live Postgres. Pin the resolved IP at validation time and pass it (not the hostname) to psycopg, or re-validate immediately before connect.

**B2-NEW. Bash YAML fallback emits malformed `instance.yaml` for URLs containing YAML-special chars** — `scripts/ops/agnes-state-applier.sh:210`
H4 fallback was meant for hosts without PyYAML. SQLAlchemy validates a `cloud_url` like `postgresql+psycopg://u:p@cloud.example/agnes?application_name=a:b` fine. The bash fallback emits the URL as an **unquoted** YAML scalar at lines 210-215. On parse, `read_backend_state` at `src/db_state_machine.py:174` treats the parse error as "no backend set → default to DuckDB". End state: migration succeeds, instance.yaml looks intact, restarted app silently routes back to DuckDB while the data lives on Postgres. **Triggerable today** because nothing rejects YAML-special chars in `cloud_url`. Fix: shell-escape the URL into a YAML double-quoted scalar (`"…"` with `"` and `\` escaped), or refuse to emit when PyYAML missing.

**B3-NEW. Stuck-running recovery marks job failed + resets state — but never restarts the stopped app/scheduler** — `scripts/ops/agnes-state-applier.sh:249`
H5 fix added recovery: applier picks pending job, stops `agnes-app-1` + `agnes-scheduler-1` at line 413, gets SIGKILL/OOMed mid-migration. Next timer tick: the stale-heartbeat detection at line 249 marks the job failed and restores `database.backend` via `write_instance_yaml`. **Then exits at lines 317-319 because no pending job remains.** Nobody ever runs `dc up -d app scheduler`. Service stays DOWN indefinitely. Recovery path needs a "post-recovery: restart app + scheduler" step before the no-pending-job exit.

---

### NEW HIGH

**H1-NEW. PG-source rollback cannot recover when `SOURCE_URL` was never set** — `app/api/db_state.py:476`, `scripts/ops/agnes-state-applier.sh:479`
H8 fix made `__rollback` pass `SOURCE_URL`. Edge case: operator starts with `database.backend: cloud` but `database.url` unset (e.g., migrated via `agnes admin db migrate cloud` from DuckDB, then admin edits instance.yaml or `database.url` got cleared by an earlier failure). POSTing target `side_car` queues a PG-source job with `source_url=None`. Migrator fails validation (no `--source-url`). Failed branch writes `write_instance_yaml "$SOURCE_BACKEND" "${SOURCE_URL:-}"` — empty url. Rollback preserves `backend=cloud` with no url. Manual YAML repair required to recover. Either refuse to queue when `source_url` would be None, or compute it from a configured fallback before the migrator runs.

---

### NEW PARTIALs

**H1-PARTIAL. Cancel ↔ flip race narrowed but split-brain still reachable** — `app/api/db_state.py:578-650`, `scripts/db_state_migrator.py:194-238`
`d066ffdf` added `_check_cancel_before_flip` immediately before `write_backend_state(target_state)`. Sequence still reachable:

1. T0: cancel reads `current_step='verify'` → passes point-of-no-return check
2. T1: migrator runs `writer.update_step('flip_backend', 95)`
3. T2: migrator runs `_check_cancel_before_flip` → sentinel not yet written → returns OK
4. T3: migrator runs `write_backend_state(target_state)` flipping instance.yaml to TARGET
5. T4: cancel handler calls `sentinel.touch()` (too late)
6. T5: cancel handler runs `write_backend_state(BackendState(source_backend))` overwriting back to SOURCE

**End: data on TARGET, instance.yaml on SOURCE — exactly the split-brain H1 was meant to prevent.** Microsecond window but reachable under fork-scheduling delay. Neither side holds `MigrationLock` during the flip. Fix: cancel handler must acquire the flock around its `write_backend_state` revert AND the migrator must hold the flock atomically across `_check_cancel_before_flip → write_backend_state`. Atomic check-and-flip.

**MED-1-PARTIAL. `_validate_cloud_url` rejects IP literals but skips DNS resolution for hostname inputs** — `app/api/db_state.py:163-164`
`46334442` rejects `127.0.0.1`, `169.254.169.254`, `10.x/172.16.x/192.168.x`, link-local, multicast, reserved. After `ipaddress.ip_address(host)` raises ValueError on a non-literal hostname, the function `return`s with no DNS check. So `postgresql+psycopg://u:p@metadata.google.internal/db`, or any attacker-controlled hostname resolving to a reserved range, bypasses. **Same TOCTOU primitive as B1-NEW** but at the validation layer instead of the migration layer — the SSRF guard is bypassed by any hostname input. Resolve via `socket.getaddrinfo` and run each returned IP through the same reserved-range ladder.

---

### Residual risks (callouts, not new findings)

- **`SupplementaryGroups=docker`** on the applier service unit ≈ root via docker socket. Documented design tradeoff; agnes-applier compromise = host-root compromise. Should be flagged in the deployment runbook.
- **Unconditional `chown 70:70 /data/postgres`** rewrites ownership of every file under the directory if the disk is reused across backends (e.g., on a Cloud SQL → side_car DR rollback). Acceptable for fresh-VM provisioning; risky if same disk is reused.
- **H4 bash YAML fallback drops all non-database top-level keys** when PyYAML missing — operator-configured logging/auth/feature-flag overlays silently disappear on the first cancel/flip. WARN log is the only signal.
- **NEW-X bare `ON CONFLICT DO NOTHING`** (commit `2169a106`) converts non-PK UNIQUE collisions into silent drops. Safe today because `validate()` catches via row-count + SHA-256 PK-set checksums. Risk: future maintainer mutating DuckDB source between copy + validate would have drift unnoticed.
- **`AGNES_ALLOW_RESERVED_CLOUD_URL=1` env opt-in** disables the MED-2 reserved-range guard. Test fixtures set it. On production, the variable must never leak into the FastAPI process environment — otherwise SSRF protection is silently disabled.

---

### Testing gaps

- No test for the H1-PARTIAL race window (post-`update_step('flip_backend')`, pre-`_check_cancel_before_flip` cancel arrival).
- No test asserts `_resolve_host`-based DNS resolution is re-checked between API queue time and migrator connect time (B1-NEW).
- No test feeds YAML-special chars through the bash fallback writer (B2-NEW). `test_applier_yaml_writer_no_pyyaml.py` covers round-trip of plain URLs only.
- No test for stuck-running recovery → app/scheduler down → next applier tick (B3-NEW). Existing recovery tests assert the JSON file state, not the docker-compose service state.
- No test for PG-source rollback with `source_url=None` (H1-NEW).
- No test asserts `_validate_cloud_url` rejects hostname inputs that resolve to reserved IPs (MED-1-PARTIAL).
- No test asserts `validate()` actually catches a synthesized non-PK UNIQUE collision (NEW-X residual).
- No test asserts non-database top-level keys SURVIVE the bash fallback path (H4 residual — operator-disappearing-overlay).

---

### Non-state-machine commits (cleared)

Scope-creep commits in this round were independently audited and do not interact with the state-machine surface:

- `f1911a3b` web timestamps in analyst timezone — frontend-only
- `5b41061e` data-semantics workspace scaffold — orthogonal
- `ec3d161b` LLM auto-doc for tables — orthogonal admin path
- `af0e2b2e` browser-based CLI auth — orthogonal (gh-style flow)
- `09a15a02` catalog status pill fix — pure UI
- `81c444f9` system.duckdb OOM resilience — defensive merge integration
- `d1f3526b` route migrator duckdb.connect through `_open_duckdb` — no new SQL/path injection surface
- `2169a106` `ON CONFLICT DO NOTHING` covers non-PK UNIQUEs — see NEW-X residual

---

**Net status:** the architecture continues to hold. The fix wave addressed every previously-flagged item correctly. The new BLOCKERs cluster narrowly around (a) **hostname resolution between validation and use** (B1-NEW + MED-1-PARTIAL), (b) **bash-YAML fallback fidelity** (B2-NEW), and (c) **failure-path completeness in the applier** (B3-NEW + H1-NEW). These are smaller, more localised than rounds 1-2. Round-4 should be a 5-10-commit pass.

ZdenekSrotyr:
Round-3 review addressed in **0.57.1** — see CHANGELOG `[0.57.1] — 2026-06-01` and the PR description's "Round-3 fixes shipped" section. All 6 findings (3 BLOCKER + 1 HIGH + 2 PARTIAL) have a fix commit + regression test on this branch. Full suite GREEN: **5986 passed / 36 skipped / 0 failed**. Ready for round-4.

ZdenekSrotyr:
Merged + released **v0.57.1** ([release notes](https://github.com/keboola/agnes-the-ai-analyst/releases/tag/v0.57.1)).

**Post-merge CI on main (commit `fc0289f6`):**
- Release workflow ✓ (build-and-push ✓, **smoke-test ✓**, e2e-bind-mount ✓, rollback-on-smoke-fail skipped per design)
- CI test suite ✓
- Lint workflows ✓

Thanks @cvrysanek for the three review rounds — 26 findings total across rounds 2 + 3 (13 BLOCKER + 16 HIGH-class + 6 MED + 2 LOW + 1 live-discovered), all addressed under TDD discipline + sub-agent execution. Branch `zs/db-state-machine` left intact (not deleted) for traceability + potential round-4 follow-up.

Two non-blocking observations from internal code-quality review of R3-T6 worth a future scrub:
1. Dead `app_restart` / `verify_health` entries in PNR step tuple at `app/api/db_state.py:722` and `:771` — pre-existing, migrator never emits those step names; extract `_PNR_STEPS = frozenset({"flip_backend"})` constant.
2. Local re-import of `MigrationLock` / `MigrationInProgressError` in `scripts/db_state_migrator.py:1199` may be redundant.

Also one non-gating signal at merge time: 3rd-party `Devin Review` check reported `fail` with no PR comment / no required-check status, so it didn't block. If you want a Devin-aware round-4, the report lives at app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/455.

PR #473, opened 2026-06-01
fix(web): render frontend timestamps in analyst's local timezone

cvrysanek:
## Summary

- **DuckDB session timezone pinned to UTC** on every connection via `src.db._open_duckdb` (uses `SET GLOBAL` — session-level `SET TimeZone` does NOT propagate to `conn.cursor()`, which every repository reads through).
- **FastAPI labels naive datetimes with an explicit UTC offset on the wire** — `AgnesJSONResponse` set as the default response class plus an override of `fastapi.encoders.ENCODERS_BY_TYPE[datetime]` so `serialize_response` produces offset-bearing ISO strings before `render()` ever runs. JS `new Date()` stops misinterpreting them as local.
- **`window.AgnesTime`** (`app/web/static/js/datetime.js`) hydrates `<time datetime>` tags to local tz; replaces the broken `fmtDate(s) = s.slice(0,16).replace('T',' ')` helpers across admin tables that used to display UTC characters as if they were local. UTC label stays as the no-JS fallback and tooltip.

## Scope

No DuckDB schema migration — deferred until the parallel Postgres migration lands (per discussion on this branch). Legacy rows written on non-UTC hosts before this branch may be off by their host's UTC offset; new writes are correct.

## Test plan

- [x] `.venv/bin/pytest tests/test_duckdb_session_tz.py -v` — 5 green (incl. cursor inherits UTC, no shift on aware-UTC write)
- [x] `.venv/bin/pytest tests/test_datetime_serialization.py -v` — 6 green (incl. real `/api/health` endpoint smoke)
- [x] `.venv/bin/pytest tests/test_health_session_pipeline.py tests/test_session_pipeline.py -v` — 42 green (sync-lag + mtime compare now UTC-naive on both sides)
- [x] Full suite `.venv/bin/pytest tests/ --tb=short -n auto -q` — 5324 passed; 7 unrelated env failures (missing `jsonschema` dep locally, `da` CLI on PATH, perf flake under parallel load, stale catalog_export dir) all reproduce on a clean checkout of `origin/main`.
- [ ] Manual: load `/admin/users`, `/admin/groups`, `/admin/marketplaces`, `/admin/store/submissions/<id>`, `/admin/scheduler/runs`, `/news`, `/marketplace` in a non-UTC tz; hover a timestamp; confirm tooltip = UTC, displayed text = local.
- [ ] Manual: disable JS in DevTools; confirm UTC fallback still renders.

## Spec / plan

- `docs/superpowers/specs/2026-05-26-frontend-timezone-fix-design.md`
- `docs/superpowers/plans/2026-05-26-frontend-timezone-fix.md`

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/473" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


cvrysanek:
## Adversarial review pass — follow-up commit pushed

`compound-engineering:ce-adversarial-reviewer` surfaced 10 findings. Acting on the four hottest now, deferring two by explicit scope, three are advisory.

### Acted on

**#1 critical — extractor / CLI / script bypass** (high confidence). Every `duckdb.connect()` site outside `src/` and `app/api/v2_*` ran without the UTC pin. On a non-UTC host the Keboola / BigQuery / Jira extractors stored every `_meta.extracted_at` timestamp as host-local-clock naive — same root cause this PR was supposed to fix.

Fix: extracted `_open_duckdb` to `src/duckdb_conn.py` (no heavy deps), routed `connectors/keboola/{access,extractor}`, `connectors/bigquery/access`, `connectors/jira/extract_init`, `cli/{lib/pull,commands/query,commands/explore,commands/snapshot}`, `scripts/{duckdb_manager,migrate_parquets_to_extracts}` through it. `connectors/bigquery/extractor.py` uses a local `_pin_session_utc` around raw `duckdb.connect` because the test suite heavily patches `connectors.bigquery.extractor.duckdb` and routing through a different module would defeat the intercept.

**#9 helper swallowed `SET` failures silently.** Added: verify `current_setting('TimeZone')` after the `SET`; `log.warning` if not UTC; raise under `AGNES_DUCKDB_TZ_STRICT=1` for CI / strict deploys.

**#4 `parseIso` accepted non-strings + mis-handled date-only.** Added: type-guard `typeof s !== "string"` rejection; explicit UTC anchoring for `YYYY-MM-DD` (avoids the one-day shift in zones west of UTC).

**Regression guard.** New `tests/test_duckdb_session_tz.py::test_no_bare_duckdb_connect_in_production_code` greps every production Python source and fails if any new `duckdb.connect(` slips in without routing through the helper or joining the explicit allow-list.

### Deferred by scope

**#3 / #8 legacy data on non-UTC hosts.** Rows written before this PR carry host-local-clock-naive values. Health-check + scan_unprocessed_for now compare against UTC-naive → first post-deploy tick may misfire on legacy rows (self-heals after one `mark_processed` cycle). Operators of non-UTC deployments may want to truncate `session_processor_state` post-deploy. Cannot backfill without per-row write-time tz which isn't recorded. Noted in CHANGELOG.

### Advisory (left as-is)

**#2 misleading docstring** — clarified in the new `src/duckdb_conn.py`: pin is per-DuckDB-instance, not process-wide; that's exactly why every call site must funnel through the helper.

**#5 AJAX-inserted `<time>` not hydrated.** Confirmed: dynamic templates (e.g. `admin_user_detail.html:821`) build `<time>` via JS and skip the hydrator. Low-severity — the JS already calls `fmtDate` so the visible text is local; tooltip absence is a minor UX gap. Will address in a follow-up if the inconsistency is noticed.

**#6 `ENCODERS_BY_TYPE` mutation** — import-time global mutation is a known FastAPI/Pydantic pattern; the test isolation footgun is real but bounded (no test in the suite asserts offset-less ISO). Re-evaluate if FastAPI freezes the registry in a future version.

### Test plan after follow-up

- `pytest tests/test_duckdb_session_tz.py` — 6 green (incl. regression guard)
- `pytest tests/test_bigquery_extractor.py tests/test_bigquery_extractor_full.py tests/test_keboola_extractor.py` — 69 green
- `pytest tests/test_datetime_serialization.py tests/test_health_session_pipeline.py tests/test_session_pipeline.py` — all green
- Full suite running; results in next commit if any new regression surfaces

PR #485, opened 2026-06-01
feat: make container memory caps overridable via .env (0.56.1)

ZdenekSrotyr:
Makes the Docker container memory caps configurable instead of hard-coded, so a deployment on a host larger than ~4 GiB can use the available RAM without forking `docker-compose.yml`.

## What

- `app` service: `mem_limit: ${AGNES_APP_MEM_LIMIT:-4g}`
- `scheduler` service: `mem_limit: ${AGNES_SCHEDULER_MEM_LIMIT:-2g}`

Defaults are unchanged (4g / 2g), so existing small / single-VM deployments behave exactly as before. The vars are read from the `env_file` (`.env`) that compose already loads, and `docker compose config` confirms interpolation:

```
default   → app 4294967296 (4g)  scheduler 2147483648 (2g)
override  → app 10737418240 (10g) scheduler 4294967296 (4g)
```

## Why

A fixed `4g` cap is right for a memory-constrained host but wrong for a larger one. DuckDB enforces `memory_limit` per-connection and (1.5+) defaults a fresh connection to ~80% of the cgroup limit, so on, say, a 16 GiB VM a `4g` container both wastes host RAM **and** lets the per-connection budgets sum past the cap — at which point the cgroup OOM-killer SIGKILLs uvicorn mid-WAL-write and the torn `system.duckdb.wal` fails replay on the next boot (the exact corruption the `stop_grace_period` comment in compose and the per-connection caps in `src/db.py` guard against). Raising this cap is meant to go hand-in-hand with raising those per-connection budgets.

## Release-cut

`[Unreleased]` held only this change, so the cut ships here: `pyproject.toml` → 0.56.1, CHANGELOG renamed, new empty `[Unreleased]`. Patch — no breaking change, defaults preserved.

## Tests

`pytest tests/ -n auto` — 5403 passed. The 3 `test_corporate_memory_v1::TestSchemaValidation` failures (`ModuleNotFoundError: jsonschema`, a missing local dev module) and 2 smoke errors are pre-existing and unrelated to this diff — confirmed via `git stash` on a clean tree (corporate_memory still fails; the 2 smoke tests pass in isolation → xdist flake). This PR touches no Python.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/485" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ZdenekSrotyr:
Scope expanded: the PR now also wires the same knob through the `infra/modules/customer-instance` Terraform module (the OSS module that the Pattern-A customer infra repos pin), so Pattern-A deployments get the override too — not just the bare-compose / Pattern-B path.

- `prod_instance` / `dev_instances` gain optional `app_mem_limit` (default `4g`) / `scheduler_mem_limit` (default `2g`) attributes.
- `startup-script.sh.tpl` renders them into `/opt/agnes/.env` next to `AGNES_TAG`.
- `terraform validate` + `fmt` pass on the module.

**Post-merge:** cut both `v0.56.1` (app) and a new `infra-v1.10.0` (module) tag on the merge commit — the customer infra repos bump their module pin to `infra-v1.10.0` to pick up the new attributes (optional + defaulted → backward-compatible, hence a minor module bump).

PR #514, opened 2026-06-03
fix(marketplace): admin god-mode subscriptions surface in served set

cvrysanek:
Admin user POSTs /api/marketplace/curated/<m>/<p>/install. The endpoint require_resource_access calls can_access, which short-circuits to True for Admin group → subscription row lands. But resolve_user_marketplace composed (resolve_allowed_plugins ∩ subscriptions) and resolve_allowed_plugins is strict (admin treated as regular group, no god-mode — pinned by test_admin_filtered_through_grants_like_anyone_else). Intersection drops admin subscriptions without explicit grants.

## Symptom (foundryai-development, prod)

Admin installs legal-marketplace/legal-common via web UI → `agnes marketplace search` reports installed=true → but `~/.agnes/marketplace/.claude-plugin/marketplace.json` from /marketplace.git/ has only plugins explicitly granted to Admin group (e.g. grpn-foundryai). `agnes refresh-marketplace` runs cleanly against the stale aggregate; never sees the new subscription; never installs into Claude Code.

## Fix scope

Targeted at resolve_user_marketplace only. When the caller is in the Admin group, widen the eligibility source to ALL marketplace_plugins rows (matching install-endpoint can_access semantic). Subscription intersection still applies — subscription remains the explicit intent. Non-admins still need explicit grants.

resolve_allowed_plugins semantic stays strict. The pinned test test_admin_filtered_through_grants_like_anyone_else still passes. New behavior is restricted to the served-set composer one layer up.

## Test plan
- [x] tests/test_marketplace_filter.py — 17 passed (15 prior + 2 new)
- [x] TestResolveUserMarketplace::test_admin_subscriptions_surface_without_explicit_grants — admin subscribes without grant, served set includes plugin
- [x] TestResolveUserMarketplace::test_non_admin_still_needs_explicit_grant — non-admin with subscription but no grant, plugin filtered
- [ ] Live verification on dev: admin installs legal-marketplace/legal-common via /marketplace UI → agnes refresh-marketplace → installed_plugins.json contains legal-marketplace-legal-common@agnes

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/514" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


cvrysanek:
Closing — diagnosis was wrong.

The original premise was "admin god-mode lets you install a curated plugin without minting an explicit `resource_grants` row, then served `/marketplace.git/` excludes it because the read path is strict RBAC." But multiple non-admin users hit the same symptom, which rules that out.

Real cause: stale Google Workspace group memberships. `user_group_members.source='google_sync'` rows refresh **only** in the browser OAuth callback (`app/auth/group_sync.fetch_user_groups` is invoked from `app/auth/providers/google.py`). CLI / PAT traffic — `agnes refresh-marketplace`, `agnes pull`, the whole `/api/marketplace/*` surface — never re-fetches groups. So:

- user is added to a Workspace group (or membership shifts)
- the grant on that group is fine, ingestion is fine
- but the user's last browser sign-in was earlier, so Agnes' snapshot of their groups is frozen
- `resolve_allowed_plugins` joins against the stale snapshot → 0 rows → plugin filtered out of `resolve_user_marketplace`'s served set
- `agnes marketplace search` still reports `installed: true` because `_curated_to_item` reads `user_curated_subscriptions` directly, no RBAC filter on that flag → divergent UX signals.

Workaround for affected users: sign out and back in on the dashboard.

Real fix tracked separately — needs server-side periodic refresh (or lazy refresh on stale snapshot, or an `agnes refresh-groups` CLI). Re-opening on a fresh branch.

This branch has the original commit + its revert (`net diff = 0`); closing rather than merging.

PR #515, opened 2026-06-03
fix(web): no-store on dynamic HTML so /home onboarded toggle takes effect on reload

cvrysanek:
## Problem

On `/home`, the **"Already set up? Mark me as onboarded →"** action (and the offboard strip) reloads the page but the setup section stays visible — the toggle looks broken.

## Root cause

Not the onboarded logic — that path is provably correct:

- `POST /api/me/onboarded` commits the `users.onboarded` UPDATE; the write is visible to the next request (verified: shared singleton DuckDB connection, cross-cursor visibility holds, single uvicorn worker).
- `home_page` reads `onboarded` with a fresh `SELECT` and the template branches on it correctly.
- The existing `test_home_no_auto_transition_after_post_until_reload` (real handlers + real DB) passes: GET → POST → GET hides the setup block.

The gap is **HTTP caching**. `/home` is an authenticated, state-dependent HTML page, but it (and all dynamic HTML) shipped with **no `Cache-Control`** — confirmed in `app/main.py` and the `Caddyfile`. So after the button POSTs, `window.location.reload()` can re-render the **browser/bfcache/disk-cached** pre-action document — the setup section never hides.

## Fix

Stamp `Cache-Control: no-store` on `text/html` responses that haven't already declared a caching policy (in the existing `_add_version_headers` middleware). Non-HTML assets — css/js/images, parquet downloads, the immutable store/marketplace media endpoints — set their own `Cache-Control` and are left untouched. Also fixes back-button staleness on every authed dashboard page.

## Test

- New `test_home_sends_no_store_cache_control` asserts `/home` returns `Cache-Control: no-store`.
- `tests/test_web_home_page.py`: 13 passed.
- Asset/immutable-endpoint suite (store, marketplace, cover photos): 501 passed — their `Cache-Control` is untouched.

## Test suite note

Full suite has 28 failures **all pre-existing / unrelated to this diff** (confirmed by re-running on a stashed clean base):

- `test_db_schema_version::test_schema_version_is_59` + `test_schema_v58_to_v59_migration` (3) — base branch is already at **schema v60** (the cli-browser-login commit); these tests still pin v59. Stale expectation, predates this PR. *(Worth a separate cleanup — out of scope here.)*
- `test_e2e_corporate_memory` (chromium) + a playwright `select_option` timeout — browser E2E, flaky under concurrent-suite contention.
- `test_catalog_export::TestExportMetrics` (2), `test_cli_diagnose` (2), `test_cli_binary_rename`, `test_api_design_rules::test_creator_post_declares_201` — unrelated to a response header; reproduce on clean base.

This diff adds **zero** new failures.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/515" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 ## [Unreleased]
 
 ### Added
+- Browser-based `agnes auth login` (gh-style loopback). Instead of
+  prompting for a plaintext password, the CLI opens the browser to
+  `/cli/auth/start`, the user signs in with whatever provider their
+  account uses (Google / magic link / password), and on approval the
+  server hands a 90-day personal access token straight back to the CLI
+  over a localhost loopback — no copy/paste, no password in the
+  terminal. The durable token never travels through the browser URL: the
+  loopback carries only a single-use, ~2-min exchange code (new
+  `cli_auth_codes` table, schema v60) that the CLI trades for the PAT via
+  `POST /cli/auth/exchange`. Server routes: `GET/POST /cli/auth/start`,
+  `POST /cli/auth/exchange`. Terminal-only fallbacks preserved:
+  `agnes auth login --password` (email+password) and `--no-browser`
+  (prints the sign-in URL); a timeout or old server prints the manual
+  `agnes auth import-token` path.
 - `instance.custom_scripts`: operator-injected HTML/JS blocks rendered
   into every page that extends `base.html`. Each entry takes `name`,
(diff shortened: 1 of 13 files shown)


cvrysanek:
Superseded by #518. The no-store cache theory was wrong: the real cause is /home reading users.onboarded from DuckDB via raw conn.execute while POST /api/me/onboarded writes through the backend-aware users_repo() — so on a Postgres-backed (db-state-machine CLOUD/SIDE_CAR) instance the write goes to Postgres and the read returns the stale DuckDB row. Not cache, not frontend. Fixed in #518.

PR #518, opened 2026-06-03
fix(web): /home reads onboarded via repo factory (works on Postgres backend)

cvrysanek:
## Problem

On a Postgres-backed instance, `/home`'s "Mark me as onboarded" button (and the hero X-close, and `agnes init`) appears to do nothing: the page reloads but the setup panel never hides — regardless of reload, hard refresh, or cache state.

## Root cause

Backend split between the write and the read:

- `POST /api/me/onboarded` writes `users.onboarded` through the **backend-aware** `users_repo()` factory.
- `app/web/router.py::home_page` read it back with a **raw `conn.execute("SELECT onboarded …")`** against `conn = Depends(_get_db)` — which is **always DuckDB**.

On a `db-state-machine` **CLOUD / SIDE_CAR (Postgres)** instance (`use_pg() == True`), the write lands in Postgres but `/home` reads the stale pre-migration DuckDB row. The flag flips in Postgres while `/home` renders the setup view forever. On a plain DuckDB instance both sides hit the same DB, so it works — which is why it reproduced on a deployed Postgres box but not on a local dev server.

### Verified on the live instance

On the affected box (`use_pg()=True`, backend = CLOUD Postgres):
- `users_repo().get_by_email(...)` → `onboarded=True`, `updated_at` = today (the button POSTs **were** persisting, to Postgres).
- The DuckDB `users` row → `onboarded=False`, `updated_at` months stale.

Not a frontend bug (the link and the X + confirm-modal paths both hide the panel correctly in a browser) and not a cache bug (hard refresh doesn't help — the server genuinely returns the stale DuckDB value).

## Fix

`/home` reads `onboarded` through `users_repo().get_by_id(...)`, so the read and the write share the active backend.

## Test

`test_home_reads_onboarded_through_repo_factory_not_raw_duckdb` simulates the split without a live Postgres: the DuckDB row says `onboarded=False` while the repo factory reports `True` (the real backend); `/home` must honor the repo and render the nav hub. Fails before the fix, passes after. `tests/test_web_home_page.py`: 18 passed.

## Follow-up (out of scope here)

`home_page` still passes the raw DuckDB `conn` to `is_user_admin(...)` and `compute_home_stats(...)`, and other web routes use `Depends(_get_db)` directly for system-table reads. These are similarly backend-unaware and likely misbehave on Postgres-backed instances. Worth a broader audit of raw-`conn` system reads vs. the repo factory.

---

<a href="https://app.devin.ai/review/keboola/agnes-the-ai-analyst/pull/518" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Fixed
+- `/home` "Mark me as onboarded" (and `agnes init`) now takes effect on a
+  Postgres-backed instance. The route read `users.onboarded` with a raw
+  `conn.execute` against DuckDB while `POST /api/me/onboarded` writes through
+  the backend-aware `users_repo()` — so on a `db-state-machine` CLOUD /
+  SIDE_CAR instance the flag was written to Postgres but read back from the
+  stale DuckDB row, leaving the setup panel visible forever regardless of
+  reloads or cache. `/home` now reads `onboarded` through `users_repo()` so
+  the read and write share the active backend.
+
 ## [0.61.5] — 2026-06-03
 
 ### Added
--- a/app/web/router.py
+++ b/app/web/router.py
@@ -764,10 +764,16 @@ async def home_page(
 
     See origin: docs/brainstorms/home-page-requirements.md.
     """
-    row = conn.execute(
-        "SELECT onboarded FROM users WHERE id = ?", [user["id"]]
-    ).fetchone()
-    onboarded = bool(row[0]) if row else False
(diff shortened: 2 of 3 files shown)


cvrysanek:
Test suite: home tests 18/18 pass incl. the new `test_home_reads_onboarded_through_repo_factory_not_raw_duckdb` regression guard (fails without the fix). Full suite has 24 failures, **all pre-existing / unrelated to this diff** (confirmed by re-running each on a stashed clean base): `test_e2e_corporate_memory[chromium]` + a playwright `select_option` timeout (browser E2E, flaky under a concurrent local test run); `test_cli_diagnose` (2), `test_duckdb_session_tz::test_no_bare_duckdb_connect_in_production_code`, `test_catalog_export::TestExportMetrics` (2), `test_cli_binary_rename`. This diff adds zero new failures.

PR #539, opened 2026-06-04
fix(db): run internal-table SQL against the active backend (Postgres)

ZdenekSrotyr:
## What — the last internal-table PG parity gap

The internal-table SQL feature lets an analyst run **arbitrary** DuckDB SQL over Agnes's own telemetry — `agnes query "SELECT … FROM agnes_telemetry"` (also `agnes_audit`, `agnes_sessions`). `execute_internal_query` wraps the SQL so each `agnes_*` name maps to the real table scoped to the caller's rows (RBAC), then runs it in DuckDB.

On a **Postgres** instance those rows live in PG, but the query still ran against the empty DuckDB system file → the same query returned **nothing** instead of the data it returns on DuckDB. (`v2_sample`'s fixed-shape preview was the easy half, fixed in #535; this is the arbitrary-SQL half.)

This is *not* a factory swap — the SQL is arbitrary DuckDB SQL (analyst GROUP BYs, date functions, JOINs); you can't run it through a repo method, and you can't run DuckDB SQL on Postgres (different dialect).

## How — keep it in DuckDB, ATTACH Postgres

On PG, `execute_internal_query` now runs in an in-memory DuckDB with the Postgres database **ATTACHed read-only** via DuckDB's `postgres` extension, and the `agnes_*` CTEs point at the attached tables (`pgsys.public.<table>`). The analyst's SQL still executes in a DuckDB engine → **identical behaviour on both backends**. The postgres extension **streams** rows from PG, so there's no per-query materialisation/memory blowup — even for an admin-unscoped query over a large `usage_events` (which is why no row cap was needed; behaviour matches DuckDB exactly).

## Security

- RBAC scoping unchanged (per-caller `WHERE` on the CTEs; admin unscoped).
- The non-admin base-table **denylist** already rejects Postgres-catalog-qualified references — the bare `users` token in `pgsys.public.users` still matches the word-boundary scan.
- **New hard guard**: a non-admin referencing the attach catalog (`pgsys`) at all is rejected, so PG system catalogs (`pgsys.pg_catalog.*`, whose names aren't in the denylist) are unreachable.
- The attach handle uses `_open_duckdb` (UTC-pinned, satisfies the tz guard) for timestamp parity.

## Test

`tests/db_pg/test_parity_internal_query.py` — on DuckDB **and** real Postgres:
- non-admin sees only their own rows; admin (unscoped) sees all;
- non-admin referencing a base table is rejected;
- (PG-only) non-admin referencing `pgsys.pg_catalog.*` is rejected.

The `[pg]` variants return empty / would leak before this change. Full suite green (6975 passed; 1 unrelated xdist-isolation flake).

## Stacking

Independent — branches off `main`. Completes the internal-table PG parity (with #535).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,25 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Fixed
+- **Postgres backend: `agnes query` against internal tables returned nothing.**
+  The internal-table SQL feature (analyst SQL over `agnes_telemetry` /
+  `agnes_audit` / `agnes_sessions`) ran the query in DuckDB against the system
+  file, so on a Postgres instance — where those rows live in PG — the same query
+  returned an empty result instead of the data DuckDB would show. It now reads
+  the caller's RBAC-filtered rows from Postgres into a per-request in-memory
+  DuckDB (one table per referenced internal source) and runs the analyst's
+  arbitrary DuckDB SQL there, so behaviour is identical on both backends. The
+  postgres extension is deliberately NOT loaded and nothing is ATTACHed, so the
+  `postgres_*` table functions (which take the catalog as a string literal and
+  would slip past identifier guards) are unavailable; and because the filter is
+  applied during materialisation, the materialised table is itself the RBAC
+  boundary — a user CTE that shadows an `agnes_*` alias still reads only the
+  caller's rows. A 1M-row-per-table materialisation cap protects against an
+  admin-unscoped query over a huge table (raises rather than risking an OOM).
+  Pinned by both-backends parity + RBAC-escape tests
+  (`tests/db_pg/test_parity_internal_query.py`).
+
 ## [0.65.8] — 2026-06-04
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


ZdenekSrotyr:
⚠️ **Converted to draft — adversarial RBAC review found 2 CRITICAL security holes in the ATTACH approach. Do not merge.**

- **postgres-extension TVF escape:** `postgres_query('pgsys','SELECT * FROM users')` / `postgres_scan('pgsys','public','users')` pass the catalog as a *string arg*, so `_strip_sql_noise` strips it → the `pgsys` guard and the table-name denylist both miss it. A non-admin could read the full Postgres `users` table (emails, password hashes, PATs).
- **CTE-shadow:** user SQL can open its own `WITH agnes_telemetry AS (…)` that shadows the scoped CTE → unscoped read (and combined with the TVF on PG, fully bypasses).
- **perf:** the 'streams from PG, no materialisation' claim is false for admin-unscoped queries (full table into DuckDB).

Reworking to a **materialisation** approach: read only the RBAC-filtered rows from PG into temp tables in an in-memory DuckDB with **no postgres extension loaded and nothing ATTACHed** — which removes the entire TVF attack surface and makes the materialised temp table itself the RBAC boundary (so CTE-shadow can't reach unfiltered data), plus a row cap for the admin-unscoped case.

PR #548, opened 2026-06-04
fix(sync): sweep orphaned kbc-export-*/kbc-slice-* scratch dirs before each Keboola sync

ZdenekSrotyr:
## Problem

A production instance hit **`/data` at 99 %** and `/admin/sync` showed `No space left on device`. Root cause: **40 GB of orphaned `kbc-export-*` staging dirs** in the temp root (`AGNES_TEMP_DIR`), accumulated from earlier failed syncs.

The Keboola extractor creates per-export staging dirs via `tempfile.TemporaryDirectory`, whose `__exit__` removes the dir on **any** return — including the ENOSPC disk-full exception path. So a normal failure cleans up after itself. The only way a dir survives is a **hard kill** (SIGKILL / OOM / the auto-upgrade `docker compose up -d` recreating the container mid-sync — the orphan-maker already documented in `app/api/sync.py`), where `__exit__` never runs.

Once orphans accumulate and fill the disk, **every subsequent sync fails on ENOSPC** — the scheduler's `data-refresh` job keeps firing every 2 min but can't self-heal, because each retry needs scratch space the orphans are holding. The pre-flight guard in `storage_api._download_single` correctly aborts new big downloads ("Free space before retrying") so it stops making *new* large orphans, but nothing reclaims the *old* ones — it needs a manual `rm` to break the loop.

## Fix

Add `storage_api.sweep_orphaned_scratch(root, max_age_seconds)`:

- Removes `kbc-export-*` dirs under the temp root, returns the count, logs each removal (no silent caps).
- **Age-gated** via `AGNES_SCRATCH_MAX_AGE_SEC` (default **1h**, comfortably longer than any single table export) so a concurrent in-flight export — e.g. in another container — is never swept out from under itself.
- No-op when `AGNES_TEMP_DIR` is unset (system `/tmp` is cleared on reboot anyway).

Called at the start of `_run_sync`, **under `_sync_lock`** and **before any new scratch is created**, so within this process it can never race a live export. Best-effort: a sweep failure prints a notice and never blocks the sync.

## Tests

- `TestSweepOrphanedScratch` (6 cases): removes old / keeps fresh / ignores non-`kbc-export-*` entries / no-op on `None` + missing root / env-default threshold.
- Full suite: **7029 passed, 70 skipped**. Two unrelated failures in `tests/test_binding_scope.py` (`_duckdb` redeem-throttle / audit-params) are **parallel-execution flakes under `-n auto`** — they pass in isolation both with and without this diff, and the test file has zero coupling to anything this change touches (no `kbc-export` / `get_temp_root` / `_run_sync` / sweep references).

## Notes

- Connector-only change; no repository methods touched, so no DuckDB↔PG parity work needed.
- Vendor-agnostic: no customer-specific content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ CalVer image tags (`stable-YYYY.MM.N`, `dev-YYYY.MM.N`) are produced for every C
 
 ## [Unreleased]
 
+### Fixed
+- The Keboola sync now sweeps orphaned `kbc-export-*` / `kbc-slice-*` staging dirs from the temp root (`AGNES_TEMP_DIR`) at the start of every run. These dirs are normally removed by `TemporaryDirectory` on any return — including the disk-full path — so the only way they survive is a hard kill (SIGKILL / OOM / container recreate) mid-export. Without a sweep they accumulated on the data disk until it filled and *every* subsequent sync failed with `No space left on device`, a self-reinforcing failure that needed a manual `rm` to break. The sweep is age-gated (`AGNES_SCRATCH_MAX_AGE_SEC`, default 1h) so a concurrent in-flight export is never deleted, and runs under the sync lock before any new scratch is created.
+
 ## [0.65.18] — 2026-06-04
 
 ### Added
--- a/app/api/sync.py
+++ b/app/api/sync.py
@@ -380,6 +380,22 @@ def _run_sync(tables: Optional[List[str]] = None):
         source_type = get_data_source_type()
         data_dir = _get_data_dir()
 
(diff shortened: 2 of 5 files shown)


ZdenekSrotyr:
Addressed architecture review: the sweep now also covers `kbc-slice-*` staging dirs created by `_download_sliced` (the sliced-CSV path), not just `kbc-export-*`. Both prefixes orphan on the same hard-kill scenario. Added `test_removes_old_slice_dirs`; full suite green (7032 passed, 0 failed — the earlier 2 `test_binding_scope` failures were confirmed `-n auto` flakes, pass on re-run).

